PR #64, opened 2015-07-29
Makefile for compiling and installing Plugins.

oobenland:


solr_plugins/Bibelstellensuche/src/de/uni_tuebingen/ub/ixTheo/bibleRangeSearch/BibleRangeParser.java
```diff
@@ -24,6 +25,21 @@ public BibleRangeParser(final String queryString, final SolrParams localParams,
         }
     }
 
+    /**
+     * Tries to extract the book index of a search query.
+     * Then creates a query string by concatenating the book index with '*'.
+     * If no book index is found, only '*' will be returned.
```
ruschein: I'd add that the "book index" is the first two positions that that it corresponds to a book of the bible and is specified in a file named blah, blah blah.  I think it is otherwise difficult for a newbie to understand what we're talking about.

PR #78, opened 2015-08-07
Bugfix for install script

oobenland:


install/scripts/set_privileges.sh
```diff
@@ -42,6 +42,8 @@ touch "$VUFIND_LOCAL_DIR/logs/record.xml"
 touch "$VUFIND_LOCAL_DIR/logs/search.xml"
 
 touch "/var/log/vufind.log"
+chown -R "$OWNER" "/var/log/vufind.log"
```
ruschein: Why "-R"?  It's a single file.

PR #100, opened 2015-08-14
Ruschein

ruschein:


cpp/enrich_keywords_with_title_words.cc
```diff
@@ -98,29 +84,67 @@ void ExtractKeywordsFromKeywordChainFields(
...
     const std::vector<DirectoryEntry> &dir_entries,
     const std::vector<std::string> &field_data,
-    const Stemmer &stemmer, std::unordered_map<std::string, std::string> * const stemmed_to_unstemmed_keywords_map)
+    const Stemmer * const stemmer,
+    std::unordered_map<std::string, std::string> * const stemmed_to_unstemmed_keywords_map)
 {
+    size_t keyword_count(0);
     std::vector<std::string> keywords;
-    MarcUtil::ExtractAllSubfields("600:610:611:630:650:653:656", dir_entries, field_data, &keywords);
+    MarcUtil::ExtractAllSubfields("600:610:611:630:650:653:656", dir_entries, field_data, &keywords, "02");
```
oobenland: Please use a constant for "02" to show that this field is ignored

cpp/lib/Stemmer.cc
```diff
@@ -12,3 +12,14 @@ Stemmer::Stemmer(const std::string &language_name_or_code) {
 std::string Stemmer::stem(const std::string &word) const {
     return std::string((const char *)::sb_stemmer_stem(stemmer_, (const sb_symbol *)word.c_str(), word.size()));
 }
+
+
+Stemmer *Stemmer::StemmerFactory(const std::string &language_name_or_code) {
+    Stemmer *new_stemmer(NULL);
```
oobenland: Maybe do the caching thing here instead of enrich_keywords_with_title_words::GetStemmer?

cpp/lib/TextUtil.cc
```diff
@@ -172,11 +176,55 @@ bool ChopIntoWords(const std::string &text, std::unordered_set<std::string> * co
...
+						       const std::vector<std::string> &needle)
+{
+    if (needle.empty())
+	return haystack.cbegin();
+
+    if (haystack.size() < needle.size())
+	return haystack.cend();
+
+    const std::vector<std::string>::const_iterator haystack_start(
+	std::find(haystack.cbegin(), haystack.cend(), needle[0]));
```
oobenland: You have to check for every occurrence of the needle[0] and then for the rest of needle.

solr_plugins/Schlagwortfolgensuche/src/de/uni_tuebingen/ub/ixTheo/keywordChainSearch/KeywordChainMetric.java
```diff
@@ -0,0 +1,77 @@
...
+
+public class KeywordChainMetric {
+
+    static int calculateCommonPrefixLength(final String s1, final String s2) {
+	final int minLength = Math.min(s1.length(), s2.length());
+	int commonPrefixLength = 0;
+
+	for (int i = 0; i < minLength; ++i) {
+	    if (s1.charAt(i) != s2.charAt(i))
+		return commonPrefixLength;
```
oobenland: You count i and commonPrefixLength. They have the same value. So, why not returning i?

PR #128, opened 2015-08-25
Utility for collecting stats about ixTheo notations in MARC data sets.

ruschein:


cpp/ixtheo_notation_stats.cc
```diff
@@ -0,0 +1,157 @@
...
+#include "util.h"
+
+
+void Usage() {
+    std::cerr << "Usage: " << progname << "marc_input code_to_description_map\n";
+    std::exit(EXIT_FAILURE);
+}
+
+
+void LoadCodeToDescriptionMap(const std::shared_ptr<FILE> &code_to_description_map_file,
```
oobenland: why do you use this list? Why didn't you simplify the syntax of the list, something like
AB=Religionswissenschaft
The parsing of this function is very confusing.

cpp/ixtheo_notation_stats.cc
```diff
@@ -0,0 +1,157 @@
...
+        ++count;
+
+        if (dir_entries[0].getTag() != "001")
+            Error("First field is not \"001\"!");
+
+	int lok_index(MarcUtil::GetFieldIndex(dir_entries, "LOK"));
+	bool found_at_least_one(false);
+	while (lok_index != -1) {
+	    const Subfields subfields(field_data[lok_index]);
+	    const auto begin_end(subfields.getIterators('0'));
```
oobenland: this is a bad name. Maybe subfield_zero_iterator?

cpp/ixtheo_notation_stats.cc
```diff
@@ -0,0 +1,157 @@
...
+	    if (std::distance(begin_end.first, begin_end.second) == 2 and begin_end.first->second == "936ln") {
```
oobenland: You should use `begin_end.end` here. Even thoug you get the result of std::distance` right, it dosn't check if there are only two elements.

cpp/ixtheo_notation_stats.cc
ruschein: Good idea!  Done!

PR #137, opened 2015-08-27
Split ExecUtil::Exec() into two functions w/ a simpler interface.

ruschein:


cpp/ixtheo_marc_pipeline.sh
```diff
@@ -49,18 +49,33 @@ add_isbns_or_issns_to_articles TitelUndLokaldaten-"${date}".mrc \
...
 echo "*** Phase 4 ***"
 echo "*** Phase 4 ***" >> "${log}"
 augment_bible_references TitelUndLokaldaten-with-issns-and-title-keywords-"${date}".mrc \
                          Normdaten-"${date}".mrc \
-                         TitelUndLokaldaten-with-issns-title-keywords-and-bible-refs-"${date}".mrc
+                         TitelUndLokaldaten-with-issns-title-keywords-and-bible-refs-"${date}".mrc >> "${log}" 2>&1
 cp *.map /var/lib/tuelib/bibleRef/
 
+# Phase 5:
+echo "*** Phase 5 ***"
```
oobenland: Maybe a readable name for each phase?

cronjobs/get_config_file_entry.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def Main():
+    if len(sys.argv) != 4:
+        print("usage: " + sys.argv[0] + " config_file section entry", file=sys.stderr)
+        sys.exit(-1)
+
+    config_file = sys.argv[1]
+    if not config_file.startswith("/"):
+        if os.access("/var/lib/tuelib/" + socket.gethostname() + "/", os.R_OK):
```
oobenland: There should be a variable for this path, so you only have to change it once.

PR #165, opened 2015-10-01
Removed an obsolete and unused program and greatly enhanced the Regex…

ruschein:
…Matcher functionality.


cpp/notify_users.cc
```diff
@@ -45,62 +50,308 @@ void Usage() {
...
+    std::string last_parem_name;
+    for (std::string line : lines) {
+	StringUtil::Trim(&line);
+	if (line.empty())
+	    continue;
+
+	switch (parse_state) {
+	case ParseState::ARRAY_EXPECTED:
+	    if (line != "Array")
+		return false;
```
oobenland: Wouldn't it be nice to get some error messages if something fails? So you know where and why it fails?

ruschein: Done!

PR #226, opened 2016-01-11
Ruschein

ruschein:


cpp/lib/include/XmlWriter.h
```diff
@@ -0,0 +1,136 @@
...
+		       const TextConversionType text_conversion_type = NoConversion);
+
+    /** Destroyes an XmlWriter object, closing any still open tags. */
+    ~XmlWriter() { closeAllTags(); }
+
+    /** \brief    Adds another attribute to be used the next time the one-argument version of openTag() gets called.
+     *  \warning  If the two-argument version of openTag() gets called all prior calls to this function will be
+     *            ignored!
+     */
+    void addAttrib(const std::string &name, const std::string &value = "")
```
oobenland: It would be nice if you use the full name: addAttribute(..)

cpp/lib/src/MarcUtil.cc
```diff
@@ -540,10 +590,18 @@ void Record::UpdateRawRecord() const {
 }
 
 
-bool ProcessRecords(FILE * const input, RecordFunc process_record, std::string * const err_msg) {
+Record Record::XmlFactory(File * const /*input*/) {
```
oobenland: Missing implementation

cpp/lib/src/MarcUtil.cc
```diff
@@ -166,9 +167,13 @@ bool Record::recordSeemsCorrect(std::string * const err_msg) const {
     }
 
     if (raw_record_.length() > 99999) {
+	/*
         *err_msg = "record length (" + std::to_string(raw_record_.length())
                    + ") exceeds maxium legal record length (99999)!";
+	*/
         return false;
+	Warning("record length (" + std::to_string(raw_record_.length())
```
oobenland: First returning and then trying to print a warning will not work as expected ;-)

cpp/lib/src/MarcUtil.cc
```diff
@@ -525,12 +530,64 @@ size_t Record::findFieldsInLocalBlock(const std::string &field_tag, const std::s
...
-        Error("Failed to write " + std::to_string(raw_record.size()) + " bytes to MARC output!");
+	throw std::runtime_error("Failed to write " + std::to_string(raw_record.size()) + " bytes to MARC output!");
+}
+
+
+void Record::write(XmlWriter * const xml_writer) const {
+    xml_writer->openTag("record");
+
+    xml_writer->openTag("leader", /* suppress_newline = */ true);
+    (*xml_writer->getFile()) << leader_.toString();
```
oobenland: Wouldn't it be nice to have a xml_writer->write(String) or xml_write->writeTagContent(String)?

PR #234, opened 2016-01-20
Ruschein

oobenland:


cpp/data/create_translation_table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE translations (
+       id INT(11) NOT NULL,
```
oobenland: Should be AUTO_INCREMENT

cpp/translation_db_tool.cc
```diff
@@ -0,0 +1,118 @@
...
+    std::cerr << "Usage: " << progname << " command [args]\n\n";
+    std::cerr << "       Possible commands are:\n";
+    std::cerr << "       get_next start_index language_code\n";
+    std::cerr << "       insert index language_code text\n";
+    std::exit(EXIT_FAILURE);
+}
+
+
+void GetNext(DbConnection * const connection, const unsigned start_index, const std::string &language_code) {
+    const std::string SELECT_STMT("SELECT * FROM translations WHERE index > " + std::to_string(start_index));
```
oobenland: SELECT `id` FROM `translations` WHERE id NOT IN (SELECT id FROM `translations` WHERE `language_code` = "fra") LIMIT 1;

PR #244, opened 2016-01-27
Include a new phase in the pipeline and improved logging.

ruschein:


cpp/ixtheo_marc_pipeline.sh
```diff
@@ -45,29 +45,36 @@ marc_grep Normdaten-"${date}".mrc 'if "001" == ".*" extract *' marc_xml \
...
+echo "*** Phase $P ***" >> "${log}"
 create_child_refs.sh TitelUndLokaldaten-"${date}".xml ÜbergeordneteTitelUndLokaldaten-"${date}".xml >> "${log}" 2>&1
 add_child_refs ÜbergeordneteTitelUndLokaldaten-"${date}".xml \
                ÜbergeordneteTitelUndLokaldaten-post-phase"$P"-"${date}".xml \
                child_refs child_titles >> "${log}" 2>&1
 
-# Phase 2:
-P=2
+# Phase 3:
+P=3
```
oobenland: It would be easier with ((++P)) so you can swap phases without changing numbers

PR #323, opened 2016-03-10
First version that may have all the required functionality.

ruschein:


cpp/merge_differential_and_full_marc_updates.cc
```diff
@@ -374,54 +369,213 @@ void AppendFileOrDie(const std::string &append_target, const std::string &append
...
+
+    const std::string old_name_suffix("." + std::to_string(apply_count - 1));
+    std::string title_marc_basename, superior_marc_basename, normdata_marc_basename;
+    GetBasenamesOrDie(&title_marc_basename, &superior_marc_basename, &normdata_marc_basename, old_name_suffix);
+
+    const std::string new_name_suffix("." + std::to_string(apply_count));
+    std::string diff_filename;
+
+    // Update the title data:
+    std::string diff_filename_pattern("diff_.*a001.raw");
```
oobenland: Files with a001.raw and b001.raw need to be combined in to a single a001.raw file

PR #352, opened 2016-04-01
Automatically set GOROOT and GOPATH.

oobenland:


go/Makefile
```diff
@@ -1,6 +1,10 @@
 .PHONY: all
+export GOROOT=$(shell go env GOROOT)
+export GOPATH=${shell readlink -f "."}
 
 all:
+	${info GOROOT="${GOROOT}"}
+	${info GOPATH="${GOPATH}"}
```
ruschein: Please remove the two lines echoing the values of the exported variables.

go/Makefile
```diff
@@ -1,6 +1,10 @@
 .PHONY: all
+export GOROOT=$(shell go env GOROOT)
+export GOPATH=${shell readlink -f "."}
```
ruschein: Please use the long form of the "-f" flag which is "--canonicalize".

PR #476, opened 2016-08-01
Ruschein

ruschein:


cpp/marc_grep.cc
```diff
@@ -389,14 +408,16 @@ void FieldGrep(const unsigned max_records, const unsigned sampling_rate, const s
         xml_writer->openTag("marc:collection",
```
oobenland: Use MarcXmlWriter to avoid xmlns and xsi attributes.

PR #484, opened 2016-08-15
Riedl

jriedl:


cpp/add_synonyms.cc
```diff
@@ -0,0 +1,204 @@
...
+            Error("Need at least one output field");
+
+        unsigned num_of_entries(primary_tags_and_subfield_codes.size());
+
+        if (synonym_tags_and_subfield_codes.size() != num_of_entries or input_tags_and_subfield_codes.size() != num_of_entries
+            or output_tags_and_subfield_codes.size() != num_of_entries)
+            Error("Number of fields in all field specifications must be identical");
+             
+        // Set up the array of synonym lists
+        std::map<std::string, std::string> *synonym_map = new std::map<std::string, std::string>[num_of_entries];
```
ruschein: Why not create this variable on the stack instead of on the heap?

cpp/add_synonyms.cc
```diff
@@ -0,0 +1,204 @@
...
+            Error("Need at least one primary field");
+
+        if (unlikely(StringUtil::Split(AUTHORITY_DATA_SYNONYM_SPEC, ":", &synonym_tags_and_subfield_codes) < 1))
+            Error("Need at least one synonym field");
+
+        if (unlikely(StringUtil::Split(TITLE_DATA_PRIMARY_SPEC, ":", &input_tags_and_subfield_codes) < 1))
+            Error("Need at least one input field");
+
+        if (unlikely(StringUtil::Split(TITLE_DATA_UNUSED_FIELDS_FOR_SYNONYMS, ":", &output_tags_and_subfield_codes) < 1))
+            Error("Need at least one output field");
```
ruschein: If splitting on a single character it is more efficient to use a character instead of a string.

cpp/add_synonyms.cc
jriedl: For variable length arrays this has been the standard approach for years. Size of the data is not known beforehand and comparatively large, so using the stack would have to rely on the map implementation to store data on the heap anyway which would level performance advantages.

cpp/add_synonyms.cc
jriedl: The character variant of StringUtil::Join uses the string version internally.

cpp/add_synonyms.cc
```diff
@@ -0,0 +1,204 @@
...
+
+        unsigned num_of_entries(primary_tags_and_subfield_codes.size());
```
ruschein: Should this be const?

PR #497, opened 2016-08-25
Obenland

oobenland:


cpp/translate_chainer.cc
```diff
@@ -136,7 +136,7 @@ void ParseTranslationsDbToolOutputAndGenerateNewDisplay(const std::string &outpu
 
 
 // The first call should only provide the 3-letter language code.
```
ruschein: Inappropriate comment.

cpp/translate_chainer.cc
```diff
@@ -149,19 +149,30 @@ void InitialCall(const std::multimap<std::string, std::string> &cgi_args) {
 
 
 // A standard call should provide "index", "language_code", "category" and "translation".
-void RegularCall(const std::multimap<std::string, std::string> &cgi_args) {
+void Insert(const std::multimap<std::string, std::string> &cgi_args) {
```
ruschein: Ditto.

cpp/translate_chainer.cc
```diff
@@ -149,19 +149,30 @@ void InitialCall(const std::multimap<std::string, std::string> &cgi_args) {
...
-    ParseTranslationsDbToolOutputAndGenerateNewDisplay(output, language_code);
+
+void EmitHeader() {
+    std::cout << "Content-Type: text/html; charset=utf-8\r\n\r\n";
+}
+
+void EmitRedirectHeader(const std::string language_code) {
+    std::cout << "Status: 302 Found\r\n";
+    std::cout << "Location: /cgi-bin/translate_chainer?language_code=" << language_code << "\r\n\r\n";
 }
```
ruschein: I assume we need to emit the same headers as in the other case or you need to call that function, too.

PR #516, opened 2016-09-05
Riedl

jriedl:


cpp/add_synonyms.cc
```diff
@@ -73,57 +73,71 @@ void ExtractSynonyms(File * const norm_data_marc_input, const std::set<std::stri
...
+                // First case: Look up synonyms only in one category
+                if (i < synonym_maps.size()) {
+                    const auto &synonym_map(synonym_maps[i]);
+                    synonyms = synonym_map.find(StringUtil::Join(primary_values, ','))->second;
+                }
+                
+                // Second case: Look up synonyms in all categories
+                else {
+                    for (auto &sm : synonym_maps)
+                       synonyms += sm[StringUtil::Join(primary_values, ',')];
```
ruschein: Do we not need some kind of separator between the synonyms?

PR #633, opened 2016-11-17
Fixing constant for length calculation

jriedl:


cpp/add_synonyms.cc
```diff
@@ -179,7 +179,7 @@ void ProcessRecord(MarcRecord * const record, const std::vector<std::map<std::st
                     if (indicator2 > 9)
                         Error ("Currently cannot handle synonyms with total length greater than " + std::to_string(9 * (MarcRecord::MAX_FIELD_LENGTH - 4)) + '\n');
```
ruschein: Please do not use magic constants.  Either use a named constant or add an explanatory comment.

PR #1203, opened 2017-10-06
NOT BACKWARDS COMPATIBLE!!! changes for TuFind

mtrojan-ub:


cpp/lib/src/FileUtil.cc
```diff
@@ -942,6 +942,30 @@ void AppendStringToFile(const std::string &path, const std::string &text) {
         Error("in FileUtil::AppendStringToFile: failed to append data to \"" + path + "\"!");
 }
 
+void AppendFileToFile(const std::string &path_source, const std::string &path_target) {
+    std::string string_source = "";
+    if (unlikely(!ReadString(path_source, &string_source)))
```
ruschein: Please always use the "word" versions of the boolean operators, in this case it would be "not".

mtrojan-ub:
dont pull yet, create new request in a few days

PR #1478, opened 2018-02-02
 zts_client uses FormatHandler classes for JSON,MARCXML,MARC21. 

mtrojan-ub:


cpp/lib/src/Zotero.cc
```diff
@@ -0,0 +1,86 @@
...
+
+
+// We try to be unique for the machine we're on.  Beyond that we may have a problem.
+std::string Zotero::GetNextSessionId() {
+    static unsigned counter;
+    static uint32_t uuid[4];
+    if (unlikely(counter == 0))
+        ::uuid_generate(reinterpret_cast<unsigned char *>(uuid));
+    ++counter;
+    return "ub_tools_zts_client_" + StringUtil::ToString(uuid[0]) + StringUtil::ToString(uuid[1])
```
ruschein: Why do you only use 3 of the 4 values of uuid?

cpp/new_journal_alert.cc
```diff
@@ -101,7 +101,7 @@ std::string GetIssueId(const JSON::ObjectNode * const doc_obj) {
     const std::string id(JSON::LookupString("/id", doc_obj, /* default_value = */ ""));
     if (unlikely(id.empty()))
         logger->error("Did not find 'id' node in JSON tree!");
-    
+
```
ruschein: Please use the ERROR macro here.

cpp/new_journal_alert.cc
```diff
@@ -111,7 +111,7 @@ std::string GetIssueTitle(const std::string &id, const JSON::ObjectNode * const
     const auto issue_title(JSON::LookupString("/title", doc_obj, /* default_value = */ NO_AVAILABLE_TITLE));
     if (unlikely(issue_title == NO_AVAILABLE_TITLE))
         logger->warning("No title found for ID " + id + "!");
-    
+
     return issue_title;
 }
```
ruschein: WARNING

php/public/zts/index.php
```diff
@@ -76,7 +76,7 @@ function updateRuntime(seconds) {
                     ?>
                 </select>
             </td>
-            <td>MARC21 currently disabled due to problems with zts_client</td>
+            <td>MARC21 and MARCXML are custom. All others are generated by zotero (JSON = native zotero API format).</td>
```
ruschein: Zotero is a name and should be capitalised.

cpp/lib/src/Zotero.cc
mtrojan-ub: blindness in the morning? caffeine hasn't kicked in yet? ;-)

PR #1633, opened 2018-03-20
first attempt on TextUtil::ExtractTextFromTei

mtrojan-ub:


mtrojan-ub:
simple version, but should be enough for TEI files used by http://idb.ub.uni-tuebingen.de/diglit

cpp/lib/src/TextUtil.cc
```diff
@@ -163,6 +165,38 @@ std::string ExtractTextFromHtml(const std::string &html, const std::string &init
 }
 
 
+std::string ExtractTextFromTei(const std::string &tei) {
+    std::string extracted_text;
+
+    StringDataSource *source(new StringDataSource(tei));
+    SimpleXmlParser<StringDataSource> *parser(new SimpleXmlParser<StringDataSource>(source));
```
ruschein: Please fix the above two lines.  They're both memory leaks.

PR #1687, opened 2018-04-13
Ruschein

ruschein:


cpp/find_latest_journal_year.cc
```diff
@@ -37,7 +37,10 @@ namespace {
 
 
 [[noreturn]] void Usage() {
-    std::cerr << "Usage: " << ::progname << " journal_list marc_data report\n";
+    std::cerr << "Usage: " << ::progname << " journal_list marc_data report\n"
+              << "       The journal_list file must be a CSV file with 3 columns and a header row.\n"
```
mtrojan-ub: ... and which separator + quote characters must be used?
(microsoft office excel CSV files often use semicolon as separator)

PR #1814, opened 2018-06-08
installer creating ub_tools database

mtrojan-ub:
- CreateDatabases will later on also be able to create/extend vufind database
- "IF NOT EXISTS" was removed, problems with mysql syntax on INDEX statements. Also, correctness is not guaranteed anyway because existing tables will not be upgraded to current structure
- not sure how to handle root password, should be empty during installation process. what if mysql has already been installed previously?

### DIFF
--- a/cpp/data/installer/ub_tools.sql
+++ b/cpp/data/installer/ub_tools.sql
@@ -1,5 +1,5 @@
 -- The sizes here must be in sync with the constants defined in rss_aggregator.cc!
-CREATE TABLE IF NOT EXISTS rss_aggregator (
+CREATE TABLE rss_aggregator (
     item_id VARCHAR(100) NOT NULL,
     item_url VARCHAR(512) NOT NULL,
     title_and_or_description TEXT NOT NULL,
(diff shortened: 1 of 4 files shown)


cpp/lib/src/Zotero.cc
```diff
@@ -477,6 +477,16 @@ std::pair<unsigned, unsigned> MarcFormatHandler::processRecord(const std::shared
         } else if (key_and_node.first == "journalAbbreviation") {
             new_record.insertField("773", { { 'p', JSON::JSONNode::CastToStringNodeOrDie(key_and_node.first,
                                                                                          key_and_node.second)->getValue() } });
+        } else if (key_and_node.first == "extra") {
+            // comment field, can contain anything, even DOI's (e.g. for a webpage which has no DOI field)
+            const std::string extra(JSON::JSONNode::CastToStringNodeOrDie(key_and_node.first,
+                                                                          key_and_node.second)->getValue());
+            static RegexMatcher * const doi_matcher(RegexMatcher::RegexMatcherFactory("^DOI:\\s*([0-9a-zA-Z./]+)$"));
+            if (doi_matcher->matched(extra)) {
+                CreateSubfieldFromStringNode(key_and_node, "856", 'u', &new_record);
```
mtrojan-ub: check BSZ definition before merging this!
- loc.gov: 856u => https://www.loc.gov/marc/holdings/hd856.html
- bsz: 024 $2doi => https://wiki.bsz-bw.de/lib/exe/fetch.php?media=v-team:daten:datendienste:pica_marc21_rda_titel.xls

PR #1821, opened 2018-06-11
zotero_cgi: unified processing for rss + crawling tasks using zts_har…

mtrojan-ub:
…vester

### DIFF
--- a/cpp/zotero_cgi.cc
+++ b/cpp/zotero_cgi.cc
@@ -224,125 +224,9 @@ std::string PrepareMapsDirectory(const std::string &orig_directory, const std::s
 }
 
 
-void UpdateProgress(std::string progress) {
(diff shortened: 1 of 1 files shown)


cpp/zotero_cgi.cc
```diff
@@ -447,40 +297,40 @@ void ProcessCrawlingAction(const std::multimap<std::string, std::string> &cgi_ar
...
         std::cout << "<tr><td>ERROR</td><td>Exitcode: " + std::to_string(exit_code) + "</td></tr>\r\n";
-    }
 
     // use <pre> instead of nl2br + htmlspecialchars
     std::cout << "<tr><td>CLI output:</td><td><pre>" + output + "</td></tr>\r\n";
     std::cout << "</table>\r\n";
 }
 
 
+void ProcessDownloadAction(const std::multimap<std::string, std::string> &cgi_args) {
```
ruschein: cgi_args should be a const reference.

PR #1864, opened 2018-06-20
Zotero::TranslationServer::Web now checking for 4XX/5XX/9XX HTTP resp…

mtrojan-ub:
…onse codes

=> or should this be done in Downloader library itself???

### DIFF
--- a/cpp/lib/src/Zotero.cc
+++ b/cpp/lib/src/Zotero.cc
@@ -813,18 +813,6 @@ std::pair<unsigned, unsigned> Harvest(const std::string &harvest_url, const std:
         return std::make_pair(0, 0);
     }
 
-    // 500 => internal server error (e.g. error in translator))
-    if (response_code == 500) {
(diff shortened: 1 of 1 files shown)


cpp/lib/src/Zotero.cc
```diff
@@ -84,6 +84,19 @@ std::string GetNextControlNumber() {
 namespace TranslationServer {
 
 
+bool CheckResponseCode(unsigned response_code, const std::string &response_body, std::string * const error_message) {
```
ruschein: Please rename this to ResponseCodeIndicatesSuccess

cpp/lib/src/Zotero.cc
```diff
@@ -84,6 +84,19 @@ std::string GetNextControlNumber() {
...
+    const std::string response_code_string(StringUtil::ToString(response_code));
+    const char response_code_category = response_code_string[0];
```
ruschein: Please use the constructor notation using parentheses.

PR #1896, opened 2018-06-27
Fix regression that  triggered a segfault when parsing IniFiles with …

shadeMe:
…global entries


mtrojan-ub:
ruschein+me already fixed this + already merged to master

PR #1976, opened 2018-07-18
Converted to the new MARC API.

ruschein:


cpp/merge_differential_and_full_marc_updates.cc
```diff
@@ -62,15 +62,32 @@ incremental_authority_dump = (?:WA-MARCcomb-sekkor)-(\d{6}).tar.gz
 namespace {
 
 
+void Usage() {
```
shadeMe: No return attribute missing.

PR #1981, opened 2018-07-18
added php-soap dependency (VuFind5 preparation)

mtrojan-ub:


mtrojan-ub:
required by zendframework/zend-soap 2.7.0
https://github.com/zendframework/zend-soap/blob/master/CHANGELOG.md

PR #2136, opened 2018-10-01
zotero_cgi: assorted fixes

shadeMe:


cpp/lib/src/FileUtil.cc
```diff
@@ -51,9 +51,11 @@ AutoTempFile::AutoTempFile(const std::string &path_prefix, bool automatically_re
 {
     std::string path_template(path_prefix + "XXXXXX");
     const int fd(::mkstemp(const_cast<char *>(path_template.c_str())));
-    if (fd == -1)
+    if (fd == -1) {
```
ruschein: We don't usually use blocks for single statements.

cpp/zotero_cgi.cc
```diff
@@ -47,8 +47,8 @@ std::string zts_url(Zotero::TranslationServer::GetUrl());
 const std::string ZTS_HARVESTER_CONF_FILE("/usr/local/ub_tools/cpp/data/zts_harvester.conf");
 const std::vector<std::pair<std::string,std::string>> OUTPUT_FORMAT_IDS_AND_EXTENSIONS {
     // custom formats
-    { "marcxml", "xml" },
-    { "marc21", "mrc" },
+    { "marcxml", "marc-xml" },
+    { "marc21", "marc-21" },
```
mtrojan-ub: this means that we would generate files with the extension ".marc-xml" and ".marc-21"???

cpp/lib/src/FileUtil.cc
shadeMe: The scope was added to wrap the multi-line statement.

cpp/zotero_cgi.cc
shadeMe: No, that's just internal format string used to specify the output format. zotero_cgi's output files don't have any extension (as the filenames are automatically generated) - The above format string is passed to the Zotero/MarcFormatHandler, which then instantiates the correct writer.

PR #2146, opened 2018-10-04
new LobidUtil

mtrojan-ub:


cpp/lib/src/LobidUtil.cc
```diff
@@ -0,0 +1,120 @@
...
+    JSON::Parser json_parser(downloader.getMessageBody());
+    if (not (json_parser.parse(&root_node)))
+        LOG_ERROR("failed to parse returned JSON: " + json_parser.getErrorMessage() + "(input was: "
+                  + downloader.getMessageBody() + ")");
+
+    const std::shared_ptr<const JSON::ObjectNode> root_object(JSON::JSONNode::CastToObjectNodeOrDie("root", root_node));
+    url_to_lookup_result_cache.emplace(url, root_object);
+
+    if (not allow_multiple_results and root_object->getIntegerValue("totalItems") > 1) {
+        LOG_WARNING("found more than one result");
```
ruschein: Should you also issue this warning if you perform a cache lookup?

cpp/lib/src/LobidUtil.cc
```diff
@@ -0,0 +1,120 @@
...
+    LOG_DEBUG(url);
+    const auto url_and_lookup_result(url_to_lookup_result_cache.find(url));
+    if (url_and_lookup_result != url_to_lookup_result_cache.end())
+        return url_and_lookup_result->second;
+
+    Downloader downloader(url);
+    if (downloader.anErrorOccurred())
+        LOG_ERROR(downloader.getLastErrorMessage());
+
+    std::shared_ptr<JSON::JSONNode> root_node(nullptr);
```
ruschein: I don't think that you need to initialise this variable.  Also, you should move it down a line to be as close to its initial use as possible.

cpp/lib/src/LobidUtil.cc
mtrojan-ub: yes, because:
- cache does not store information about "allow_multiple_results", so there might not have a warning on the previous operation
- it's good to have a log entry at the right position even if it fails multiple times

PR #2210, opened 2018-10-19
Fix for TueFind issue #425.

ruschein:


mtrojan-ub:
see ubtue/tuefind#425

PR #2212, opened 2018-10-19
DbConnection and related bugfixes, switch ControlNumberGuesser backend to SQLite

shadeMe:
Also simplified the ControlNumberGuesser API. Performance is slightly worse than the kyoto cabinet impl but still manageable (10-15 mins for full database creation). The generated database file is ~500MB, which is an order of magnitude smaller than the older impl's.


shadeMe:
Added support for DOIs. The match rate is still abysmal at ~40%, though.

PR #2225, opened 2018-10-26
Fulltext optimizations for OpenDigi TEI files (+DatenProbleme 773g changes)

mtrojan-ub:


cpp/lib/src/SmartDownloader.cc
```diff
@@ -189,56 +188,22 @@ bool DiglitSmartDownloader::downloadDocImpl(const std::string &url, const TimeLi
                                             std::string * const document, std::string * const http_header_charset,
                                             std::string * const error_message)
 {
+    std::string url_improved(url);
```
ruschein: improved_url would be a better name.

cpp/lib/src/MediaTypeUtil.cc
```diff
@@ -114,6 +115,24 @@ std::string GetMediaType(const std::string &document, const bool auto_simplify)
 }
 
 
+std::string GetMediaType(const std::string &document, std::string * const subtype, const bool auto_simplify) {
+    std::string media_type = GetMediaType(document, auto_simplify);
```
ruschein: No initalisations with "=", please.  Also, I think "media_type" can be const.

cpp/lib/src/MediaTypeUtil.cc
```diff
@@ -114,6 +115,24 @@ std::string GetMediaType(const std::string &document, const bool auto_simplify)
...
+
+    if (media_type == "text/xml") {
+        XMLParser parser(document, XMLParser::XML_STRING);
+
+        std::map<std::string, std::string> attrib_map;
+        XMLParser::XMLPart part;
+        if (parser.skipTo(XMLParser::XMLPart::OPENING_TAG, "", &part)) {
+            if (part.data_ == "TEI")
```
ruschein: This strategy seems a bit risky.  Can't we use a more robust way of determining the subtype?

cpp/lib/src/MediaTypeUtil.cc
```diff
@@ -114,6 +115,24 @@ std::string GetMediaType(const std::string &document, const bool auto_simplify)
...
+                *subtype = "tei";
+        }
+    }
```
ruschein: Should this be "tei" or "TEI"?

cpp/lib/src/SmartDownloader.cc
```diff
@@ -189,56 +188,22 @@ bool DiglitSmartDownloader::downloadDocImpl(const std::string &url, const TimeLi
...
+    if (RegexMatcher::Matched("/diglit/", url))
+        url_improved = StringUtil::ReplaceString("/diglit/", "/opendigi/", &url_improved);
+    if (RegexMatcher::Matched("/opendigi/", url_improved))
+        url_improved = url_improved + "/ocr";
```
ruschein: Please use +=, it's more efficient and also easier for humans to parse.

cpp/lib/src/MediaTypeUtil.cc
mtrojan-ub: The returned media type is lowercase (e.g. "text/xml"), so the subtype should be lowercase as well.

cpp/lib/src/MediaTypeUtil.cc
mtrojan-ub: now also checking xmlns

cpp/lib/src/SmartDownloader.cc
mtrojan-ub: done

cpp/lib/src/SmartDownloader.cc
mtrojan-ub: renamed

PR #2236, opened 2018-11-05
Added a vector-flavoured version of StartsWith.

ruschein:


cpp/lib/src/RegexMatcher.cc
```diff
@@ -173,12 +174,20 @@ bool RegexMatcher::matched(const std::string &subject, std::string * const err_m
...
                            std::string * const err_msg, size_t * const start_pos, size_t * const end_pos)
 {
+    const auto regex_and_matcher(regex_to_matcher_map.find(regex));
+    if (regex_and_matcher != regex_to_matcher_map.cend())
+        return regex_and_matcher->second->matched(subject, err_msg, start_pos, end_pos);
+
     RegexMatcher * const matcher(RegexMatcher::RegexMatcherFactory(regex, err_msg, options));
     if (matcher == nullptr)
         LOG_ERROR("Failed to compile pattern \"" + regex + "\": " + *err_msg);
+    regex_to_matcher_map[regex] = matcher;
```
mtrojan-ub: The cache key should not only contain the regex, but also the options.
See [old version](https://github.com/ubtue/ub_tools/commit/5fd03d54b082ce0abd9e9293f6227d07500486b1#diff-c606510165749882d9d48c057b17767e)

PR #2272, opened 2018-11-13
Fix for DatenProbleme issue 119.

ruschein:


mtrojan-ub:
see ubtue/DatenProbleme#119

PR #2322, opened 2018-11-23
MARC & Zotero pipeline fixes

shadeMe:


cpp/zts_harvester_delivery_pipeline.sh
```diff
@@ -82,6 +88,7 @@ declare -a dest_filepaths
 
 StartPhase "Harvest URLs"
 LOGGER_FORMAT=no_decorations,strip_call_site \
+BACKTRACE=1 \
```
ruschein: I don't think this has any effect because BACKTRACE needs to be exported to achieve the desired effect.

PR #2374, opened 2018-12-06
A better Elasticsearch class??

ruschein:


cpp/lib/src/Elasticsearch.cc
```diff
@@ -18,235 +19,58 @@
...
+    downloader_params.authentication_username_ = username_;
+    downloader_params.authentication_password_ = password_;
+    downloader_params.ignore_ssl_certificates_ = ignore_ssl_certificates_;
+    downloader_params.additional_headers_.push_back("Content-Type: application/json");
 
-    if (data != nullptr) {
-        params.additional_headers_.push_back("Content-Type: application/json");
-        LOG_DEBUG(data->toString());
-    }
+    const Url delete_url(host_ + "/" + index_ + "/_delete_by_query");
```
shadeMe: If I recall correctly, there should be an update document endpoint as well. We could use that in lieu of deleting the existing entry and reinserting it. Will probably be faster.

cpp/lib/src/Elasticsearch.cc
```diff
@@ -18,235 +19,58 @@
...
-            const std::shared_ptr<const JSON::ObjectNode> hit_object(JSON::JSONNode::CastToObjectNodeOrDie("hit", hit_node));
-            Document document;
-            document.id_ = hit_object->getStringValue("_id");
-            if (hit_object->hasNode("_source")) {
-                const std::shared_ptr<const JSON::ObjectNode> fields_object(hit_object->getObjectNode("_source"));
-                document.fields_ = JSONToFields(fields_object);
-            }
-            documents[document.id_] = document;
-        }
+    const Url insert_url(host_ + "/" + index_ + "/_delete_by_query");
```
shadeMe: Incorrect endpoint.

PR #2442, opened 2019-01-09
mySQLSyncMultipleResults

mtrojan-ub:


cpp/lib/src/DbConnection.cc
```diff
@@ -487,6 +503,17 @@ void DbConnection::mySQLSelectDatabase(const std::string& database_name) {
 }
 
 
+void DbConnection::mySQLSyncMultipleResults() {
+    int next_result_exists;
+    do {
+        MYSQL_RES * const result_set (::mysql_store_result(&mysql_));
```
ruschein: Extraneous space.

cpp/lib/src/DbConnection.cc
```diff
@@ -255,9 +255,12 @@ bool DbConnection::queryFile(const std::string &filename) {
     if (not FileUtil::ReadString(filename, &statements))
         LOG_ERROR("failed to read \"" + filename + "\"!");
 
-    if (type_ == T_MYSQL)
-        return query(StringUtil::TrimWhite(&statements));
-    else {
+    if (type_ == T_MYSQL) {
+        bool query_result(query(StringUtil::TrimWhite(&statements)));
```
ruschein: Missing const.

cpp/lib/src/DbConnection.cc
```diff
@@ -487,6 +503,17 @@ void DbConnection::mySQLSelectDatabase(const std::string& database_name) {
 }
 
 
+void DbConnection::mySQLSyncMultipleResults() {
+    int next_result_exists;
+    do {
+        MYSQL_RES * const result_set (::mysql_store_result(&mysql_));
+        if (result_set)
```
ruschein: Should be compared against nullptr.

cpp/lib/src/DbConnection.cc
ruschein: Missing const.

PR #2592, opened 2019-03-08
A new pipleline tool for canon law references.

ruschein:


cpp/pipelines/augment_canones_references.cc
```diff
@@ -0,0 +1,210 @@
...
+#include "MARC.h"
+#include "StringUtil.h"
+#include "UBTools.h"
+#include "util.h"
+
+
+namespace {
+
+
+void ParseRanges(const std::string &ppn, const std::string &ranges, unsigned * const range_start, unsigned * const range_end) {
```
shadeMe: It would be a good idea to write a bit of documentation for the different magic numbers being used. The info on the wiki page doesn't directly translate to the code here (at least at first blush).

cpp/pipelines/augment_canones_references.cc
```diff
@@ -0,0 +1,210 @@
...
+            *range_start *= 10000;
+            *range_end   *= 10000;
+        }
+
+        return;
+    } else if (parts.size() == 2) {
+        if (unlikely(not StringUtil::ToUnsigned(parts[0], &canones)))
+            LOG_ERROR("bad codex canones \"" + parts[0] + "\"! (PPN: " + ppn + ")");
+
+        const auto dash_pos(parts[1].find('-'));
```
shadeMe: Is it possible to have ranges for both Pars? As in, _CIC83 1044,1-2,2-3_ or _CIC83 1044,3,2-5_. Those two cases aren't currently evaluated by the code.

PR #2598, opened 2019-03-08
Zotero: add new config entry to remove fields

shadeMe:
zts_harvester: update config for 'Journal of church and state'

refs ubtue/DatenProbleme#183

### DIFF
--- a/cpp/data/zts_harvester.conf
+++ b/cpp/data/zts_harvester.conf
@@ -115,6 +115,8 @@ author_gnd_lookup_query_params = dateOfBirth:19*
 #
 #                                         Example: "0247 \037a%hdl%\0372hdl" where %hdl% is replaced with the contents of the extra data
 #                                                                            named 'hdl'
+#   remove_field_XXX                    - Used to remove MARC fields from the generated MARC record when their content matches a RegEx pattern.
+#                                         Example: remove_field_520 = ^Editorial$
 #   exclusion_filter_XXX                - Used to exclude records from being written to the MARC output file when their content matches a RegEx
 #                                         pattern.
 #
@@ -1712,6 +1714,9 @@ zotero_type               = RSS
 zotero_group              = IxTheo
 zotero_url                = https://academic.oup.com/rss/site_5471/3332.xml
 zotero_expected_languages = eng
+remove_field_520          = .   # remove abstracts unconditionally
+zotero_review_regex       = "^(B|b)ook (R|r)eviews$"
+exclusion_filter_650a     = "^(F|f)eatures$"
 
 [International bulletin of mission research]
 zeder_id                  = 479
--- a/cpp/data/zts_harvester.conf
+++ b/cpp/data/zts_harvester.conf
@@ -116,7 +116,10 @@ author_gnd_lookup_query_params = dateOfBirth:19*
 #                                         Example: "0247 \037a%hdl%\0372hdl" where %hdl% is replaced with the contents of the extra data
 #                                                                            named 'hdl'
 #   remove_field_XXX                    - Used to remove MARC fields from the generated MARC record when their content matches a RegEx pattern.
-#                                         Example: remove_field_520 = ^Editorial$
+#                                         The suffix 'XXX' must have the following format: <tag><subfield_code>. Removes the entire field
+#                                         from the record.
+#
+#                                         Example: remove_field_520a = ^Editorial$
 #   exclusion_filter_XXX                - Used to exclude records from being written to the MARC output file when their content matches a RegEx
 #                                         pattern.
 #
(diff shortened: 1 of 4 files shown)


cpp/lib/include/Zotero.h
```diff
@@ -214,6 +214,7 @@ struct SiteParams {
     std::vector<std::string> additional_fields_;
     std::vector<std::string> non_standard_metadata_fields_;
     std::unordered_map<std::string, std::unique_ptr<RegexMatcher>> field_exclusion_filters_;
+    std::unordered_map<std::string, std::unique_ptr<RegexMatcher>> field_removal_filters_;
```
ruschein: Do we really anticipate that many map entries that we need an unordered_map?  The same question should be asked of the field_exclusion_filters_ field.

cpp/lib/src/Zotero.cc
```diff
@@ -703,6 +703,17 @@ void MarcFormatHandler::generateMarcRecord(MARC::Record * const record, const st
 
     if (not node_parameters.journal_name_.empty())
         record->insertField("JOU", { { 'a', node_parameters.journal_name_ } });
+
+    // remove any fields that match removal patterns
+    for (const auto &filter : site_params_->field_removal_filters_) {
+        auto field(filter.first);
+        if (record->fieldOrSubfieldMatched(filter.first, filter.second.get())) {
+            if (field.length() != 3)
```
ruschein: This test belongs to the place where we load the config and then we should probably abort hard if we don't meet our requirement.  (Also maybe we should also support subfields like 520a etc.)

cpp/lib/src/Zotero.cc
```diff
@@ -703,6 +703,17 @@ void MarcFormatHandler::generateMarcRecord(MARC::Record * const record, const st
...
+                field.erase(3);
```
ruschein: This probably does not do what you expect.  I use resize() for this purpose.

cpp/zts_harvester.cc
```diff
@@ -111,6 +111,13 @@ void ReadGenericSiteAugmentParams(const IniFile &ini_file, const IniFile::Sectio
 
             site_params->field_exclusion_filters_.insert(std::make_pair(field_name,
                                                          std::unique_ptr<RegexMatcher>(RegexMatcher::RegexMatcherFactoryOrDie(entry.value_))));
+        } else if (StringUtil::StartsWith(entry.name_, "remove_field_")) {
+            const auto field_name(entry.name_.substr(__builtin_strlen("remove_field_")));
+            if (field_name.length() != MARC::Record::TAG_LENGTH && field_name.length() != MARC::Record::TAG_LENGTH + 1)
```
ruschein: Please use "and" and not "&&".

cpp/data/zts_harvester.conf
```diff
@@ -115,6 +115,8 @@ author_gnd_lookup_query_params = dateOfBirth:19*
 #
 #                                         Example: "0247 \037a%hdl%\0372hdl" where %hdl% is replaced with the contents of the extra data
 #                                                                            named 'hdl'
+#   remove_field_XXX                    - Used to remove MARC fields from the generated MARC record when their content matches a RegEx pattern.
```
ruschein: You forgot to document that subfield references are also supported.

cpp/lib/include/Zotero.h
ruschein: Looking at the usage of this field: Do we really need any kind of map or would a vector of pairs suffice?

cpp/lib/src/Zotero.cc
```diff
@@ -703,6 +703,17 @@ void MarcFormatHandler::generateMarcRecord(MARC::Record * const record, const st
 
     if (not node_parameters.journal_name_.empty())
         record->insertField("JOU", { { 'a', node_parameters.journal_name_ } });
+
+    // remove any fields that match removal patterns
+    for (const auto &filter : site_params_->field_removal_filters_) {
+        auto field(filter.first);
```
ruschein: Please rename this to tag_and_optional_subfield_code.

cpp/lib/src/Zotero.cc
```diff
@@ -703,6 +703,17 @@ void MarcFormatHandler::generateMarcRecord(MARC::Record * const record, const st
...
+        if (record->fieldOrSubfieldMatched(filter.first, filter.second.get())) {
+            if (field.length() != 3)
+                field.erase(3);
+            record->erase(field);
```
ruschein: How about deleting the lines that are supposed to optionally delete the subfield code and rewriting this as "record->erase(tag_and_optional_subfield_code.substr(0, MARC::Record::TAG_LENGTH);"?

PR #2615, opened 2019-03-15
rewrite_keywords_and_authors_from_authority_data: added title ppn to …

mtrojan-ub:
…log messages if authority lookup fails


cpp/pipelines/rewrite_keywords_and_authors_from_authority_data.cc
```diff
@@ -56,7 +56,8 @@ MARC::Record::const_iterator GetFirstPrimaryField(const MARC::Record& authority_
 
 
 bool GetAuthorityRecordFromPPN(const std::string &bsz_authority_ppn, MARC::Record * const authority_record,
-                               MARC::Reader * const authority_reader, const std::unordered_map<std::string, off_t> &authority_offsets)
+                               MARC::Reader * const authority_reader, const std::unordered_map<std::string, off_t> &authority_offsets,
+                               const MARC::Record * const record)
```
shadeMe: The `record` parameter should be a const reference.

PR #2762, opened 2019-05-23
Fixe Tuefind issue 408.

ruschein:


mtrojan-ub:
refs ubtue/tuefind#408

PR #2835, opened 2019-07-03
Zotero MARC fixes and changes

shadeMe:
refs ubtue/DatenProbleme#229, ubtue/DatenProbleme#228, ubtue/DatenProbleme#233

### DIFF
--- a/cpp/lib/src/Zotero.cc
+++ b/cpp/lib/src/Zotero.cc
@@ -644,12 +644,12 @@ void MarcFormatHandler::generateMarcRecord(MARC::Record * const record, const st
     if (not superior_ppn.empty())
         _773_subfields.appendSubfield('w', "(DE-627)" + superior_ppn);
 
-    // 773g, example: "52(2018), 1, S. 1-40" => <volume>(<year>), <issue>, S. <pages>
+    // 773g, example: "52 (2018), 1, S. 1-40" => <volume>(<year>), <issue>, S. <pages>
     const bool _773_subfields_iaxw_present(not _773_subfields.empty());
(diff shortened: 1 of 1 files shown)


cpp/lib/src/Zotero.cc
```diff
@@ -547,7 +547,9 @@ void MarcFormatHandler::generateMarcRecord(MARC::Record * const record, const st
         record->insertField("007", "cr|||||");
 
     // Authors/Creators (use reverse iterator to keep order, because "insertField" inserts at first possible position)
+    // The first creator is always saved in the "100" field, all following creators go into the 700 field
     const std::string creator_tag((node_parameters.creators_.size() == 1) ? "100" : "700");
```
ruschein: Please remove "creator_tag" as it seems to be unused.

cpp/lib/src/Zotero.cc
```diff
@@ -547,7 +547,9 @@ void MarcFormatHandler::generateMarcRecord(MARC::Record * const record, const st
...
+    int num_creators_left(node_parameters.creators_.size());
```
ruschein: Nit: please use unsigned integer types for counts.

cpp/lib/src/Zotero.cc
```diff
@@ -580,7 +587,7 @@ void MarcFormatHandler::generateMarcRecord(MARC::Record * const record, const st
     // Abstract Note
     const std::string abstract_note(node_parameters.abstract_note_);
```
ruschein: Mirco optimisation: please use a reference here.

PR #2839, opened 2019-07-04
Added support for balcklisting of URL's when crawling.

ruschein:


cpp/zts_harvester.cc
```diff
@@ -71,6 +71,10 @@ void ReadGenericSiteAugmentParams(const IniFile &ini_file, const IniFile::Sectio
     if (not extraction_regex.empty())
         site_params->extraction_regex_.reset(RegexMatcher::RegexMatcherFactoryOrDie(extraction_regex));
 
+    const auto banned_url_regex_str(bundle_reader.zotero(section.getSectionName()).value(JournalConfig::Zotero::BANNED_URL_REGEX, ""));
```
shadeMe: The variable `section_name` can be used here.

PR #2988, opened 2019-09-17
New tool.

ruschein:


cpp/find_bib_ref_candidates_in_titles.cc
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+    return false;
+}
+
+
+std::vector<std::string> TokenizeText(std::string text) {
+    TextUtil::NormaliseDashes(&text);
+
+    std::vector<std::string> tokens;
```
shadeMe: Why not use `StringUtil::Spilit` here?

PR #3022, opened 2019-10-09
removed deprecated system_monitor

mtrojan-ub:


mtrojan-ub:
old data on network share FID-Entwicklung\system-monitor should also be removed after merging this pull request

PR #3102, opened 2019-11-07
Probperly order and combine subfield references.

ruschein:


cpp/marc_tools/marc_filter.cc
```diff
@@ -802,11 +802,38 @@ unsigned TestAndConvertCount(char ***argvp) {
...
+    std::vector<std::string> coalesced_specs;
+    coalesced_specs.reserve(subfield_specs->size());
+
+    std::string current_tag;
+
+    for (const auto &subfield_spec : *subfield_specs) {
+        const std::string tag(subfield_spec.substr(0, MARC::Record::TAG_LENGTH));
+        if (tag != current_tag) {
+            coalesced_specs.emplace_back(subfield_spec);
+            current_tag = subfield_spec.substr(0, MARC::Record::TAG_LENGTH);
```
shadeMe: std::swap() could be used here.

cpp/marc_tools/marc_filter.cc
```diff
@@ -802,11 +802,38 @@ unsigned TestAndConvertCount(char ***argvp) {
...
+        } else {
+            for (const char subfield_code : subfield_spec.substr(MARC::Record::TAG_LENGTH)) {
+                if (std::strchr(coalesced_specs.back().c_str() + MARC::Record::TAG_LENGTH, subfield_code) == nullptr)
```
shadeMe: Any reason you're not using std::string::find() here?

ruschein: It's because I wanted to skip over the first TAG_LENGTH characters w/o creating a new temporary string object with substr.

PR #3138, opened 2019-11-26
First stab at approximating the format described in tuefind issue #994.

ruschein:


mtrojan-ub:
... wait after next live system update on thursday before merge

PR #3222, opened 2020-01-09
Increase the VARCHAR max length to 768 in the codebase, patch deliver…

shadeMe:
…ed_marc_records table

### DIFF
--- a/cpp/data/installer/ub_tools.sql
+++ b/cpp/data/installer/ub_tools.sql
@@ -1,5 +1,5 @@
--- under CentOS we still have MariaDB 5, which has a limitation of 767 bytes for keys.
--- this means e.g. for VARCHAR with utf8mb4, we can use at most a VARCHAR(191)!
+-- under CentOS we still have MariaDB 5, which has a limitation of 768 bytes for index keys.
+-- this means we need to use a prefix length <= 768 when creating an index on a particular VARCHAR column.
 -- Also, the default collating sequence is a Swedish one.  This leads to aliasing problems for characters with
 -- diacritical marks => we need to override it and use utf8mb4_bin.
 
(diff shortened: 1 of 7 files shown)


cpp/data/installer/ub_tools.sql
```diff
@@ -41,20 +41,20 @@ CREATE TABLE delivered_marc_records (
     hash VARCHAR(40) NOT NULL,
     zeder_id VARCHAR(10) NOT NULL,
     delivered_at TIMESTAMP NOT NULL DEFAULT NOW(),
-    journal_name VARCHAR(191) NOT NULL,
-    main_title VARCHAR(191) NOT NULL,
+    journal_name VARCHAR(1000) NOT NULL,
+    main_title VARCHAR(1000) NOT NULL,
```
mtrojan-ub: what about the field length in other similar tables? e.g.
- delivered_marc_records_superior_info.title
- metadata_presence_tracer.journal_name and metadata_field_name
- rss_aggregator.item_id

cpp/data/installer/ub_tools.sql
```diff
@@ -41,20 +41,20 @@ CREATE TABLE delivered_marc_records (
...
     publication_year CHAR(4) DEFAULT NULL,
     volume CHAR(40) DEFAULT NULL,
```
mtrojan-ub: not part of this change so far, but: why is CHAR used here instead of VARCHAR? (for volume, issue, pages)

cpp/lib/include/SqlUtil.h
```diff
@@ -46,7 +46,7 @@ namespace SqlUtil {
 extern const std::string DATETIME_RANGE_MIN;
 extern const std::string DATETIME_RANGE_MAX;
 // As supported by MariaDB on CentOS
-constexpr size_t VARCHAR_UTF8_MAX_LENGTH(191);
+constexpr size_t VARCHAR_UTF8_MAX_LENGTH(768);
```
mtrojan-ub: should this be renamed to VARCHAR_UTF8_MAX_INDEX_LENGTH?

PR #3342, opened 2020-03-04
Harvester fixes

shadeMe:


cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -246,37 +246,37 @@ bool ZoteroItemMatchesExclusionFilters(const Util::HarvestableItem &download_ite
 void ConvertZoteroItemToMetadataRecord(const std::shared_ptr<JSON::ObjectNode> &zotero_item,
                                        MetadataRecord * const metadata_record)
 {
-    metadata_record->item_type_ = zotero_item->getStringValue("itemType");
-    metadata_record->title_ = zotero_item->getOptionalStringValue("title");
-    metadata_record->short_title_ = zotero_item->getOptionalStringValue("shortTitle");
-    metadata_record->abstract_note_ = zotero_item->getOptionalStringValue("abstractNote");
-    metadata_record->publication_title_ = zotero_item->getOptionalStringValue("publicationTitle");
+    metadata_record->item_type_ = HtmlUtil::StripHtmlTags(zotero_item->getStringValue("itemType"));
```
ruschein: Consider an inline helper that combines HtmlUtil::StripHtmlTags with getStringValue defined just above this function.

cpp/lib/src/Locale.cc
```diff
@@ -27,32 +27,33 @@
...
-    if (locale == nullptr)
-        throw std::runtime_error("in Locale::Locale: can't get locale for requested category!");
+    if (restore_) {
+        if (::uselocale(old_locale_) == 0)
+            LOG_ERROR("failed to restore thread locale");
+    }
 
-    return locale;
+    if (is_valid_)
+        ::freelocale(new_locale_);
```
ruschein: Do you need to call ::uselocale(0) too?

cpp/validate_harvested_records.cc
```diff
@@ -208,7 +199,7 @@ void SendEmail(const std::string &email_address, const std::string &message_subj
                           /* use_ssl = */ true, /* use_authentication = */ true));
 
     if (reply_code >= 300)
-        LOG_ERROR("failed to send email, the response code was: " + std::to_string(reply_code));
+        LOG_WARNING("failed to send email, the response code was: " + std::to_string(reply_code));
```
ruschein: hmm

cpp/validate_harvested_records.cc
```diff
@@ -229,30 +220,43 @@ int Main(int argc, char *argv[]) {
...
-        if (journal_name_and_info == journal_name_to_info_map.end()) {
-            first_record = true;
-            JournalInfo new_journal_info;
-            LoadFromDatabaseOrCreateFromScratch(&db_connection, journal_name, &new_journal_info);
-            journal_name_to_info_map[journal_name] = new_journal_info;
-            journal_name_and_info = journal_name_to_info_map.find(journal_name);
-        }
+        bool validation_failed(false);
+
+        // Intentionally true to allow early breaking
```
ruschein: I suggest getting rid of the "while (true)" and the validation_failed variable and instead using the following code at the bottom:

    valid_records_writer->write(record);
    return;
validation_failed:
    delinquent_records_writer->write(record);

and replacing occurences of

   validation_failed = true;
   break;

with

    goto validation_failed;

cpp/validate_harvested_records.cc
```diff
@@ -229,30 +220,43 @@ int Main(int argc, char *argv[]) {
...
+        while (true) {
+            const auto journal_name(record.getSuperiorTitle());
+            if (journal_name.empty()) {
+                LOG_WARNING("Record w/ control number \"" + record.getControlNumber() + "\" is missing a superior title!");
```
ruschein: Wouldn't a LOG_ERROR be reasonable here because it would be some logic error if we attempted to harvest an article for a journal for which we have no title?

PR #3362, opened 2020-03-18
Riedl

jriedl:


cpp/update_full_text_db.cc
```diff
@@ -276,6 +277,7 @@ bool ProcessRecordUrls(MARC::Record * const record, const unsigned pdf_extractio
                        const bool use_only_open_access_links, const bool store_pdfs_as_html,
                        const bool use_separate_entries_per_url = false,
                        const bool include_all_tocs = false,
+                       const bool only_pdf_fulltexts = false,
                        const bool skip_reviews = false) {
```
ruschein: When there are continuation lines the opening brace should go on a line by itself.

PR #3363, opened 2020-03-20
Ruschein

ruschein:


cpp/convert_json_to_marc.cc
```diff
@@ -0,0 +1,493 @@
...
+      subfield_codes_to_json_tags_(subfield_codes_to_json_tags), subfield_codes_to_prefixes_(subfield_codes_to_prefixes),
+      subfield_codes_to_fixed_subfields_(subfield_codes_to_fixed_subfields), json_tag_(json_tag),
+      field_contents_prefix_(field_contents_prefix), required_(required)
+{
+    if (not overflow_tag_.empty() and repeat_field_)
+        LOG_ERROR("field \"" + name_ + "\" can't have both, an over flow tag and being a repeat field!");
+
+    if (subfield_codes_to_json_tags_.empty() and json_tag_.empty())
+        LOG_ERROR("field \"" + name_
+                  + "\" neither a mapping to the contents of a control field nor the contents of data subfields has been specified!");
```
shadeMe: "... nor have the contents of data subfields specified" would be more correct.

cpp/convert_json_to_marc.cc
```diff
@@ -0,0 +1,493 @@
+/** \file    convert_json_to_marc.cc
```
shadeMe: It would be a good idea to upload a test JSON file that has examples for all the cases covered by the tool. Would make it easier to follow what is actually being done by the tool. You could put in the the _tests_ folder.

cpp/convert_json_to_marc.cc
```diff
@@ -0,0 +1,493 @@
...
+        }
+    }
+
+    if (escaped)
+        LOG_ERROR("bad regex pattern to item type mapping: \"" + pattern_and_type + "\"!");
+}
+
+
+MARC::Record::BibliographicLevel MapTypeStringToBibliographicLevel(const std::string &item_type) {
+    if (item_type == "monograph")
```
shadeMe: Since this is supposed to be a generic tool, wouldn't it be better to have the item type strings saved in the config file? A different JSON schema might use other identifiers for those bibliographic levels.

cpp/convert_json_to_marc.cc
```diff
@@ -0,0 +1,493 @@
...
+}
+
+
+JSONNodeToBibliographicLevelMapper::JSONNodeToBibliographicLevelMapper(const std::string &item_type_tag, const std::string &item_type_map)
+    : json_tag_(item_type_tag)
+{
+    default_ = MARC::Record::UNDEFINED;
+
+    std::vector<std::string> pattens_and_types;
+    StringUtil::Split(item_type_map, '|', &pattens_and_types);
```
shadeMe: The pipe character is used in regular expressions. You'll need to find another delimiter or make sure that it doesn't split inside patterns.

cpp/convert_json_to_marc.cc
```diff
@@ -0,0 +1,493 @@
...
+    if (subfield_codes_to_json_tags_.empty() and json_tag_.empty())
+        LOG_ERROR("field \"" + name_
+                  + "\" neither a mapping to the contents of a control field nor the contents of data subfields has been specified!");
+}
+
+
+class JSONNodeToBibliographicLevelMapper {
+    std::string json_tag_;
+    MARC::Record::BibliographicLevel default_;
+    std::vector<std::pair<RegexMatcher *, MARC::Record::BibliographicLevel>> regex_to_bibliographic_level_map;
```
shadeMe: It'd be best practice to use the std::unique_ptr to wrap the raw Regexmatcher pointer here.

cpp/convert_json_to_marc.cc
```diff
@@ -0,0 +1,493 @@
...
+    if (not global_section.hasEntry("item_type_tag") and global_section.hasEntry("item_type_map"))
+        LOG_ERROR("Global section \"has item_type_map\" but not \"item_type_tag\"!");
+
+    std::string item_type_tag, item_type_map;
+    if (global_section.hasEntry("item_type_tag")) {
+        item_type_tag = global_section.getString("item_type_tag");
+        item_type_map = global_section.getString("item_type_map");
+    }
+
+    *json_node_to_bibliographic_level_mapper = new JSONNodeToBibliographicLevelMapper(item_type_tag, item_type_map);
```
shadeMe: Same as before - Best not pass around bare pointers. Even if it doesn't leak right now (I imagine the rationale is that the program is has a very short (half-)life, and therefore memory leaks aren't problematic), it just might if the program's scope changes in the future.

cpp/convert_json_to_marc.cc
```diff
@@ -0,0 +1,493 @@
...
+            const auto tag(section.getString("tag", ""));
+            if (tag.empty())
+                LOG_ERROR("missing tag in section \"" + section_name + "\" in \"" + ini_file.getFilename() + "\"!");
+            if (tag.length() != MARC::Record::TAG_LENGTH)
+                LOG_ERROR("invalid tag \"" + tag + "\" in section \"" + section_name + "\" in \"" + ini_file.getFilename() + "\"!");
+
+            std::vector<std::pair<char, std::string>> subfield_codes_to_json_tags, subfield_codes_to_prefixes, subfield_codes_to_fixed_subfields;
+            for (const auto section_entry : section) {
+                if (StringUtil::StartsWith(section_entry.name_, "add_fixed_subfield_")) {
+                    if (section_entry.name_.length() != __builtin_strlen("add_fixed_subfield_?"))
```
shadeMe: Perhaps add a comment here to explain that the '?' character is a placeholder, or just add 1 to the prefix's length.

cpp/convert_json_to_marc.cc
```diff
@@ -0,0 +1,493 @@
...
+void GenerateSingleMARCRecordFromJSON(const std::shared_ptr<const JSON::ObjectNode> &object,
+                                      const JSONNodeToBibliographicLevelMapper &json_node_to_bibliographic_level_mapper,
+                                      const std::vector<FieldDescriptor> &field_descriptors, MARC::Writer * const marc_writer)
+{
+    std::string control_number;
+    const auto descriptor_for_field_001(std::find_if(field_descriptors.begin(), field_descriptors.end(),
+                                                     [](const FieldDescriptor &descriptor){ return descriptor.tag_ == "001"; }));
+    if (descriptor_for_field_001 != field_descriptors.end()) {
+        control_number = object->getOptionalStringValue(descriptor_for_field_001->json_tag_);
+        if (not control_number.empty())
```
shadeMe: Shouldn't this be `if (control_number.empty())`?

cpp/convert_json_to_marc.cc
ruschein: Please have a look at the new wording...

cpp/convert_json_to_marc.cc
ruschein: I added a file in the tests folder.

cpp/convert_json_to_marc.cc
ruschein: There is only a fixed set of possible bibliographic levels as per the MARC-21 standard.  There are still various other possible problems with this tool not being fully generic.  I decided to not go much beyond what is needed for the data from CORE for now.  We may never need to support anything else and if we do it will be easier to decide what has to be extended.  Also we don't have any outside users of our tool so we don't care about how generic it is.

cpp/convert_json_to_marc.cc
ruschein: I now support backslash escapes for colons and vertical bars.

cpp/convert_json_to_marc.cc
ruschein: I added a destructor for the JSONNodeToBibliographicLevelMapper class.

cpp/convert_json_to_marc.cc
ruschein: No, because if we have no data for a field we don't just want to shove the prefix into the field.  It's not the same as a default value.

cpp/convert_json_to_marc.cc
ruschein: Done.

PR #3364, opened 2020-03-20
New simplified tool for importing/updating entries from Zeder

shadeMe:


cpp/lib/include/ZoteroHarvesterZederInterop.h
```diff
@@ -0,0 +1,58 @@
...
+extern const std::map<Config::JournalParams::IniKey, std::function<std::string(const Zeder::Entry &, const Zeder::Flavour)>> INI_KEY_TO_ZEDER_RESOLVER_MAP;
+
+
+// Uses the above defined maps to retrieve the corresponding value of the INI key.
+std::string GetJournalParamsIniValueFromZederEntry(const Zeder::Entry &zeder_entry, const Zeder::Flavour zeder_flavour,
+                                                   const Config::JournalParams::IniKey ini_key);
+
+Zeder::Flavour GetZederInstanceForJournal(const Config::JournalParams &journal_params);
+Zeder::Flavour GetZederInstanceFromMarcRecord(const MARC::Record &record);
+
```
ruschein: One too many blank lines.

cpp/lib/src/ZoteroHarvesterConfig.cc
```diff
@@ -265,15 +265,44 @@ const std::map<JournalParams::IniKey, std::string> JournalParams::KEY_TO_STRING_
...
+    { "zotero_type",                HARVESTER_OPERATION },
+    { "zotero_delivery_mode",       UPLOAD_OPERATION },
+    { "online_ppn",                 ONLINE_PPN },
+    { "print_ppn",                  PRINT_PPN },
+    { "online_issn",                ONLINE_ISSN },
+    { "print_issn",                 PRINT_ISSN },
+    { "zotero_strptime_format",     STRPTIME_FORMAT_STRING },
+    { "zotero_update_window",       UPDATE_WINDOW },
+    { "zotero_review_regex",        REVIEW_REGEX },
+    { "zotero_expected_languages",  EXPECTED_LANGUAGES },
```
ruschein: I apologize for nitpicking.  The minimal number of spaces in such a tabular arrangement should be 1.  Also we usually like up the closing braces on each line, too.  See also the previous map initialization above.

cpp/lib/src/ZoteroHarvesterZederInterop.cc
```diff
@@ -0,0 +1,178 @@
...
+
+
+namespace ZoteroHarvester {
+
+
+namespace ZederInterop {
+
+
+const std::map<Config::JournalParams::IniKey, std::string> INI_KEY_TO_ZEDER_COLUMN_MAP {
+    { Config::JournalParams::IniKey::NAME, "tit" },
```
ruschein: Same as the other other tabular structures.

cpp/zeder_to_zotero_importer.cc
```diff
@@ -0,0 +1,432 @@
...
+
+
+namespace {
+
+
+using namespace ZoteroHarvester;
+
+
+[[noreturn]] void Usage() {
+    std::cerr << "Usage: " << ::progname << " [options] config_file_path mode zeder_flavour zeder_ids fields_to_update\n"
```
ruschein: Please use the Usage function from util.h.

cpp/zeder_to_zotero_importer.cc
```diff
@@ -0,0 +1,432 @@
...
+              << "\tfields_to_update                Comma-separated list of the following fields to update: \n"
+              << "\t                                \tONLINE_PPN, PRINT_PPN, ONLINE_ISSN, PRINT_ISSN, EXPECTED_LANGUAGES, ENTRY_POINT_URL, UPLOAD_OPERATION, UPDATE_WINDOW.\n"
+              << "\t                                Ignored when importing entries (all importable fields will be imported).\n\n";
+    std::exit(EXIT_FAILURE);
+}
+
+
+struct CommandLineArgs {
+    enum Mode { INVALID, IMPORT, UPDATE };
+
```
ruschein: We use public to separate the type declarations from the field declarations.  Alternatively you can declare the enum field on the same line as the Mode type.

cpp/zeder_to_zotero_importer.cc
```diff
@@ -0,0 +1,432 @@
...
+    Mode mode_;
+    Zeder::Flavour zeder_flavour_;
+    std::set<unsigned> zeder_ids_;
+    std::set<Config::JournalParams::IniKey> fields_to_update_;
+public:
+    explicit CommandLineArgs() : overwrite_on_import_(false), mode_(INVALID) {}
+};
+
+
+void ParseCommandLineArgs(int * const argc, char *** const argv, CommandLineArgs * const commandline_args) {
```
ruschein: Why are you you passing in the addresses of the parameters to main instead of a const copy for argc and a const reference for argv?

cpp/zeder_to_zotero_importer.cc
```diff
@@ -0,0 +1,432 @@
...
+
+    const std::string zeder_flavour((*argv)[1]);
+    --*argc, ++*argv;
+
+    commandline_args->zeder_flavour_ = Zeder::ParseFlavour(zeder_flavour);
+
+    const std::string zeder_id_list((*argv)[1]);
+    --*argc, ++*argv;
+
+    std::set<std::string> buffer;
```
ruschein: Would zeder_ids be a better name than buffer?

cpp/zeder_to_zotero_importer.cc
```diff
@@ -0,0 +1,432 @@
...
+        for (const auto &id_str : buffer)
+            commandline_args->zeder_ids_.emplace(StringUtil::ToUnsigned(id_str));
+    } else if (commandline_args->mode_ == CommandLineArgs::Mode::IMPORT)
+        LOG_ERROR("cannot import all Zeder entries at once");
+
+    if (commandline_args->mode_ == CommandLineArgs::Mode::IMPORT)
+        return;
+
+    static const std::map<std::string, Config::JournalParams::IniKey> ALLOWED_INI_KEYS {
+        { "ENTRY_POINT_URL",    Config::JournalParams::ENTRY_POINT_URL },
```
ruschein: Same formatting comment as with other static map initializations.

cpp/zeder_to_zotero_importer.cc
shadeMe: It's used for multiple argument types, therefore went with a generic name.

cpp/zeder_to_zotero_importer.cc
shadeMe: Because I modify those values while parsing the args.

PR #3368, opened 2020-03-26
Mkannan

shadeMe:


cpp/lib/src/ZoteroHarvesterZederInterop.cc
```diff
@@ -147,15 +147,28 @@ std::string GetJournalParamsIniValueFromZederEntry(const Zeder::Entry &zeder_ent
 }
 
 
-Zeder::Flavour GetZederInstanceForJournal(const Config::JournalParams &journal_params) {
-    if (journal_params.group_ == "IxTheo" or journal_params.group_ == "ixtheo")
+static Zeder::Flavour GetZederInstanceFromGroupName(const std::string &group_name, const std::string &journal_name = "") {
+    if (group_name == "IxTheo" or group_name == "ixtheo")
```
ruschein: How about strcasecmp instead.  That's robust and future proof.

PR #3376, opened 2020-04-01
Riedl

jriedl:


cpp/lib/src/FullTextCache.cc
```diff
@@ -72,7 +72,22 @@ bool FullTextCache::entryExpired(const std::string &id, std::vector<std::string>
...
+    return true;
+}
+
+
+bool FullTextCache::singleUrlExpired(const std::string &id, const std::string &url) {
+    Entry entry;
+    if (not getEntry(id, &entry))
+        return true;
+    const time_t now(std::time(nullptr));
+    if (entry.expiration_ == TimeUtil::BAD_TIME_T or now < entry.expiration_) {
```
ruschein, line 84: Probably a micro optimization (I Don't know how likely it is that entry.expiration_ == TimeUtil::BAD_TIME_T.) but, if you replace "now" on this line with "std::time(nullptr)" and delete the previous line, you may save some calls to std::time because of boolean short circuiting.

PR #3377, opened 2020-04-02
Ruschein

ruschein:


jriedl:
Doesn't the information given in 249 refer to an additional resource ("beigefügtes Werk") that cannot necessarily be used for the main entry?

ruschein:
The way I understood this is that the "beigefügtes Werk" has a separate record and that record does not have its own 245 field.

PR #3388, opened 2020-04-06
rss_aggregator: skip feed instead of crashing on exception

mtrojan-ub:


ruschein:
Please add a notification mechanism so that we don't silently loose feeds.

cpp/rss_aggregator.cc
```diff
@@ -334,9 +334,13 @@ int Main(int argc, char *argv[]) {
...
                 LOG_INFO("Processing section \"" + section_name + "\".");
-                const unsigned new_item_count(ProcessSection(one_shot, section, &downloader, &db_connection,
-                                                             DEFAULT_DOWNLOADER_TIME_LIMIT, DEFAULT_POLL_INTERVAL, ticks));
-                LOG_INFO("Downloaded " + std::to_string(new_item_count) + " new items.");
+                try {
+                    const unsigned new_item_count(ProcessSection(one_shot, section, &downloader, &db_connection,
+                                                                 DEFAULT_DOWNLOADER_TIME_LIMIT, DEFAULT_POLL_INTERVAL, ticks));
+                    LOG_INFO("Downloaded " + std::to_string(new_item_count) + " new items.");
+                } catch (const std::runtime_error &x) {
+                    LOG_WARNING("Error processing section \"" + section_name + "\": " + x.what());
```
ruschein, line 342: Should we also send an email here instead?  Or collect all these failures and send an email with priority 4 at the end?

PR #3396, opened 2020-04-07
We now send our own error email messages.

ruschein:


cpp/generate_subscription_packets.sh
```diff
@@ -24,7 +24,10 @@ function Usage() {
 }
 
 
-generate_subscription_packets /usr/local/var/lib/tuefind/relbib_packets.conf /tmp/relbib_bundles.out
+readonly email_address=$1
```
mtrojan-ub, line 27: There is no check if $1 is set. Also Usage() is never called. And if Usage runs into exit 1, the trap will be called, and SendEmail will have an error => so Usage should be checked at least before the trap statement, maybe even before no_problems_found.

cpp/rss_aggregator.cc
```diff
@@ -367,10 +327,71 @@ int Main(int argc, char *argv[]) {
...
+    const std::string PROXY(ini_file.getString("", "proxy", ""));
+    if (not PROXY.empty()) {
+        LOG_INFO("using proxy: " + PROXY);
+        params.proxy_host_and_port_ = PROXY;
+    }
+    Downloader downloader(params);
+
+    try {
+        return ProcessFeeds(one_shot, xml_output_filename, &ini_file, &db_connection, &downloader);
+    } catch (const std::exception &x) {
```
mtrojan-ub, line 388: Do we really want to catch std::exception or would std::runtime_error be better?

cpp/rss_aggregator.cc
```diff
@@ -367,10 +327,71 @@ int Main(int argc, char *argv[]) {
...
+
+    Downloader::Params params;
+    const std::string PROXY(ini_file.getString("", "proxy", ""));
+    if (not PROXY.empty()) {
+        LOG_INFO("using proxy: " + PROXY);
+        params.proxy_host_and_port_ = PROXY;
+    }
+    Downloader downloader(params);
+
+    try {
```
mtrojan-ub, line 386: should the try...catch be moved into the infinite for(;;) loop in ProcessFeeds, and should the treatment be different if one_shot is false?

cpp/rss_aggregator.cc
```diff
@@ -61,7 +63,7 @@ void SigHupHandler(int /* signum */) {
 
 
 [[noreturn]] void Usage() {
-    ::Usage("[--one-shot] [--config-file=config_file_path] [--process-name=new_process_name] xml_output_path\n"
+    ::Usage("[--one-shot] [--config-file=config_file_path] [--process-name=new_process_name] email_address xml_output_path\n"
```
mtrojan-ub, line 66: don't forget installer/server crontab

PR #3397, opened 2020-04-08
Add explicit parameter to skip DB updates

jriedl:


cpp/merge_differential_and_full_marc_updates_merge_superior.sh
```diff
@@ -59,10 +59,15 @@ function CreateArchiveWithMergedTitles() {
...
-KEEP_ITERMEDIATE_FILES=
+KEEP_INTERMEDIATE_FILES=
+SKIP_DB_UPDATES=
 if [[ $# > 1 ]]; then
     if [[ $1 == "--keep-intermediate-files" ]]; then
-        KEEP_ITERMEDIATE_FILES="--keep-intermediate-files"
+        KEEP_INTERMEDIATE_FILES="--keep-intermediate-files"
+        shift
+    fi
+    if [[ $1 == "--skip-db-updates" ]]; then
```
ruschein, line 69: This code looks like you can only specify --keep-intermediate-files or --skip-db-updates o/w you would have to use a loop here.  If I am correct the usage message should reflect this and look like [(--keep-intermediate-files|--skip-db-updates)]

cpp/pipelines/merge_print_and_online.cc
```diff
@@ -1060,12 +1060,19 @@ int Main(int argc, char *argv[]) {
...
 
     bool debug(false);
+    bool skip_db_updates(false);
     if (std::strcmp(argv[1], "--debug") == 0) {
         debug = true;
         --argc, ++argv;
         if (argc < 4)
             Usage();
     }
+    if (std::strcmp(argv[1], "--skip-db-updates") == 0) {
```
ruschein, line 1070: bool skip_db_updates(false); should go here.

cpp/merge_differential_and_full_marc_updates_merge_superior.sh
jriedl: "shift" should adjust the positional parameters so both can be used at the same time.

ruschein: While that is true, what happens when you call the script with only the --keep-intermediate-files flag and an email address?  Don't you run into line 72 even though the --skip-db-updates flag is optional?

PR #3406, opened 2020-04-16
zotero_harvester + zotero_cgi: Allow overriding of config settings

mtrojan-ub:


cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -290,7 +290,8 @@ class HarvestTask {
     FileUtil::AutoTempFile log_path_;
     std::unique_ptr<FileUtil::AutoTempFile> out_path_;
 public:
-    HarvestTask(const std::string &section, const std::string &output_format_id, const std::string &bsz_upload_group);
+    HarvestTask(const std::string &section, const std::string &output_format_id,
+                const std::string &bsz_upload_group, const std::string &config_overrides="");
```
ruschein, line 294: Please put spaces around the equal sign.

cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -323,6 +325,11 @@ HarvestTask::HarvestTask(const std::string &section, const std::string &output_f
     args.emplace_back("--output-filename=" + basename);
     args.emplace_back(ZTS_HARVESTER_CONF_FILE);
     args.emplace_back("JOURNAL");
+    if (not config_overrides.empty()) {
+        std::string arg("--config-overrides=" + config_overrides);
+        StringUtil::RemoveChars("\r", &arg);
+        args.emplace_back(arg);
```
ruschein, line 331: The 3 lines are nicer as a one-liner: args.emplace_back(StringUtil::RemoveChars("\n", "--config-overrides=" + config_overrides));

cpp/lib/include/ZoteroHarvesterConfig.h
```diff
@@ -215,7 +215,8 @@ class EnhancementMaps {
 void LoadHarvesterConfigFile(const std::string &config_filepath, std::unique_ptr<GlobalParams> * const global_params,
                              std::vector<std::unique_ptr<GroupParams>> * const group_params,
                              std::vector<std::unique_ptr<JournalParams>> * const journal_params,
-                             std::unique_ptr<IniFile> * const config_file = nullptr);
+                             std::unique_ptr<IniFile> * const config_file = nullptr,
+                             const IniFile::Section config_overrides=IniFile::Section());
```
ruschein, line 219: Spaces!

cpp/zotero_harvester.cc
```diff
@@ -55,7 +55,9 @@ using namespace ZoteroHarvester;
               << "\tSelection modes: UPLOAD, URL, JOURNAL\n"
               << "\t\tUPLOAD - Only those journals that have the specified upload operation (either LIVE or TEST) set will be processed.\n"
               << "\t\tURL - Only the specified URL is processed as a DIRECT harvester operation.\n"
-              << "\t\tJOURNAL - If no arguments are provided, all journals are processed. Otherwise, only those journals specified are processed.\n\n";
+              << "\t\tJOURNAL - If no arguments are provided, all journals are processed. Otherwise, only those journals specified are processed.\n"
```
ruschein, line 58: either "those journals that are specified" or "the specified journals".  Your choice!

cpp/zotero_harvester.cc
```diff
@@ -178,9 +188,12 @@ struct HarvesterConfigData {
...
 
 
-void LoadHarvesterConfig(const std::string &config_path, HarvesterConfigData * const harvester_config) {
+void LoadHarvesterConfig(const std::string &config_path, HarvesterConfigData * const harvester_config,
+                         const IniFile::Section &config_overrides=IniFile::Section())
+{
     Config::LoadHarvesterConfigFile(config_path, &harvester_config->global_params_,
-                                    &harvester_config->group_params_, &harvester_config->journal_params_);
+                                    &harvester_config->group_params_, &harvester_config->journal_params_,
+                                    nullptr, config_overrides);
```
ruschein, line 196: Please insert /* dummy_argument_name = */ in front of nullptr, w/ the appropriate dummy argument name.

cpp/zotero_harvester.cc
```diff
@@ -55,7 +55,9 @@ using namespace ZoteroHarvester;
               << "\tSelection modes: UPLOAD, URL, JOURNAL\n"
               << "\t\tUPLOAD - Only those journals that have the specified upload operation (either LIVE or TEST) set will be processed.\n"
               << "\t\tURL - Only the specified URL is processed as a DIRECT harvester operation.\n"
-              << "\t\tJOURNAL - If no arguments are provided, all journals are processed. Otherwise, only those journals specified are processed.\n\n";
+              << "\t\tJOURNAL - If no arguments are provided, all journals are processed. Otherwise, only those journals specified are processed.\n"
+              << "\t\t          If \"--config-overrides=\" are specified, it will be treated as raw text to override parts of the config file sections.\n"
```
shadeMe, line 59: The "--config-overrides" help text should be moved above with the rest of the optional parameters. The overrides should apply to all the selection modes, in my opinion. The current implementation will automatically take apply to the UPLOAD and JOURNAL modes as we're applying the overrides at the config parsing stage. The URL mode, however, will need more work. 

My suggestion is that we just mention in the help text that only UPLOAD and JOURNAL selection modes are affected by the overrides.

cpp/zotero_harvester.cc
```diff
@@ -152,7 +155,14 @@ void ParseCommandLineArgs(int * const argc, char *** const argv, CommandLineArgs
             return; // intentional early return
         }
         case CommandLineArgs::SelectionMode::JOURNAL:
-            commandline_args->selected_journals_.emplace(current_arg);
+            if (StringUtil::StartsWith(current_arg, "--config-overrides=")) {
```
shadeMe, line 158: This should be moved to the top, in the While loop with the rest of the optional params.

cpp/zotero_harvester.cc
```diff
@@ -152,7 +155,14 @@ void ParseCommandLineArgs(int * const argc, char *** const argv, CommandLineArgs
...
+                const std::string config_overrides(current_arg + __builtin_strlen("--config-overrides="));
+                FileUtil::AutoTempFile tempfile("/tmp/ATF", "", false);
```
shadeMe, line 160: Kind iffy about using a temp file here. The "right" way to go about it would be to modify the `IniFile` class to support the direct passing of file contents to its constructor, but it's more a "good-to-have" thing at this point.

cpp/zotero_harvester.cc
```diff
@@ -152,7 +155,14 @@ void ParseCommandLineArgs(int * const argc, char *** const argv, CommandLineArgs
...
+                FileUtil::WriteStringOrDie(tempfile.getFilePath(), config_overrides);
+                IniFile ini_tempfile(tempfile.getFilePath());
+                commandline_args->config_overrides_ = *ini_tempfile.begin();
+            } else
+                commandline_args->selected_journals_.emplace(current_arg);
```
shadeMe, line 165: This line shouldn't be in an `else` clause.

cpp/zotero_harvester.cc
```diff
@@ -178,9 +188,12 @@ struct HarvesterConfigData {
 };
 
 
-void LoadHarvesterConfig(const std::string &config_path, HarvesterConfigData * const harvester_config) {
+void LoadHarvesterConfig(const std::string &config_path, HarvesterConfigData * const harvester_config,
+                         const IniFile::Section &config_overrides=IniFile::Section())
```
shadeMe, line 192: Don't see why we need a default parameter here. We're always passing `commandline_args.config_overrides_` to this function.

cpp/zotero_harvester.cc
mtrojan-ub: I think this requires lots of changes and should be part of a different pull request.

PR #3411, opened 2020-04-16
zotero_cgi: Added new tab "Test URL"

mtrojan-ub:


cpp/data/zotero_cgi/index.html
```diff
@@ -123,7 +124,7 @@ <h1>Zotero Harvesting</h1>
                                 </td>
                                 <td>
                                     {IF all_journal_methods == "RSS"}<a href="#rss" onclick="TryRss('{all_journal_titles}')">try rss</a>{ENDIF}
-                                    {IF all_journal_methods == "DIRECT"}<a href="#rss" onclick="TryDirect('{all_journal_titles}')">try direct</a>{ENDIF}
+                                    {IF all_journal_methods == "DIRECT"}<a href="#direct" onclick="TryDirect('{all_journal_titles}')">try direct</a>{ENDIF}
                                     {IF all_journal_methods == "CRAWL"}<a href="#crawling" onclick="TryCrawling('{all_journal_titles}')" disabled class="disabled" title="Crawling disabled as discussed in meeting 2018-06-12">try crawl</a>{ENDIF}
```
shadeMe, line 128: I think we can remove the "Crawling disabled" link alt-text now.

cpp/data/zotero_cgi/index.html
```diff
@@ -123,7 +124,7 @@ <h1>Zotero Harvesting</h1>
                                 </td>
                                 <td>
                                     {IF all_journal_methods == "RSS"}<a href="#rss" onclick="TryRss('{all_journal_titles}')">try rss</a>{ENDIF}
-                                    {IF all_journal_methods == "DIRECT"}<a href="#rss" onclick="TryDirect('{all_journal_titles}')">try direct</a>{ENDIF}
+                                    {IF all_journal_methods == "DIRECT"}<a href="#direct" onclick="TryDirect('{all_journal_titles}')">try direct</a>{ENDIF}
```
shadeMe, line 127: We actually don't support DIRECT as a harvester operation in the config file anymore (see the new wiki entry for more details). Journals are either classified as RSS or CRAWL, i.e, if they have a RSS feed, we use the RSS operation. If not, we use the CRAWL operation. The legacy DIRECT can be emulated with the CRAWL operation by setting the crawl depth to zero (which is the default value).

cpp/data/zotero_cgi/index.html
shadeMe, line 128: I think we can remove the "Crawling disabled" link alt-text now.

cpp/data/zotero_cgi/index.html
```diff
@@ -179,12 +184,9 @@ <h1>Zotero Harvesting</h1>
                                         <option{IF selected_output_format_id == output_format_ids} selected="selected"{ENDIF}>{output_format_ids}</option>
                                     {ENDLOOP}
                                 </select>
+                                MARC21 and MARCXML are custom. All others are generated by Zotero (JSON = native Zotero API format).
```
shadeMe, line 187: This is no longer accurate - The harvester only supports MARC-XML as its output. In fact, we can disable the output format dropdown on the web interface.

cpp/data/zotero_cgi/index.html
mtrojan-ub: OK i will remove it, do the librarians already know this?

shadeMe: I doubt it. It's mentioned in the wiki, though.

mtrojan-ub: Removed, but i don't know whether there would really be an error if DIRECT is set in zotero_harvester.conf (seems like it is mapped via an array where DIRECT is still registered)

mtrojan-ub:
... still needs some work to be done, found 1-2 strange bugs
- "Default" seems to be replaced by other journal without any reason sometimes
- When page is loaded, a "Try URL" is started with an empty journal

PR #3412, opened 2020-04-17
Changed to use Zeder instead of MARC data.

ruschein:


cpp/generate_issn_to_titles_and_ppns_map.cc
```diff
@@ -42,45 +44,79 @@ std::string EscapeColons(const std::string &unescaped_string) {
...
 } // unnamed namespace
 
 
 int Main(int argc, char **argv) {
-    if (argc != 3)
-        ::Usage("marc_input mapfile_output");
+    if (argc != 2)
+        ::Usage("mapfile_output");
+
+
```
mtrojan-ub, line 109: Multiple blank lines here

cpp/generate_issn_to_titles_and_ppns_map.cc
```diff
@@ -42,45 +44,79 @@ std::string EscapeColons(const std::string &unescaped_string) {
...
+
+    const auto temp_file(FileUtil::OpenTempFileOrDie("/tmp/XXXXXX"));
```
mtrojan-ub, line 111: Replacing this by `FileUtil::AutoTempFile temp_file;` might be shorter

ruschein: No it isn't because AutoTempFile only gives us a std::string and we need a File *.

PR #3417, opened 2020-04-27
zeder_to_zotero_importer: zeder_newly_synced_entry

mtrojan-ub:
ToDo before enabling cronjob:
- rename "newly_synced_entry"? (e.g. "newly_imported_entry")
- add documentation for "newly_synced_entry" to zotero_harvester.conf
- prevent automatic delivery of entries marked as "newly_synced_entry"

### DIFF
--- a/cpp/zeder_to_zotero_importer.cc
+++ b/cpp/zeder_to_zotero_importer.cc
@@ -292,7 +292,7 @@ unsigned ImportZederEntries(const Zeder::EntryCollection &zeder_entries, Harvest
         bool new_section(false);
         if (existing_journal_section == nullptr) {
             existing_journal_section = harvester_config->addNewConfigSection(title);
-            existing_journal_section->insert("newly_synced_entry", "true");
+            existing_journal_section->insert(Config::JournalParams::GetIniKeyString(Config::JournalParams::ZEDER_NEWLY_SYNCED_ENTRY), "true");
             new_section = true;
         }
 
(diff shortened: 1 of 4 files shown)


cpp/zeder_to_zotero_importer.cc
```diff
@@ -48,7 +48,8 @@ using namespace ZoteroHarvester;
             "\t                                Special-case for updating: Use '*' to update all entries found in the config that belong to the Zeder flavour\n"
```
ruschein, line 48: I think you should add "or *" here.

cpp/zeder_to_zotero_importer.cc
```diff
@@ -270,13 +271,28 @@ unsigned ImportZederEntries(const Zeder::EntryCollection &zeder_entries, Harvest
...
         auto existing_journal_section(harvester_config->lookupConfig(zeder_id, zeder_flavour));
         if (existing_journal_section != nullptr and not overwrite) {
-            LOG_WARNING("couldn't import Zeder entry " + std::to_string(zeder_id) + " (" + title + "): already exists");
+            if (not autodetect_new_datasets) {
+                LOG_WARNING("couldn't import Zeder entry " + std::to_string(zeder_id) + " (" + title + "): already exists");
+                continue;
+            } else {
+                LOG_INFO("Skipping Zeder entry " + std::to_string(zeder_id) + " (" + title + "): already exists");
+                continue;
+            }
```
ruschein, line 280: As you continue in both cases, how about factoring that out and putting it below the else-case?  I'd also switch the if- and else- clauses because then you van drop the "not".

cpp/zeder_to_zotero_importer.cc
```diff
@@ -332,17 +348,20 @@ unsigned ImportZederEntries(const Zeder::EntryCollection &zeder_entries, Harvest
                 if (skip_entry) {
                     if (new_section)
                         harvester_config->removeConfigSection(title);
-                    continue;
+                    goto next_entry;
```
ruschein, line 351: Wouldn't a break also work here?

mtrojan-ub: almost, but if the entry is skipped we want ++num_entries_imported to be skipped as well

cpp/zeder_to_zotero_importer.cc
mtrojan-ub: I don't understand... it already says "Use '*'"?

cpp/zeder_to_zotero_importer.cc
mtrojan-ub: done!

PR #3418, opened 2020-04-29
Riedl

jriedl:


cpp/update_full_text_db.cc
```diff
@@ -327,12 +327,6 @@ bool ProcessRecordUrls(MARC::Record * const record, const unsigned pdf_extractio
         }
         if (not at_least_one_expired)
             return true;
-        else {
-            for (auto url_and_text_type(urls_and_text_types.begin()); url_and_text_type != urls_and_text_types.end(); ++url_and_text_type)
```
ruschein: Why not an implicit for-loop?

cpp/update_full_text_db.cc
```diff
@@ -327,12 +327,6 @@ bool ProcessRecordUrls(MARC::Record * const record, const unsigned pdf_extractio
...
-                std::cerr << record->getControlNumber() <<" " << url_and_text_type->url_ << '\n';
-        }
```
ruschein: Missing space, also please prefer character constants for single characters over string constants.  Furthermore, why don't you use LOG_? instead of streaming to std::cerr?

cpp/update_full_text_db.cc
jriedl: This is left over debug code that has been removed (cf. 7a22ef8).

cpp/update_full_text_db.cc
jriedl: See above.

cpp/update_full_text_db.cc
```diff
@@ -391,6 +397,18 @@ bool ProcessRecordUrls(MARC::Record * const record, const unsigned pdf_extractio
         already_present_text_types |= text_type;
     }
 
+    // If we are in only_fulltext_pdfs-mode each record without pdf-links would be downloaded on each create_full_text_db run
+    // only to be discarded because it does not match our rules
+    // So, if we are in this mode and no text has been stored to the cache insert a dummy entry to save time and bandwidth
```
ruschein, line 402: It's not "store to" but "store in".

PR #3423, opened 2020-05-04
Riedl

jriedl:


cpp/installer.cc
```diff
@@ -779,6 +790,19 @@ void ConfigureVuFind(const VuFindSystemType vufind_system_type, const OSSystemTy
 }
 
 
+void InstallFulltextBackendCronjobs() {
```
mtrojan-ub, line 793: upper/lowercase should be consistent, either alwas "Fulltext" or "FullText"

PR #3426, opened 2020-05-05
Riedl

jriedl:


docker/fulltext-centos/Dockerfile
```diff
@@ -4,7 +4,7 @@ MAINTAINER Johannes Riedl <johannes.riedl@uni-tuebingen.de>
 # add needed configuration files
 # (+simulate mounted /mnt/ZE020150)
 ADD mnt/ZE020150 /mnt/ZE020150
-
+ADD Elasticsearch.conf /usr/local/var/lib/tuelib/Elasticsearch.conf
```
mtrojan-ub, line 7: Does this really work? Directory might not exist before installer is executed

jriedl: It should exist: https://www.reddit.com/r/docker/comments/7t5953/dockerfile_add_to_a_nonexisting_directory/

PR #3428, opened 2020-05-05
Ruschein

ruschein:


cpp/convert_json_to_marc.cc
```diff
@@ -58,9 +58,10 @@ namespace {
 struct JournalTitleAndPPN {
     std::string journal_title_;
     std::string ppn_;
+    std::string online_issn_;
```
mtrojan-ub, line 61: either rename struct from JournalTitleAndPPN to JournalTitleAndPPNAndOnlineISSN or, before space & nerves run out, to something scalable/generic like JournalInfo

cpp/convert_json_to_marc.cc
```diff
@@ -99,10 +100,10 @@ void LoadISSNsToJournalTitlesAndPPNsMap(std::unordered_map<std::string, JournalT
         const std::string line(input->getline());
         if (line.empty())
             continue;
-        const std::vector<std::string> parts(SplitLineOnColons(line));
-        if (parts.size() != 3 or parts[0].empty() or parts[1].empty()) // ISSN and titles are required, PPN's are optionsl.
+        std::vector<std::string> parts(SplitLineOnColons(line));
+        if (parts.size() != 4 or parts[0].empty() or parts[1].empty()) // ISSN and titles are required, PPN's and online ISSN's are optionsl.
```
mtrojan-ub, line 104: What does "options1" in the comment mean?

cpp/journal_timeliness_checker.cc
```diff
@@ -94,13 +94,8 @@ int Main(int argc, char *argv[]) {
...
 
-        std::string journal_ppn(section.getString("online_ppn", ""));
-        if (journal_ppn.empty())
-            journal_ppn = section.getString("print_ppn", "");
-        if (journal_ppn.empty()) {
-            LOG_WARNING("no PPN found for \"" + journal_name + "\"!");
-            continue;
-        }
+        const std::string zeder_id(section.getString("zeder_id"));
+        const std::string zeder_instance(TextUtil::UTF8ToLower(section.getString("zotero_group")));
```
mtrojan-ub, line 98: zotero_group might also be RelBib, but i'm not sure if we should still allow that

mtrojan-ub:
I will merge, but journal_timeliness_checker should use ZoteroHarvesterConfig instead of loading the IniFile directly in the future

PR #3433, opened 2020-05-06
installer.cc: Fixed missing path separator in SSH_KEYS_DIR

mtrojan-ub:


mtrojan-ub:
Already fixed in #3432

PR #3435, opened 2020-05-07
Riedl

jriedl:


cpp/installer.cc
```diff
@@ -464,6 +464,12 @@ void InstallUBTools(const bool make_install, const OSSystemType os_system_type)
         ExecUtil::ExecOrDie(ExecUtil::LocateOrDie("mkdir"), { "-p", UBTools::GetTuelibPath() });
     }
 
+    // ..and /usr/local/var/log/tuefind
+    if (not FileUtil::Exists(UBTools::GetTuefindLogPath())) {
+        Echo("creating " + UBTools::GetTuefindLogPath());
+        ExecUtil::ExecOrDie(ExecUtil::LocateOrDie("mkdir"), { "-p", UBTools::GetTuefindLogPath() });
```
ruschein, line 470: Please use FileUtil::MakeDirectory instead of shelling out.  In general, please prefer our functions over shelling out.

cpp/installer.cc
```diff
@@ -717,20 +723,32 @@ void ConfigureSolrUserAndService(const VuFindSystemType system_type, const bool
 }
 
 
-void SetEnvironmentVariables(const std::string &vufind_system_type_string) {
+void PermanentlySetEnvironmentVariables(std::vector<std::pair<std::string, std::string>> &keys_and_values,
+                                        const std::string &script_path)
+{
+    std::string variables;
+    for (const auto &key_and_value : keys_and_values)
```
ruschein, line 730: I think the C++-17 syntax is much nicer: "for (const auto &[key, value] : keys_and_values)" etc.

cpp/installer.cc
```diff
@@ -717,20 +723,32 @@ void ConfigureSolrUserAndService(const VuFindSystemType system_type, const bool
...
-    std::string variables;
-    for (const auto &key_and_value : keys_and_values)
-        variables += "export " + key_and_value.first + "=" + key_and_value.second + "\n";
-
-    const std::string vufind_script_path("/etc/profile.d/vufind.sh");
-    FileUtil::WriteString(vufind_script_path, variables);
-    MiscUtil::LoadExports(vufind_script_path, /* overwrite = */ true);
+void SetFulltextEnvironmentVariables() {
+    // Currently only the IxTheo approach is supported
+    std::vector<std::pair<std::string, std::string>> keys_and_values {
```
ruschein, line 748: Can this be const?

cpp/lib/include/UBTools.h
```diff
@@ -30,5 +30,11 @@ inline std::string GetTuelibPath() {
     return "/usr/local/var/lib/tuelib/";
 }
 
+// \return A slash-terminated absolute path.
+inline std::string GetTuefindLogPath() {
```
ruschein, line 34: We never seem to use Tuefind in our code, I only found instances of TueFind.

PR #3465, opened 2020-05-28
zeder_to_zotero_importer

mtrojan-ub:


cpp/zeder_to_zotero_importer.cc
```diff
@@ -28,6 +28,11 @@
 #include "ZoteroHarvesterZederInterop.h"
 
 
+#ifdef LICENSE
```
ruschein, line 31: I think a comment is in order here explaining why we need this.

PR #3468, opened 2020-06-02
Allow our range-query tools to optionally output date ranges.

ruschein:


mtrojan-ub:
Are there also changes necessary in TueFind?
- https://github.com/ubtue/tuefind/blob/master/module/TueFindSearch/src/TueFindSearch/Backend/Solr/QueryBuilder.php
- https://github.com/ubtue/tuefind/blob/master/module/IxTheo/src/IxTheo/Search/Backend/Solr/QueryBuilder.php

PR #3471, opened 2020-06-03
Incron setup

jriedl:


cpp/data/pipeline_script_triggers
```diff
@@ -0,0 +1 @@
+/usr/local/var/tmp/ IN_CREATE,IN_DELETE /usr/local/bin/trigger_pipeline_script.sh $@$# $%
```
ruschein, line 1: Any reason why this is indented?

cpp/elasticsearch/extract_existing_fulltext_ids.sh
```diff
@@ -31,6 +31,18 @@ function GetDownloadStats() {
     echo -n $(echo "$response" | wc -c) >&2
     echo " bytes" >&2
     echo "------------------------------------" >&2
+    error=$(CheckError $response)
+    if [[ $(echo -n "$error" | wc --chars) -ne 0 ]]; then
```
ruschein, line 35: Should we also send an error email?

cpp/elasticsearch/extract_existing_fulltext_ids.sh
```diff
@@ -43,7 +55,7 @@ fi
 id_output_file=$1
 
 echo "Checking write access to output file..."
-> $id_output_file
+[ -w "$id_output_file" ] || (echo "Cannot write to to $id_output_file" && exit 1)
```
ruschein, line 58: Any reason you perform the echo and exit in a subshell?  The main shell will continue...

cpp/pipelines/enable_pipeline_triggers.sh
```diff
@@ -0,0 +1,25 @@
...
+
+if [[ -f /etc/redhat-release ]]; then
+    systemctl start incrond
+elif [[ -f /etc/debian_version ]]; then
+    systemctl start incron
+else
+    echo "Could not detect Linux distribution"
+    exit 1
+fi
+
```
ruschein, line 22: Why the blank lines at the end?

cpp/data/pipeline_script_triggers
jriedl: This seems to be a display problem. There is no indentation in the file.

cpp/elasticsearch/extract_existing_fulltext_ids.sh
jriedl: The script is normally called from within a cronjob that should fail if a subscript fails.

cpp/elasticsearch/extract_existing_fulltext_ids.sh
jriedl: Like this it is a oneliner. The main shell will also exit due to the -o errexit flag

ruschein: It's also an ugly hack and relies on the reader thinking about the errexit flag.  Generally we want code to be "obvious" and locally comprehensible, if possible.

PR #3479, opened 2020-06-10
Ruschein

ruschein:


cpp/lib/src/DbConnection.cc
```diff
@@ -436,6 +436,26 @@ std::string DbConnection::escapeString(const std::string &unescaped_string, cons
 }
 
 
+// Creates two hex digits for each character in "s".
+static std::string StringToHexString(const std::string &s) {
```
mtrojan-ub, line 440: Move to StringUtil?

cpp/lib/src/LocalDataDB.cc
```diff
@@ -0,0 +1,215 @@
...
+ */
+#include "LocalDataDB.h"
+#include "Compiler.h"
+#include "DbConnection.h"
+#include "StringUtil.h"
+#include "UBTools.h"
+
+
+LocalDataDB::LocalDataDB(const OpenMode open_mode) {
+    db_connection_ = new DbConnection(UBTools::GetTuelibPath() + "local_data.sq3" /* must be the same path as in fetch_marc_updates.py */,
```
mtrojan-ub, line 28: Path should be in a public variable

cronjobs/fetch_marc_updates.py
```diff
@@ -206,6 +206,7 @@ def Main():
     if complete_data_filenames is not None:
         download_cutoff_date = ExtractDateFromFilename(complete_data_filenames[0])
         downloaded_at_least_some_new_titles = True
+        util.Remove("/usr/local/var/lib/tuelib/local_data.sq3") # Must be the same path as in local_data.sq3
```
mtrojan-ub, line 209: re-use from public variable in LocalDataDb.h

cpp/lib/src/LocalDataDB.cc
```diff
@@ -0,0 +1,215 @@
...
+}
+
+
+static std::vector<std::string> BlobToLocalFieldsVector(const std::string &local_fields_blob, const std::string &title_ppn) {
+    std::vector<std::string> local_fields;
+
+    size_t processed_size(0);
+    do {
+        // Convert the 4 character hex string to the size of the following field contents:
+        const size_t field_contents_size(StringUtil::ToUnsignedLong(local_fields_blob.substr(processed_size, 4), 16));
```
mtrojan-ub, line 62: 4 is used in many functions and should be declared as a const in class context

cpp/lib/src/LocalDataDB.cc
```diff
@@ -0,0 +1,215 @@
...
+    auto result_set(db_connection_->getLastResultSet());
+    if (result_set.empty())
+        return false;
+    const auto row(result_set.getNextRow());
+    const auto title_ppn(row["title_ppn"]);
+
+    // 2. Retrieve the local data associated w/ the title PPN:
+    auto local_fields(getLocalFields(title_ppn));
+
+    // 3. Remove the local data associsted w/ the local PPN:
```
mtrojan-ub, line 200: typo "associsted"

cpp/lib/src/LocalDataDB.cc
```diff
@@ -0,0 +1,215 @@
...
+    const auto filtered_local_fields(RemoveLocalDataSet(local_ppn, local_fields));
+
+    // 4. Update our SQL tables:
+    db_connection_->queryOrDie("DELETE FROM local_ppns_to_title_ppns_map WHERE ocal_ppn="
```
mtrojan-ub, line 204: typo "ocal_ppn"

cpp/lib/src/LocalDataDB.cc
```diff
@@ -0,0 +1,215 @@
...
+        return;
+
+    db_connection_->queryOrDie("CREATE TABLE IF NOT EXISTS local_data ("
+                               "    title_ppn TEXT PRIMARY KEY,"
+                               "    local_fields BLOB NOT NULL"
+                               ") WITHOUT ROWID");
+
+    db_connection_->queryOrDie("CREATE TABLE IF NOT EXISTS local_ppns_to_title_ppns_map ("
+                               "    local_ppn TEXT PRIMARY KEY,"
+                               "    title_ppn TEXT NOT NULL"
```
mtrojan-ub, line 40: Should this be a foreign key to local_data?
Should anything be unique?

cpp/lib/src/LocalDataDB.cc
```diff
@@ -0,0 +1,215 @@
...
+    db_connection_->queryOrDie("SELECT local_fields FROM local_data WHERE title_ppn = "
+                               + db_connection_->escapeAndQuoteString(title_ppn));
+    auto result_set(db_connection_->getLastResultSet());
+    if (result_set.empty())
+        return false;
+
+    const auto row(result_set.getNextRow());
+    const auto previous_local_fields(BlobToLocalFieldsVector(row["local_fields"], title_ppn));
+    const auto local_ppns(ExtractLocalPPNsFromLocalFieldsVector(previous_local_fields));
+    for (const auto &local_ppn : local_ppns)
```
mtrojan-ub, line 156: Couldn't this be solved via a foreign key with cascade delete?

cpp/lib/src/LocalDataDB.cc
```diff
@@ -0,0 +1,215 @@
...
+            filtered_local_fields.emplace_back(local_field);
+    }
+
+    return filtered_local_fields;
+}
+
+
+bool LocalDataDB::removeLocalDataSet(const std::string &local_ppn) {
+    // 1. Determine the title PPN associated w/ the local PPN:
+    db_connection_->queryOrDie("SELECT title_ppn FROM local_ppns_to_title_ppns_map WHERE local_ppn="
```
mtrojan-ub, line 189: Couldn't this be solved via a foreign key with cascade delete?

cpp/lib/src/DbConnection.cc
ruschein: I don't know if this is generic enough.  I can't think of a typical application.

cpp/lib/src/LocalDataDB.cc
ruschein: Why?

cpp/lib/src/DbConnection.cc
```diff
@@ -463,17 +461,17 @@ void DbConnection::init(const std::string &database_name, const std::string &use
...
+        LOG_ERROR("mysql_init() failed! (1)");
 
     if (::mysql_real_connect(&mysql_, host.c_str(), user.c_str(), passwd.c_str(), database_name.c_str(), port,
                              /* unix_socket = */nullptr, /* client_flag = */CLIENT_MULTI_STATEMENTS) == nullptr)
 
         // password is intentionally omitted here!
-        throw std::runtime_error("in DbConnection::init: mysql_real_connect() failed! (" + getLastErrorMessage()
-                                 + ", host=\"" + host + "\", user=\"" + user + "\", passwd=\"********\", database_name=\""
-                                 + database_name + "\", port=" + std::to_string(port) + ")");
+        LOG_ERROR("mysql_real_connect() failed! (" + getLastErrorMessage()
```
mtrojan-ub, line 470: The password has been intentionally omitted from log file due to security reasons, see comment on line 469

cpp/marc_tools/delete_ids.cc
```diff
@@ -110,11 +129,13 @@ int Main(int argc, char *argv[]) {
     std::unordered_set<std::string> title_deletion_ids, local_deletion_ids;
     BSZUtil::ExtractDeletionIds(deletion_list.get(), &title_deletion_ids, &local_deletion_ids);
 
+    DbConnection db_connection(UBTools::GetTuelibPath() + "local_data.sq3", DbConnection::READONLY);
```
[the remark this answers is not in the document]
ruschein: Obsolete.

cpp/lib/src/LocalDataDB.cc
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    return local_fields_blob;
+}
+
+
+void LocalDataDB::insertOrReplace(const std::string &title_ppn, const std::vector<std::string> &local_fields) {
+    // 1. Clear out any local PPNs associated with "title_ppn":
+    db_connection_->queryOrDie("SELECT local_fields FROM local_data WHERE title_ppn = "
+                               + db_connection_->escapeAndQuoteString(title_ppn));
```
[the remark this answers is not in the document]
ruschein: Unfortunately not because it is based on escapeString() and that uses mysql_real_escape_string which requires	mysql_.

cpp/lib/src/LocalDataDB.cc
ruschein: Fixed in the commit from the ruschein2 branch.

cpp/lib/src/LocalDataDB.cc
ruschein: I don't think so because the local PPN's are not column values in the other table.

cpp/apply_differential_update.cc
```diff
@@ -44,6 +44,7 @@ namespace {
     ::Usage("[--min-log-level=log_level] [--keep-intermediate-files] input_directory "
             "difference_archive output_directory\n"
             "       Log levels are DEBUG, INFO, WARNING and ERROR with INFO being the default.\n");
+    std::exit(EXIT_FAILURE);
```
mtrojan-ub, line 47: Why? ::Usage already calls std::exit(EXIT_FAILURE)

cpp/apply_differential_update.cc
```diff
@@ -56,8 +57,11 @@ void CopyAndCollectPPNs(LocalDataDB * const local_data_db, MARC::Reader * const
             const auto first_local_field(record.findTag("LOK"));
             if (unlikely(first_local_field != record.end())) {
                 std::vector<std::string> local_fields;
-                for (auto local_field(first_local_field); local_field != record.end(); ++local_field)
+                auto local_field(first_local_field);
```
mtrojan-ub, line 60: Why? Doesn't this have the same effect, but needs 2 more lines of code?

mtrojan-ub:
Careful, this PR will revoke changes from #3499

cpp/store_local_data.cc
```diff
@@ -0,0 +1,77 @@
+/** \brief Saves local MARC data in a database for later retrieval with the add_local_data tool.
```
mtrojan-ub, line 1: store_local_data should be added to cpp/.gitignore

cpp/store_local_data.cc
```diff
@@ -0,0 +1,77 @@
...
+        auto local_field(record.findTag("LOK"));
+        if (unlikely(local_field == record.end())) {
+            LOG_WARNING("records w/ PPN " + PPN + " has no local fields!");
+            continue;
+        }
+
+        std::vector<std::string> local_fields;
+        do
+            local_fields.emplace_back(local_field->getContents());
+        while (++local_field != record.end());
```
mtrojan-ub, line 54: There might be other Tags between LOK and record.end() inserted by us, depending on the phase of the pipeline (e.g. "ORI")

ruschein: True and I did consider that.  As we don't know which current and future phases need the local data I considered that we need to re-add the local data early in the pipeline and therefore it should not cause a problem.  I also added comments in the pipeline scripts.

cpp/apply_differential_update.cc
ruschein: Obsolete

cpp/lib/src/LocalDataDB.cc
ruschein: Please have a look at the newest version.

cpp/lib/src/LocalDataDB.cc
```diff
@@ -57,8 +65,10 @@ LocalDataDB::~LocalDataDB() {
 
 
 void LocalDataDB::clear() {
-    db_connection_->queryOrDie("DROP TABLE IF EXISTS local_data");
     db_connection_->queryOrDie("DROP TABLE IF EXISTS local_ppns_to_title_ppns_map");
+    db_connection_->queryOrDie("DROP TABLE IF EXISTS local_data");
```
mtrojan-ub, line 69: Is this really "clear()"? I would suggest that clear would use TRUNCATE TABLE, but if you drop+recreate all tables its more like RESET or sth.

ruschein: It is a clear() as it clears out the tables.  That I drop and re-create the tables is an implementation detail that the caller does neither know, nor care about.  See also https://stackoverflow.com/questions/4280041/truncate-a-sqlite-table-if-it-exists

PR #3493, opened 2020-06-24
Made ReplaceEntitiesUTF8 more flexible.

ruschein:


cpp/lib/src/HtmlUtil.cc
```diff
@@ -889,8 +889,13 @@ std::string &ReplaceEntitiesUTF8(std::string * const s) {
             }
 
             // Output the entity:
-            if (not DecodeEntityUTF8(entity, &result))
-                result += "�";
+            if (not DecodeEntityUTF8(entity, &result)) {
+                if (unknown_entity_mode == IGNORE_UNKNOWN_ENTITIES) {
```
mtrojan-ub, line 893: Not sure if this is good enough. Imagine a big text where only one & is not encoded. The while-loop would put all remaining characters from the beginning of the string into the same entity and never try to decode the rest of the text.

So it would be better to break the while loop (886) as soon as the first non-entity-character is encountered.

ruschein: Incorrect: Please notice that the loop collecting characters for what might be an entity ends when it sees another ampersand or semicolon.

mtrojan-ub: Thats true, but far from sufficient. 

Imagine your text is
`Method & theory in the study of religion : journal of the North American Association for the Studies`

Then your entity candidate will be right now
`& theory in the study of religion : journal of the North American Association for the Studies`

1) we should break on any non-allowed entity character. Allowed at this position is only letter, digit, hash sign or semicolon.
2) it will not be enough to only break the while loop, because then at least one character will be lost in this case, so special handling will be needed

mtrojan-ub: Other example:
Method & theory in the study of religion :&nbsp;journal of the North American Association for the Studies

PR #3494, opened 2020-06-25
If at first you don't succeed, try, try again!

ruschein:


cpp/lib/src/HtmlUtil.cc
```diff
@@ -888,22 +891,31 @@ std::string &ReplaceEntitiesUTF8(std::string * const s, const UnknownEntityMode
                 ++ch;
             }
 
+            // Invalid end of entity candidate?
+            if (ch == s->end() or *ch == '&') {
+                if (unknown_entity_mode == IGNORE_UNKNOWN_ENTITIES) {
+                    result += '&';
```
mtrojan-ub, line 897: This looks risky... might lead to `&am` becoming `&am&`

PR #3498, opened 2020-06-25
zts_harvester_delivery_pipeline.sh

mtrojan-ub:


cpp/zts_harvester_delivery_pipeline.sh
```diff
@@ -91,61 +96,70 @@ log="${logdir}/${log_filename%.*}.log"
 rm -f "${log}"
 
 # Cleanup files/folders from a previous run
-mkdir -p $harvester_output_directory
-rm -r -f -d $harvester_output_directory/ixtheo
-rm -r -f -d $harvester_output_directory/krimdok
-rm -r -f -d $harvester_output_directory/ubtuebingen
+mkdir -p $HARVESTER_OUTPUT_DIRECTORY
```
ruschein, line 99: Please prefer the long flags like "--parents" in scripts.  They are self-documenting.

PR #3510, opened 2020-07-01
Derive Download URL from embedded metadata in OJS links

jriedl:


cpp/lib/src/SmartDownloader.cc
```diff
@@ -323,6 +323,30 @@ bool LocGovSmartDownloader::downloadDocImpl(const std::string &url, const TimeLi
 }
 
 
+bool OJSSmartDownloader::downloadDocImpl(const std::string &url, const TimeLimit &time_limit,
+                                        std::string * const document, std::string * const http_header_charset,
+                                        std::string * const error_message) {
```
ruschein, line 328: The brace should be on the next line.

cpp/lib/src/SmartDownloader.cc
```diff
@@ -323,6 +323,30 @@ bool LocGovSmartDownloader::downloadDocImpl(const std::string &url, const TimeLi
...
+     if (not RegexMatcher::Matched("/ojs/", url))
+         return false;
+     // Try to extract the Pdf Url from embedded metadata
```
ruschein, line 331: PDF and URL are acronyms and are written with all caps in English.

PR #3515, opened 2020-07-03
Fix SQL error detection problems.

ruschein:


cpp/lib/src/DbConnection.cc
```diff
@@ -127,6 +127,37 @@ int DbConnection::getLastErrorCode() const {
 }
 
 
+static std::vector<std::string> SplitMySQLStatements(const std::string &query) {
+    std::vector<std::string> statements;
+
+    std::string statement;
```
mtrojan-ub, line 133: should be inside the for loop

cpp/lib/src/DbConnection.cc
```diff
@@ -127,6 +127,37 @@ int DbConnection::getLastErrorCode() const {
...
+        } else if (unlikely(ch == ';')) {
+            statements.emplace_back(statement);
+            statement.clear();
+        } else {
+            if (ch == '\'' or ch == '"')
+                current_quote = ch;
+            statement += ch;
+        }
+    }
+    if (not statement.empty())
```
mtrojan-ub, line 154: should be inside the for loop

cpp/lib/src/DbConnection.cc
```diff
@@ -127,6 +127,37 @@ int DbConnection::getLastErrorCode() const {
...
+                current_quote = '\0';
+            else if (ch == '\\')
+                escaped = true;
+        } else if (unlikely(ch == ';')) {
+            statements.emplace_back(statement);
+            statement.clear();
+        } else {
+            if (ch == '\'' or ch == '"')
+                current_quote = ch;
+            statement += ch;
```
mtrojan-ub, line 151: What about MySQL comments?
https://dev.mysql.com/doc/refman/8.0/en/comments.html

cpp/lib/src/DbConnection.cc
mtrojan-ub: hmm or maybe not

cpp/lib/src/DbConnection.cc
mtrojan-ub: hmm or maybe not

cpp/lib/src/DbConnection.cc
ruschein: Why?

cpp/lib/src/DbConnection.cc
ruschein: Nope.

cpp/lib/src/DbConnection.cc
```diff
@@ -127,6 +127,37 @@ int DbConnection::getLastErrorCode() const {
...
+static std::vector<std::string> SplitMySQLStatements(const std::string &query) {
+    std::vector<std::string> statements;
+
+    std::string statement;
+    char current_quote('\0'); // NUL means not currently in a string constant.
+    bool escaped(false); // backslash not yet seen
+    for (const char ch : query) {
+        if (current_quote != '\0') {
+            statement += ch;
+            if (escaped)
```
mtrojan-ub, line 139: I thought escaping is only possible when inside a string constant, not outside

mtrojan-ub: ah right.. sorry my brain hurts

cpp/lib/src/DbConnection.cc
```diff
@@ -127,6 +127,37 @@ int DbConnection::getLastErrorCode() const {
...
+                escaped = false;
+            else if (ch == current_quote)
+                current_quote = '\0';
+            else if (ch == '\\')
+                escaped = true;
+        } else if (unlikely(ch == ';')) {
+            statements.emplace_back(statement);
```
mtrojan-ub, line 146: Trim before emplacing?

cpp/lib/src/DbConnection.cc
```diff
@@ -271,11 +305,9 @@ bool DbConnection::queryFile(const std::string &filename) {
         return false;
 
     if (type_ == T_MYSQL) {
-        ::mysql_set_server_option(&mysql_, MYSQL_OPTION_MULTI_STATEMENTS_ON);
         const bool query_result(query(StringUtil::TrimWhite(&statements)));
         if (query_result)
             mySQLSyncMultipleResults();
```
mtrojan-ub, line 310: If query handles everything we do not need to sync all results anymore here

cpp/lib/src/DbConnection.cc
```diff
@@ -127,6 +127,58 @@ int DbConnection::getLastErrorCode() const {
 }
 
 
+enum CommentFlavour { NO_COMMENT, C_STYLE_COMMENT, END_OF_LINE_COMMENT };
```
mtrojan-ub, line 130: C_STYLE_COMMENT is never used

cpp/lib/src/DbConnection.cc
```diff
@@ -127,6 +127,58 @@ int DbConnection::getLastErrorCode() const {
...
+                escaped = true;
+        } else if (unlikely(*ch == ';')) {
+            statements.emplace_back(statement);
+            statement.clear();
+        } else if (unlikely(*ch == '#'))
+            comment_flavour = END_OF_LINE_COMMENT;
+        else {
+            // Check for comments starting with "/*" or "-- ":
+            if ((*ch == '/' and ch + 1 < query.cend() and *(ch + 1) == '*')
+                or (*ch == '-' and ch + 2 < query.cend() and *(ch + 1) == '-' and *(ch + 2) == ' '))
```
mtrojan-ub, line 161: END_OF_LINE_COMMENT should be set as well if "-- "

cpp/lib/src/DbConnection.cc
```diff
@@ -127,6 +127,61 @@ int DbConnection::getLastErrorCode() const {
...
```
mtrojan-ub, line 161: move the second line into an elseif and you dont need the other if...else branches

cpp/lib/src/DbConnection.cc
```diff
@@ -127,6 +127,61 @@ int DbConnection::getLastErrorCode() const {
...
+                    ++ch;
+                    comment_flavour = C_STYLE_COMMENT;
+                } else {
+                    ch += 2;
+                    comment_flavour = END_OF_LINE_COMMENT;
+                }
+                continue;
+            }
+
+            if (*ch == '\'' or *ch == '"')
```
mtrojan-ub, line 173: ... and this into else

PR #3522, opened 2020-07-08
Death to trailing period for keywords!

ruschein:
Fix for tuefind issue 576.


cpp/pipelines/ixtheo_marc_pipeline_fifo.sh
```diff
@@ -66,6 +66,7 @@ StartPhase "Filter out Self-referential 856 Fields" \
     --remove-fields 'LOK:086630(.*)\x{1F}x' `# Remove internal bibliographic comments` \
     --filter-chars 130a:240a:245a '@' \
     --remove-subfields '6002:blmsh' '6102:blmsh' '6302:blmsh' '6892:blmsh' '6502:blmsh' '6512:blmsh' '6552:blmsh' \
+    --replace 600a:610a:630a:648a:650a:650x:651a:655a "(.*)\\.$" "\\1" `# Remove trailing periods for the following keyword normalisation.` \
```
mtrojan-ub, line 69: Why not "\\.$" ""? (Better performance)

ruschein: The behavious of marc_filter would be to replace the entire subfield with an empty string if a subfield ends with a period.  I just tested it.

PR #3526, opened 2020-07-16
Likely fix of pipeline problem.

ruschein:


cpp/generate_keyword_normalizations.sh
```diff
@@ -0,0 +1,37 @@
...
+
+
+# Argument processing
+if [[ $# != 1 ]]; then
+    Usage
+fi
+email_address=$1
+
+
+/usr/local/bin/generate_keyword_normalizations ptah.ub.uni-tuebingen.de:8080 /tmp/new_keyword_normalization.map
```
mtrojan-ub, line 35: server shouldn't be hardcoded

PR #3535, opened 2020-07-20
Riedl

jriedl:


cronjobs/github_api_util.py
```diff
@@ -0,0 +1,44 @@
...
+    github_oauth_token = os.environ.get('GITHUB_OAUTH_TOKEN')
+    headers = { 'Authorization' :  'token ' + github_oauth_token }
+    url = 'https://api.github.com/repos/ubtue/' + repository + '/issues/' + issue_number
+    req = requests.post(url, json=data, headers=headers)
+    req.raise_for_status()
+    print(req.text)
+
+
+def GetISSNMatcher():
+    return re.compile('([0-9]{4}-[0-9]{3}[0-9X])')
```
ruschein, line 37: Hyphens are optional for ISSN's. From the Zotero forum "Still, though, the hyphens in ISSNs and ISBNs should never be treated as anything but presentational polish."

jriedl: In general yes. However, we evaluate ZEDER data or user generated issue subjects the are geared to ZEDER, where the hyphen is always present. Thus, in this context, omitting '?' in the regex is a safeguard against erroneously extracting PPNs or other arbitrary numbers.

ruschein: In that case, why not duplicate the functionality that we have in our C++ function MiscUtil::IsPossibleISSN?

cronjobs/github_api_util.py
```diff
@@ -0,0 +1,44 @@
+# Python 3 module
+# -*- coding: utf-8 -*-
+import json
+import os
+import re
+import requests
+import util
+
```
ruschein, line 8: Missing black line.

cronjobs/github_api_util.py
```diff
@@ -0,0 +1,44 @@
...
+def GetAllIssuesForUBTueRepository(repository):
```
ruschein, line 9: Please use type annotations in all of our new Python code.

cronjobs/zjs_create_issues.py
```diff
@@ -0,0 +1,78 @@
...
+def CreateNewZoteroJournalStatusIssues():
+    journal_types = [ "IxTheo", "KrimDok" ]
+    zeder_titles = {}
+    zotero_group = "zotero_group"
+    config = GetZoteroConfiguration()
+    github_existing_issues = github_api_util.GetAllIssuesForUBTueRepository(ZOTERO_JOURNAL_STATUS_REPO)
+    issn_matcher = github_api_util.GetISSNMatcher()
+    for section in config.sections():
+        if config.has_option(section, zotero_group) and config.get(section, zotero_group) in journal_types:
+            issn = config.get(section, "online_issn") if config.has_option(section, "online_issn") \
```
ruschein, line 53: I find this very hard to parse.  Please only use at most one trailing if and then only for short statements as a poor man's ternary operator.

cronjobs/zjs_create_issues.py
```diff
@@ -0,0 +1,78 @@
...
+                    else config.get(section, "print_issn") if config.has_option(section, "print_issn") \
+                    else NO_KNOWN_ISSN
+            #Filter garbage
+            issn = issn if issn_matcher.match(issn) else NO_KNOWN_ISSN
+            try:
+                    for issue in github_existing_issues:
```
ruschein, line 59: We only indent 4 spaces per nesting level.

cronjobs/zjs_create_issues.py
```diff
@@ -0,0 +1,78 @@
...
+                        if issn != NO_KNOWN_ISSN:
+                            if issue['title'].find(issn) != -1:
+                                raise Entry_Present()
```
ruschein, line 62: Why not just use "continue" here?

cronjobs/zjs_create_issues.py
```diff
@@ -0,0 +1,78 @@
...
+                        else:
+                            if issue['title'].find(section) != -1:
+                                raise Entry_Present()
```
ruschein, line 65: Ditto.

cronjobs/zjs_tag_from_datenprobleme.py
```diff
@@ -0,0 +1,92 @@
+#!/bin/python3
+# -*- coding: utf-8 -*-
+import json
+import re
+import urllib.parse
+import github_api_util
+
```
ruschein, line 7: Missing blank line.

cronjobs/zjs_tag_from_datenprobleme.py
```diff
@@ -0,0 +1,92 @@
...
+
+def ExtractOpenAndClosedWithoutOpenISSNs(issues):
+    closed_issues_issns = set()
+    open_issues_issns = set()
+    issn_matcher = github_api_util.GetISSNMatcher()
+    for issue in issues:
+        title = issue['title']
+        issns = issn_matcher.findall(title)
+        if len(issns):
```
ruschein, line 20: Why is this test needed?

cronjobs/zjs_tag_from_datenprobleme.py
```diff
@@ -0,0 +1,92 @@
...
+            for issn in issns:
+                 if issue['state'] == "open":
```
ruschein, line 22: Indentation looks off.

cronjobs/zjs_tag_from_datenprobleme.py
```diff
@@ -0,0 +1,92 @@
...
+        new_labels.append(label_to_add)
+    return new_labels
+
+
+def LabelsAreIdentical(issue, new_labels):
+    current_labels = issue['labels']
+    old_labels = []
+    for label in current_labels:
+        old_labels.append(label['name'])
+    old_labels_set = set(old_labels)
```
ruschein, line 46: Why do you need two sets of variables here?  Just use the sets only.

cronjobs/zjs_create_issues.py
jriedl: This is is continue in the outer for loop

cronjobs/github_api_util.py
jriedl: I cannot find annotations anywhere in our code base.

ruschein: That's because we were on Python 2 not that long ago and that did not support type annotations.  In general we want all the support in finding mistakes that our compilers or interpreters provide.

cronjobs/zjs_create_issues.py
```diff
@@ -0,0 +1,82 @@
...
+# -*- coding: utf-8 -*-
+import configparser
+import json
+import os
+import stdnum.issn as issn_checker
+import urllib.request
+import github_api_util
+
+
+ZOTERO_JOURNAL_STATUS_REPO='zotero-journal-status'
```
ruschein, line 11: Missing spaces

PR #3536, opened 2020-07-21
Fixed inconsistent spelling.

ruschein:


cpp/pipelines/ixtheo_marc_pipeline_fifo.sh
```diff
@@ -66,7 +66,7 @@ StartPhase "Filter out Self-referential 856 Fields" \
     --filter-chars 130a:240a:245a '@' \
     --remove-subfields '6002:blmsh' '6102:blmsh' '6302:blmsh' '6892:blmsh' '6502:blmsh' '6512:blmsh' '6552:blmsh' \
     --replace 600a:610a:630a:648a:650a:650x:651a:655a "(.*)\\.$" "\\1" `# Remove trailing periods for the following keyword normalisation.` \
-    --replace-strings 600a:610a:630a:648a:650a:650x:651a:655a /usr/local/var/lib/tuelib/keyword_normalisation.map \
+    --replace-strings 600a:610a:630a:648a:650a:650x:651a:655a /usr/local/var/lib/tuelib/keyword_normalization.map \
```
jriedl, line 69: There only seems to be a file `/usr/local/ub_tools/cpp/data/installer/keyword_normalization.map`, i.e. probably there are no regular updates to the file in `/usr/local/var/lib/tuelib` (which is still called `keyword_normalisation.map`)

PR #3537, opened 2020-07-22
RegexMatcher

mtrojan-ub:


cpp/lib/include/RegexMatcher.h
```diff
@@ -186,6 +186,8 @@ class RegexMatcher {
      */
     static bool Matched(const std::string &regex, const std::string &subject, const unsigned options = 0,
                         std::string * const err_msg = nullptr, size_t * const start_pos = nullptr, size_t * const end_pos = nullptr);
+
+    static std::string Escape(const std::string &subpattern);
```
ruschein, line 190: Please add a comment describing what this does.

cpp/lib/src/RegexMatcher.cc
```diff
@@ -385,6 +385,15 @@ bool RegexMatcher::Matched(const std::string &regex, const std::string &subject,
 }
 
 
+std::string RegexMatcher::Escape(const std::string &subpattern) {
+    // see `man pcrepattern`
+    // escape backslashes first to keep from overwriting other escape sequences
+    std::string escaped_subpattern(subpattern);
+    StringUtil::BackslashEscape("\\^$.[]|()?*+{}", &escaped_subpattern);
```
ruschein, line 392: You can return the result of this function call directly.

PR #3540, opened 2020-07-28
Generate CSV as per Martin's request.

ruschein:


cpp/lib/src/TextUtil.cc
```diff
@@ -733,9 +733,13 @@ std::string CSVEscape(const std::string &value, const bool add_quotes) {
         escaped_value += '"';
 
     for (const char ch : value) {
-        if (unlikely(ch == '"'))
-            escaped_value += '"';
-        escaped_value += ch;
+        if (not add_quotes and unlikely(ch == ','))
+            escaped_value += "\",\"";
+        else {
+            if (unlikely(ch == '"'))
```
jriedl, line 739: If seems to have no effect.

ruschein: Strange because my new output is different and has escaped commas!

PR #3542, opened 2020-07-31
Ruschein

ruschein:


cpp/bulk_mailer.cc
```diff
@@ -0,0 +1,107 @@
+/** \file    bulk_mailer.cc
+ *  \brief   Sends emails to a list of users.
+ *  \author  Dr. Johannes Ruscheinski
+ */
+
+/*
+    Copyright (C) 2020 Library of the University of Tübingen
```
mtrojan-ub, line 7: Needs to be `Tübingen University Library`, see mail (29.04.2020)

cpp/bulk_mailer.cc
```diff
@@ -0,0 +1,107 @@
...
+
+
+namespace {
+
+
+[[noreturn]] void Usage() {
+    ::Usage("mail_contents sql_statement auxillary_email_address_list reply_to_email_address\n"
+            "\tmail_contents: the file containing the email message\n"
+            "\t\t(the first line will be used as the mail's subject)\n"
+            "\tsql_statement: a query to select the relevant email addresses from the vufind user table\n"
```
mtrojan-ub, line 40: Since CollectRecipientsFromSqlTable returns row[0], it might be a good idea to explicitly say that the given SQL statement may only select 1 column

cpp/bulk_mailer.cc
```diff
@@ -0,0 +1,107 @@
...
+    const auto old_size(recipients->size());
+
+    const auto db_connection(VuFind::GetDbConnection());
+    db_connection->queryOrDie(select_statment);
+    DbResultSet result_set(db_connection->getLastResultSet());
+    DbRow row;
+    while (row = result_set.getNextRow())
+        recipients->emplace_back(row[0]);
+
+    LOG_INFO("Added " + std::to_string(recipients->size() - old_size) + " recipients due to the database query.");
```
mtrojan-ub, line 56: "Added recipients" could be mistaken for some logic using an INSERT statement. If the function is named "Collect...", why not use this term for logging as well?

cpp/bulk_mailer.cc
```diff
@@ -0,0 +1,107 @@
...
+}
+
+
+void CollectRecipientsFromFile(const std::string &filename, std::vector<std::string> * const recipients) {
+    const auto old_size(recipients->size());
+
+    for (auto line : FileUtil::ReadLines(filename))
+        recipients->emplace_back(line);
+
+    LOG_INFO("Added " + std::to_string(recipients->size() - old_size) + " from " + filename + ".");
```
mtrojan-ub, line 66: Log message could also be changed here, see previous comment

cpp/data/installer/tuefind.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE vufind.user ADD ub_subscribed_to_newsletter BOOL NOT NULL DEFAULT FALSE;
```
mtrojan-ub, line 1: Contents of this file should be moved to ubtue/tuefind repository (module/TueFind/sql/mysql.sql
(this file is already used by the installer)

mtrojan-ub: Also, prefix should not be "ub_" but "tuefind_".

cpp/data/sql_updates/vufind.1
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE vufind.user ADD ub_subscribed_to_newsletter BOOL NOT NULL DEFAULT FALSE;
```
mtrojan-ub, line 1: change prefix here as well

cpp/installer.cc
```diff
@@ -313,6 +313,7 @@ void CreateUbToolsDatabase(const OSSystemType os_system_type) {
         DbConnection::MySQLGrantAllPrivileges(sql_database, sql_username, root_username, root_password);
         DbConnection::MySQLGrantAllPrivileges(sql_database + "_tmp", sql_username, root_username, root_password);
         DbConnection::MySQLImportFile(INSTALLER_DATA_DIRECTORY + "/ub_tools.sql", sql_database, root_username, root_password);
+        DbConnection::MySQLImportFile(INSTALLER_DATA_DIRECTORY + "/tuefind.sql", sql_database, root_username, root_password);
```
mtrojan-ub, line 316: File in TueFind module is already covert by installer

cpp/mail_krimdok_newsletter.sh
```diff
@@ -0,0 +1,24 @@
...
+    exit 0
+fi
+
+time_since_last_modification=$(($(date +%s) - $(stat --format=%Y -- "$most_recent_email_contents")))
+if [ $time_since_last_modification -gt 86400 ]; then
+    echo "No recent newsletters found"
+    exit 0
+fi
+
+bulk_mailer "$most_recent_email_contents" \
```
mtrojan-ub, line 20: credentials need to be changed before merge

cpp/mail_krimdok_newsletter.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -o errexit -o nounset
+
+
+cd .
```
mtrojan-ub, line 5: strange looking newlines & unnecessary looking `cd .`

cpp/mail_krimdok_newsletter.sh
```diff
@@ -0,0 +1,24 @@
...
+
+
+most_recent_email_contents=$(ls -Art | tail -n 1)
+if [ -z "$most_recent_email_contents" ]; then
+    echo "No newsletters found"
+    exit 0
+fi
+
+time_since_last_modification=$(($(date +%s) - $(stat --format=%Y -- "$most_recent_email_contents")))
+if [ $time_since_last_modification -gt 86400 ]; then
```
mtrojan-ub, line 15: This does not really look fail-safe, it might be better to rename the file to a ".sent" extension which is skipped when looking for new newsletters, or something like that.

cpp/data/NewsletterUploadForm.html
```diff
@@ -0,0 +1,11 @@
+<html>
+  <head>
+    <title>Newsletter Upload Form</title>
+  </head>
+  <body>
+    <form enctype="multipart/form-data" action="/var/www/cgi-bin/newsletter_uploader" method="post">
```
mtrojan-ub, line 6: - newsletter_uploader CGI program missing
- /var/www/cgi-bin will not work when called via browser

cpp/data/NewsletterUploadForm.html
```diff
@@ -0,0 +1,11 @@
...
+      <p>File: <input type="file" name="file"></p>
```
mtrojan-ub, line 7: What exactly is the structure of the file? Please provide documentation.

cpp/bulk_mailer.cc
```diff
@@ -0,0 +1,107 @@
...
+#include "VuFind.h"
+#include "util.h"
+
+
+namespace {
+
+
+[[noreturn]] void Usage() {
+    ::Usage("mail_contents sql_statement auxillary_email_address_list reply_to_email_address\n"
+            "\tmail_contents: the file containing the email message\n"
```
mtrojan-ub, line 38: Please provide documentation about file structure here (e.g. how to store mail subject, body, maybe attachments?)

cpp/data/NewsletterUploadForm.html
ruschein: https://www.oreilly.com/library/view/cgi-programming-on/9781565921689/07_chapter-04.html

cpp/data/NewsletterUploadForm.html
ruschein: Documented in the usage message of the newsletter_uploader program and in NewsletterUploadForm.html.

cpp/bulk_mailer.cc
ruschein: There is no support for attachments planned and the file structure has been explained in the HTML form.

cpp/Makefile
```diff
@@ -66,6 +66,9 @@ data_install: Makefile data/*.conf data/*.map data/*.ttf data/IxTheo_Notation.cs
 	if [ -d $(INCROND_DIR) ]; then \
 		cp data/pipeline_script_triggers $(INCROND_DIR); \
 	fi
+	@echo "Installing newsletter HTML form"
+	cp data/NewsletterUploadForm.html /usr/local/vufind/public/
+	chcon unconfined_u:object_r:usr_t:s0 /usr/local/vufind/public/NewsletterUploadForm.html
```
mtrojan-ub, line 71: This should be handled by installer.cc by adding a proper SELinux File Context like in all other cases.

cpp/cgi_progs/newsletter_uploader.py
```diff
@@ -0,0 +1,30 @@
...
+
+# A nested FieldStorage instance holds the file
+file_field = form['file']
+
+
+# Test if the file was uploaded
+if file_field:
+    # Strip leading path from file name to avoid directory traversal attacks
+    filename = os.path.basename(file_field.filename)
+    open("/usr/local/var/lib/tuelib/newsletters" + filename, 'w').write(file_field.file.read())
```
mtrojan-ub, line 19: What if newsletters subdir doesn't exist yet? (create dynamically, or let installer create it?)

cpp/mail_krimdok_newsletter.sh
```diff
@@ -0,0 +1,18 @@
...
+
+
+BACKUP_DIR=/usr/local/var/lib/tuelib/sent_newsletters
+mkdir --parents "$BACKUP_DIR"
+REPLY_TO=johannes.ruscheinski@uni-tuebingen.de
+SQL_QUERY="SELECT 'johannes.ruscheinski@uni-tuebingen.de'"
+declare -i count=0
+for regular_file in $(ls -p | grep -v /); do
+    bulk_mailer "$regular_file" \
+                "$SQL_QUERY" \
```
mtrojan-ub, line 12: SQL query should be escaped before using it as arg (might contain double quotes, backslashes and so on)

cpp/mail_krimdok_newsletter.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -o errexit -o nounset
+
+
+BACKUP_DIR=/usr/local/var/lib/tuelib/sent_newsletters
```
mtrojan-ub, line 5: a subdir inside the newsletters dir might be better

cpp/mail_krimdok_newsletter.sh
```diff
@@ -0,0 +1,18 @@
...
+mkdir --parents "$BACKUP_DIR"
+REPLY_TO=johannes.ruscheinski@uni-tuebingen.de
```
mtrojan-ub, line 7: Don't forget to change the hardcoded reply later

cpp/mail_krimdok_newsletter.sh
```diff
@@ -0,0 +1,18 @@
...
+SQL_QUERY="SELECT 'johannes.ruscheinski@uni-tuebingen.de'"
+declare -i count=0
+for regular_file in $(ls -p | grep -v /); do
+    bulk_mailer "$regular_file" \
+                "$SQL_QUERY" \
+                /usr/local/var/lib/tuelib/auxillary_newsletter_email_addresses \
```
mtrojan-ub, line 13: Where does this file come from? And wouldn't it be better to prefix it with krimdok_ or relbib_ etc.

cpp/data/NewsletterUploadForm.html
```diff
@@ -0,0 +1,14 @@
+<html>
+  <head>
+    <title>Newsletter Upload Form</title>
+  </head>
+  <body>
+    The uploaded file must be a plin-text file.  The first line will be used as the subject of the
```
mtrojan-ub, line 6: typo: plin-text => plain-text

cpp/data/NewsletterUploadForm.html
```diff
@@ -0,0 +1,14 @@
...
+    email message and the remainder as the message body.  The encoding of the file must be UTF-8 with
+    no leading BOM.
+    <form enctype="multipart/form-data" action="/var/www/cgi-bin/newsletter_uploader" method="post">
```
mtrojan-ub, line 9: I still doubt this works, but still => shouldn't the .py extension be added to the script name?

cpp/cgi_progs/newsletter_uploader.py
mtrojan-ub, line 19: Shouldn't there be a / between `/newsletters `and `filename`?

cpp/installer.cc
```diff
@@ -75,7 +76,7 @@
 
 
 [[noreturn]] void Usage() {
-    ::Usage(std::string("--ub-tools-only|--fulltext-backend|(vufind_system_type [--omit-cronjobs] [--omit-systemctl])\n") +
+    ::Usage(std::string("(--production|--test) --ub-tools-only|--fulltext-backend|(vufind_system_type [--omit-cronjobs] [--omit-systemctl])\n") +
```
mtrojan-ub, line 79: Please also change all affected Dockerfiles that call install_ubtools.sh (which passes the parameters on to installer)

cpp/installer.cc
```diff
@@ -567,10 +572,8 @@ void InstallVuFindCronjobs(const VuFindSystemType vufind_system_type) {
         names_to_values_map.insertScalar("relbib_host", GetStringFromTerminal("RelBib Hostname"));
     }
 
-    if (vufind_system_type == KRIMDOK)
-        InstallCronjobs("krimdok.cronjobs", start_vufind_autogenerated, end_vufind_autogenerated);
-    else
-        InstallCronjobs("ixtheo_cronjobs", start_vufind_autogenerated, end_vufind_autogenerated, names_to_values_map);
+    InstallCronjobs(production, (vufind_system_type == KRIMDOK ? "krimdok.cronjobs" : "ixtheo_cronjobs"),
```
mtrojan-ub, line 575: `ixtheo_cronjobs` => should this be `ixtheo.cronjobs`?

cpp/data/sql_updates/vufind.1
mtrojan-ub: Why is this conversation resolved? Nothing has changed.
Conversations should normally only be resolved by the person who begins them.

cpp/mail_krimdok_newsletter.sh
ruschein: Obsolete

cpp/mail_krimdok_newsletter.sh
ruschein: Yup!

cpp/data/NewsletterUploadForm.html
ruschein: I found several examples that support that a bare path works.  I have added the missing extension.

cpp/cgi_progs/newsletter_uploader.py
ruschein: Absoluetely!

cpp/installer.cc
ruschein: Yes, indeed!

cpp/mail_krimdok_newsletter.sh
ruschein: Agreed.

cpp/mail_krimdok_newsletter.sh
ruschein: It comes from the FID directory on the network drive and there can nn conflict right now also because KrimDok systems have no overlap w/ anything else.

cpp/installer.cc
mtrojan-ub, line 79: Does the usage still make sense like this?
- Lacking documentation => what does production / test acually mean in which system context?
- Is it still necessary to have --omit-cronjobs and --omit-systemctl flags or is this indirectly defined via the production flag?

docker/fulltext-centos/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN dnf config-manager --add-repo https://download.opensuse.org/repositories/hom
 RUN rpm --import https://build.opensuse.org/projects/home:Alexander_Pozdnyakov/public_key
 RUN dnf --assumeyes update
 RUN dnf --assumeyes install tesseract-devel net-tools vim
-RUN /tmp/install_ubtools.sh --fulltext-backend
+RUN /tmp/install_ubtools.sh --production --fulltext-backend
```
mtrojan-ub, line 20: Really? docker containers are always only for testing as far as i know

cpp/mail_krimdok_newsletter.sh
```diff
@@ -0,0 +1,17 @@
...
+
+
+BACKUP_DIR=/usr/local/var/lib/tuelib/newsletters/sent
+mkdir --parents "$BACKUP_DIR"
+REPLY_TO=johannes.ruscheinski@uni-tuebingen.de
+declare -i count=0
+for regular_file in $(ls -p | grep -v /); do
+    bulk_mailer "$regular_file" \
+                "SELECT 'johannes.ruscheinski@uni-tuebingen.de'" \
+                /usr/local/var/lib/tuelib/krimdok_auxillary_newsletter_email_addresses \
```
mtrojan-ub, line 12: Where does this file come from?

cpp/Makefile
ruschein: Done.

cpp/cgi_progs/newsletter_uploader.py
ruschein: Installer does it now.

cpp/mail_krimdok_newsletter.sh
ruschein: Inlined the SQL statement.

cpp/installer.cc
ruschein: Well, production means production.  That is, if there is a difference between a test environment and a production environment this would select for that.  There are, possibly several, installation types where this has no effect so far as they have no differences or concepts of what a production or test environment are.  An example is probably if you choose --ub-tools-only.  Then either flag is fine.

ruschein: On second thought we definitely need those as they are needed for the difference between sobek and ptah on one side and the Docker containers on the other side.

PR #3544, opened 2020-07-31
export_vufind_redirects

mtrojan-ub:


cpp/export_vufind_redirects.cc
```diff
@@ -0,0 +1,49 @@
...
+#include <cstdlib>
+#include "FileUtil.h"
+#include "TextUtil.h"
+#include "VuFind.h"
+#include "util.h"
+
+
+int Main(int argc, char **argv) {
+    if (argc != 2) {
+        std::cerr << "Usage: " << ::progname << " export_file\n";
```
ruschein, line 29: Please use the Usage function from util.h.

cpp/export_vufind_redirects.cc
```diff
@@ -0,0 +1,49 @@
...
+        std::exit(EXIT_FAILURE);
+    }
+    
+    const std::string export_file(argv[1]);
+    
+    auto db_connection(VuFind::GetDbConnection());
```
ruschein, line 35: Can be const, I assume.

cpp/export_vufind_redirects.cc
```diff
@@ -0,0 +1,49 @@
...
+    db_connection->queryOrDie("SELECT * FROM tuefind_redirect");
+    auto result_set(db_connection->getLastResultSet());
+    
+    std::string csv;
```
ruschein, line 39: If this can get large it would be much better to write this to the output file continuously instead of accumulating the output in a string first.

cpp/lib/include/VuFind.h
```diff
@@ -22,10 +22,7 @@
 
 #include <memory>
 #include <string>
-
-
-// Forward declaration:
-class DbConnection;
+#include "DbConnection.h"
```
ruschein, line 25: Why?

mtrojan-ub: Because if you have a script which includes only VuFind.h and tries to call GetDbConnection, this will fail if you don't include DbConnection.h in VuFind.h

cpp/export_vufind_redirects.cc
```diff
@@ -0,0 +1,48 @@
...
+    File csv_file(argv[1], "w", File::ThrowOnOpenBehaviour::THROW_ON_ERROR);
+    csv_file.write("url;group;timestamp\n");
+    while (const auto &db_row = result_set.getNextRow()) {
+        std::string csv_row;
+        csv_row += TextUtil::CSVEscape(db_row["url"]) + ";";
+        csv_row += TextUtil::CSVEscape(db_row["group_name"]) + ";";
+        csv_row += TextUtil::CSVEscape(db_row["timestamp"]) + "\n";
+        csv_file.write(csv_row);
+    }
+    csv_file.close();
```
ruschein, line 45: You don't need this line.  The destructor of the File class already takes care of it.

cpp/upload_to_bsz_ftp_server.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 # Script to upload files safely to the BSZ FTP server
+#
+# Note that the used ftp command will not terminate if an error occurs.
+# This script will be replaced by a python script soon.
```
ruschein, line 6: We tend not to have a blank line above the "set".

cronjobs/bsz_util.py
```diff
@@ -82,18 +94,18 @@ def GetBackupDirectoryPath(config):
 
 # Check whether all the instances are needed
 def NeedsBothInstances(filename_regex):
-     without_localdata_pattern = "_o[)]?-[(]?.*[)]?"
-     without_localdata_regex = re.compile(without_localdata_pattern)
-     return without_localdata_regex.search(filename_regex.pattern) is not None
+    without_localdata_pattern = "_o[)]?-[(]?.*[)]?"
```
ruschein, line 97: General comment: It would be nice to use type hinting wherever possible, of course taking into account what the version of Python 3 on CentOS supports.

cronjobs/bsz_util.py
```diff
@@ -82,18 +94,18 @@ def GetBackupDirectoryPath(config):
...
 # For IxTheo our setup requires that we obtain both the files with and without local data because otherwise
 # we get data inconsistencies
 def AreBothInstancesPresent(filename_regex, remote_files):
-     if not remote_files:
+    if not remote_files:
         return True
-     matching_remote_files = list(filter(filename_regex.match, remote_files))
-     return True if len(matching_remote_files) % 2 == 0 else False
+    matching_remote_files = list(filter(filename_regex.match, remote_files))
+    return True if len(matching_remote_files) % 2 == 0 else False
```
ruschein, line 108: Arrgh!  Delete "True if" and "else False" from this line!

cronjobs/util.py
```diff
@@ -452,6 +452,37 @@ def FTPLogin(ftp_host, ftp_user, ftp_passwd):
     return ftp
 
 
+def FTPChangeDirectory(ftp, remote_dir_path):
```
ruschein, line 455: It might have been nice to implement our own FTP class instead of a series of functions...

mtrojan-ub: great idea, i'm on it

cpp/export_vufind_redirects.cc
```diff
@@ -0,0 +1,45 @@
...
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include <iostream>
```
ruschein, line 19: Do you actually use iostream or FileUtil?

cronjobs/ftp_connection.py
```diff
@@ -0,0 +1,88 @@
+# Python 3 module
+# -*- coding: utf-8 -*-
+
```
ruschein, line 3: I think we never have a blank line here.

cronjobs/ftp_connection.py
```diff
@@ -0,0 +1,88 @@
...
+from ftplib import FTP
+import os
+
+
+class FTPConnection:
+
```
ruschein, line 9: ditto

cronjobs/ftp_connection.py
```diff
@@ -0,0 +1,88 @@
...
+        error_message = "failed to change directory: " + remote_dir_path
+        try:
+            res = self._ftp.cwd(remote_dir_path)
+            if not res.startswith('250 '):
+                Error(error_message)
+        except Exception as e:
+            Error(error_message + " (" + str(e) + ")")
+
+    def listDirectory(self):
+        error_message = "failed to list directory"
```
ruschein, line 47: Why not just include this on line 51?

cronjobs/ftp_connection.py
```diff
@@ -0,0 +1,88 @@
...
+        try:
+            return self._ftp.nlst()
+        except Exception as e:
+            Error(error_message + " (" + str(e) + ")")
+
+    def downloadFile(self, remote_file_name, local_file_path):
```
ruschein, line 53: I suggest making the 2nd argument optional and preset it to None, then you can assign remote_file_name to local_file_path when the latter is None.

cronjobs/ftp_connection.py
```diff
@@ -0,0 +1,88 @@
...
+            remote_file_name = os.path.basename(local_file_path)
+        error_message = "failed to upload file " + local_file_path + " to " + remote_file_name
+        try:
+            with open(local_file_path, 'rb') as fp:
+                res = self._ftp.storbinary("STOR " + remote_file_name, fp)
+                if not res.startswith('226 '):
+                    Error(error_message)
+        except Exception as e:
+            Error(error_message + " (" + str(e) + ")")
+
```
ruschein, line 80: Missing blank line.

cronjobs/ftp_connection.py
```diff
@@ -0,0 +1,88 @@
...
+    def renameFile(self, remote_file_name_old, remote_file_name_new):
```
ruschein, line 81: Maybe call this renameRemoteFile?

cronjobs/upload_to_bsz_ftp_server.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# A tool to upload files to the BSZ FTP server.
+# In earlier times we had a bash script for this which was using the "ftp" command,
+# but due to problems with error handling granularity we switched to python.
+
```
ruschein, line 7: We tend not to have a blank line here in Python.

cpp/export_vufind_redirects.cc
```diff
@@ -0,0 +1,48 @@
...
+    auto result_set(db_connection->getLastResultSet());
+
+    File csv_file(argv[1], "w", File::ThrowOnOpenBehaviour::THROW_ON_ERROR);
+    csv_file.write("url;group;timestamp\n");
+    while (const auto &db_row = result_set.getNextRow()) {
+        std::string csv_row;
+        csv_row += TextUtil::CSVEscape(db_row["url"]) + ";";
+        csv_row += TextUtil::CSVEscape(db_row["group_name"]) + ";";
+        csv_row += TextUtil::CSVEscape(db_row["timestamp"]) + "\n";
+        csv_file.write(csv_row);
```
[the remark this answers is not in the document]
mtrojan-ub: done

cpp/export_vufind_redirects.cc
mtrojan-ub: removed

cronjobs/bsz_util.py
mtrojan-ub: done

cronjobs/util.py
mtrojan-ub: done

cronjobs/ftp_connection.py
mtrojan-ub: Why? The only file you can rename in FTP context is the remote file. Providing a method inside a FTP class to rename a local file doesn't make sense in my opinion.

cpp/export_vufind_redirects.cc
mtrojan-ub: not anymore => removed

cronjobs/ftp_connection.py
mtrojan-ub: done

cronjobs/ftp_connection.py
mtrojan-ub: done

cronjobs/ftp_connection.py
mtrojan-ub: copy/paste => changed

cronjobs/ftp_connection.py
mtrojan-ub: done

cronjobs/upload_to_bsz_ftp_server.py
mtrojan-ub: done

cronjobs/ftp_connection.py
mtrojan-ub: Somehow that sounds strange, i hope you meant the 3rd argument and not the 2nd (counting `self` as first)

cpp/export_vufind_redirects.cc
```diff
@@ -0,0 +1,43 @@
...
+
+int Main(int argc, char **argv) {
+    if (argc != 2)
+        ::Usage("export_file");
+
+    const auto db_connection(VuFind::GetDbConnection());
+    db_connection->queryOrDie("SELECT * FROM tuefind_redirect");
+    auto result_set(db_connection->getLastResultSet());
+
+    File csv_file(argv[1], "w", File::ThrowOnOpenBehaviour::THROW_ON_ERROR);
```
ruschein, line 34: Not important, but, you might use FileUtil::OpenOutputOrDie.

cpp/export_vufind_redirects.cc
```diff
@@ -0,0 +1,43 @@
...
+    csv_file.write("url;group;timestamp\n");
+    while (const auto &db_row = result_set.getNextRow()) {
```
ruschein, line 36: This should not be a reference because DbResultSet::getNextRow does not return references.  Unfortunately in C++ if you create a const reference, the compiler will generate a temporary object if needed and then you use a pointer indirection to reference it unnecessarily. :-(

cronjobs/ftp_connection.py
```diff
@@ -0,0 +1,88 @@
...
+        try:
+            return self._ftp.nlst()
+        except Exception as e:
+            Error("failed to list directory (" + str(e) + ")")
+
+    def downloadFile(self, remote_file_name, local_file_path=None):
+        if local_file_path is None:
+            local_file_path = remote_file_name
+        try:
+            output = open(local_file_path, "wb")
```
ruschein, line 54: FYI (no change necessary): On UNIX-like systems there is no difference between text and binary mode.  So, unless you want something to be portable to Windows the "b" is not necessary.  It's harmless and could be viewed as a hint to the reader that we intent to read or write binary files.

cronjobs/ftp_connection.py
ruschein: Does everybody know this?

mtrojan-ub: Yes 😆

cpp/export_vufind_redirects.cc
mtrojan-ub: changed

cpp/export_vufind_redirects.cc
mtrojan-ub: changed

cronjobs/ftp_connection.py
mtrojan-ub: Nice to know, ty!

cronjobs/ftp_connection.py
ruschein: I see...

PR #3562, opened 2020-08-24
GetAuthorGNDNumber

mtrojan-ub:


cpp/lib/include/BSZUtil.h
```diff
@@ -53,6 +53,11 @@ enum ArchiveType { TITLE_RECORDS, SUPERIOR_TITLES, AUTHORITY_RECORDS, LOCAL_RECO
 ArchiveType GetArchiveType(const std::string &member_name);
 
 
+// Try to resolve the GND number for an author using the BSZ (SWB / OGND) website.
+// The base URL can e.g. be taken from zotero_harvester.conf
+std::string GetAuthorGNDNumber(const std::string &author, const std::string &author_lookup_base_url);
```
ruschein, line 58: You need to document that this function is thread safe here.

ruschein (changes requested):

cpp/lib/include/BSZUtil.h
mtrojan-ub: done!

PR #3569, opened 2020-08-26
zeder_to_zotero_importer.cc: Allow multiple ISSN's in same field

mtrojan-ub:


cpp/zeder_to_zotero_importer.cc
```diff
@@ -263,13 +265,23 @@ void DetermineZederEntriesToBeDownloaded(const CommandLineArgs &commandline_args
...
-    if ((key == Config::JournalParams::IniKey::ONLINE_ISSN or key == Config::JournalParams::IniKey::PRINT_ISSN)
-        and not MiscUtil::IsPossibleISSN(value))
-            return false;
+bool GetValidValue(const Config::JournalParams::IniKey &key, std::string * const value) {
+    if (value->empty())
+        return true;
+
+    if (key == Config::JournalParams::IniKey::ONLINE_ISSN or key == Config::JournalParams::IniKey::PRINT_ISSN) {
+        std::vector<std::string> issns;
+        StringUtil::SplitThenTrimWhite(*value, ";", &issns);
```
ruschein, line 274: When the separator is a single character, please prefer the flavour with a char instead of std::string.  It is more efficient.

ruschein (changes requested):

PR #3570, opened 2020-08-26
Added a simple mbox parser.

ruschein:


cpp/lib/include/StringUtil.h
```diff
@@ -1985,6 +1985,11 @@ inline bool ConsistsOfAllASCIIDigits(const std::string &s) {
 }
 
 
+inline bool IsPrintableASCIICharacter(const char ch) {
+    return static_cast<unsigned char>(ch) >= 33 and static_cast<unsigned char>(ch) <= 126;
```
mtrojan-ub, line 1989: Shouldn't a space (32) also be a printable ASCII character?

mtrojan-ub: If not, it should at least be commented

ruschein: Should be ok now...

PR #3572, opened 2020-08-28
cgi_progs/ini_validator.cc

mtrojan-ub:
logger->redirectOutput not working. No error calling it, but the Program will abort and the message will neither be on stdout nor in apache error.log anymore.

### DIFF
--- a/cpp/cgi_progs/ini_validator.cc
+++ b/cpp/cgi_progs/ini_validator.cc
@@ -44,14 +44,14 @@ std::string GetCGIParameterOrDefault(const std::multimap<std::string, std::strin
 
 
 void ShowUploadForm() {
-    std::cout << "<h1>INI validator</h1>\n";
-    std::cout << "<p>Please paste the content of the INI file here:</p>\n";
-    std::cout << "<form id=\"upload_form\" method=\"post\">\n";
-    std::cout << "\t<textarea name=\"ini_content\" rows=\"20\" cols=\"100\"></textarea>\n";
-    std::cout << "\t<input type=\"hidden\" name=\"action\" value=\"validate\">\n";
-    std::cout << "\t<br><br>\n";
-    std::cout << "\t<button onclick=\"document.getElementById(\"upload_form\").submit()\">Validate</button>\n";
-    std::cout << "</form>\n";
+    std::cout << "<h1>INI validator</h1>\n"
+              << "<p>Please paste the contents of the INI file here:</p>\n"
+              << "<form id=\"upload_form\" method=\"post\">\n"
+              << "\t<textarea name=\"ini_content\" rows=\"20\" cols=\"100\"></textarea>\n"
+              << "\t<input type=\"hidden\" name=\"action\" value=\"validate\">\n"
+              << "\t<br><br>\n"
+              << "\t<button onclick=\"document.getElementById(\"upload_form\").submit()\">Validate</button>\n"
+              << "</form>\n";
 }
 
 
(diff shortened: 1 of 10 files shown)


cpp/cgi_progs/ini_validator.cc
```diff
@@ -0,0 +1,95 @@
...
+    std::cout << "<h1>Validate</h1>";
+
+    FileUtil::AutoTempFile temp_file("/tmp/ATF", "", false);
+    std::string ini_content(GetCGIParameterOrDefault(cgi_args, "ini_content", ""));
+    StringUtil::RemoveChars("\r", &ini_content);
+    FileUtil::WriteStringOrDie(temp_file.getFilePath(), ini_content);
+
+    // This does not work yet, seems like due to CGI stdout redirect
+    // std::cout does not have the STDOUT_FILENO anymore,
+    // but how do you get the real fd?
```
ruschein, line 69: You can't.  You can get file descriptors. from FILE objects but not fstreams.

cpp/cgi_progs/ini_validator.cc
```diff
@@ -0,0 +1,98 @@
...
+    if (key_and_value == cgi_args.cend())
+        return default_value;
+
+    return key_and_value->second;
+}
+
+
+void ShowUploadForm() {
+    std::cout << "<h1>INI validator</h1>\n";
+    std::cout << "<p>Please paste the content of the INI file here:</p>\n";
```
ruschein, line 48: I think "contents" sounds more natural.

cpp/cgi_progs/ini_validator.cc
```diff
@@ -0,0 +1,98 @@
...
+    const auto key_and_value(cgi_args.find(parameter_name));
+    if (key_and_value == cgi_args.cend())
+        return default_value;
+
+    return key_and_value->second;
+}
+
+
+void ShowUploadForm() {
+    std::cout << "<h1>INI validator</h1>\n";
```
ruschein, line 47: I think, a single "std::cout <<" would make this way more readable.

cpp/cgi_progs/ini_validator.cc
```diff
@@ -0,0 +1,98 @@
...
+    std::cout << "\t<button onclick=\"document.getElementById(\"upload_form\").submit()\">Validate</button>\n";
+    std::cout << "</form>\n";
+}
+
+
+void Validate(const std::multimap<std::string, std::string> &cgi_args) {
+
+    std::cout << "<h1>Validate</h1>";
+
+    FileUtil::AutoTempFile temp_file("/tmp/ATF", "", false);
```
ruschein, line 62: Please add a comment before "false".

cpp/cgi_progs/ini_validator.cc
```diff
@@ -0,0 +1,98 @@
...
+    std::string ini_content(GetCGIParameterOrDefault(cgi_args, "ini_content", ""));
+    StringUtil::RemoveChars("\r", &ini_content);
+    FileUtil::WriteStringOrDie(temp_file.getFilePath(), ini_content);
+
+    // Redirect stderr to stdout
+    // flush is important, else we have invalid script headers!
+    logger->redirectOutput(STDOUT_FILENO);
+    std::cout << "<font color=\"red\">";
+    std::cout.flush();
+    IniFile ini_file(temp_file.getFilePath());
```
ruschein, line 72: A comment here explaining how we expect a "throw" or other abort would be very useful.

ruschein (changes requested):

cpp/cgi_progs/ini_validator.cc
mtrojan-ub: I will remove it completely, false was set to avoid deletion for debugging purposes.

cpp/cgi_progs/ini_validator.cc
mtrojan-ub: There is a missing change that will be committed soon

cpp/cgi_progs/ini_validator.cc
```diff
@@ -0,0 +1,111 @@
...
+              << "\t<input type=\"hidden\" name=\"action\" value=\"validate\">\n"
+              << "\t<br><br>\n"
+              << "\t<button onclick=\"document.getElementById(\"upload_form\").submit()\">Validate</button>\n"
+              << "</form>\n";
+}
+
+
+void Validate(const std::multimap<std::string, std::string> &cgi_args) {
+
+    std::cout << "<h1>Validate</h1>";
```
ruschein, line 60: I would also add a newline here because o/w when you do a "view source" in a browser, you get a huge mess.

cpp/cgi_progs/ini_validator.cc
```diff
@@ -0,0 +1,111 @@
...
+
+    FileUtil::AutoTempFile temp_file;
+    std::string ini_content(GetCGIParameterOrDefault(cgi_args, "ini_content", ""));
+    StringUtil::RemoveChars("\r", &ini_content);
+    FileUtil::WriteStringOrDie(temp_file.getFilePath(), ini_content);
+
+    // Redirect stderr to stdout
+    // flush is important, else we have invalid script headers!
+    std::cout << "<font color=\"red\">";
```
ruschein, line 69: Ditto, also, if we used newlines we wouldn't need the flush.

mtrojan-ub: I don't think so:
Flush is not always triggered by newline. Especially in web server / cgi context, there are buffer sizes in the web server configuration. Flush forces a flush even if the buffer isn't full. Otherwise, on a huge HTML page with like 400 newlines, the web server would have to send a follow-up TCP segment to the browser over the network after each newline, and that would be pretty inefficient.

PR #3574, opened 2020-09-01
Added batching of database inserts.

ruschein:


cpp/lib/src/DbConnection.cc
```diff
@@ -449,6 +449,61 @@ void DbConnection::insertIntoTableOrDie(const std::string &table_name,
...
+                first = false;
+
+            insert_stmt += escapeAndQuoteString(column_value);
+        }
+        insert_stmt += ')';
+
+        if (column_values != values.cend() - 1)
+            insert_stmt += ',';
+    }
+
```
jriedl, line 499: Wouldn't StringUtil::Join after escapeAndQuoteString on the elements be a more straightforward approach?

ruschein: It would require two additional temporary vector.  The first, let's call it quoted_values and the second one would be the one generated by StringUtil::Join.  I'd rather not do that.

cpp/lib/src/DbConnection.cc
```diff
@@ -449,6 +449,61 @@ void DbConnection::insertIntoTableOrDie(const std::string &table_name,
...
+{
+    if (column_names.empty())
+        LOG_ERROR("at least one column name must be provided!");
+
+    if (not where_clause.empty() and duplicate_key_behaviour != DKB_REPLACE)
+        LOG_ERROR("\"where_clause\" is only valid when using the DKB_REPLACE mode!");
+
+    std::string insert_stmt(duplicate_key_behaviour == DKB_REPLACE ? "REPLACE" : "INSERT");
+    if (duplicate_key_behaviour == DKB_IGNORE)
+        insert_stmt += (type_ == T_MYSQL) ? " IGNORE" : " OR IGNORE";
```
jriedl, line 464: Should errors really generally be ignored or is it rather a "INSERT .. ON DUPLICATE KEY UPDATE" semantic intended?

cpp/lib/src/DbConnection.cc
jriedl: Ok, but doesn't StringUtil::Join() only assemble the vector argument to a string?

PR #3575, opened 2020-09-02
Added missing column values.

ruschein:


cpp/pipelines/collect_journal_stats.cc
```diff
@@ -176,6 +176,9 @@ void ProcessRecords(MARC::Reader * const reader, MARC::Writer * const writer, co
         if (SplitPageNumbers(pages, &start_page, &end_page)) {
             new_column_values.emplace_back(StringUtil::ToString(start_page));
             new_column_values.emplace_back(StringUtil::ToString(end_page));
+        } else {
+            new_column_values.emplace_back("NULL");
```
mtrojan-ub, line 180: Does that really work correctly? i guess that InsertIntoTableOrDie will quote it and insert the string "NULL".

ruschein: Then how do you specify NULL for a column?  Just have commas w/o values?

mtrojan-ub: NULL needs to be written as `NULL` and must not be quoted to `"NULL"`

PR #3581, opened 2020-09-08
Ruschein

ruschein:


cpp/pipelines/krimdok_marc_pipeline.sh
```diff
@@ -120,6 +120,15 @@ MakeFIFO GesamtTiteldaten-post-phase"$PHASE"-"${date}".mrc
 EndPhase || Abort) &
 
 
+# Note: It is necessary to run this phase after articles have had their journal's PPN's inserted!
+MakeFIFO GesamtTiteldaten-post-phase"$PHASE"-"${date}".mrc
```
jriedl, line 124: MakeFIFO immidemately before wait leads to deadlock.

mtrojan-ub: dito & done

PR #3582, opened 2020-09-08
newsletter_uploader.py

mtrojan-ub:


cpp/cgi_progs/Makefile
```diff
@@ -24,6 +24,7 @@ install: $(PROGS)
 	@echo "Installing CGI programs..."
 	mkdir --parents /var/www/cgi-bin
 	cp $(PROGS) /var/www/cgi-bin/
+	cp newsletter_uploader.py /var/www/cgi-bin/
```
ruschein, line 27: Maybe we should make a CGI_DIR variable.

cpp/cgi_progs/newsletter_uploader.py
```diff
@@ -1,30 +1,42 @@
 #!/usr/bin/env python
 import cgi
 import cgitb
+import datetime
 import os
 
 
+print("Content-Type: text/html\n")
```
ruschein, line 8: I think the standard says that HTTP headers have to end in carriage-return line-feed pairs.

cpp/cgi_progs/newsletter_uploader.py
```diff
@@ -1,30 +1,42 @@
...
-    filename = os.path.basename(file_field.filename)
-    open("/usr/local/var/lib/tuelib/newsletters/" + filename, 'w').write(file_field.file.read())
-    message = 'The file "' + filename + '" was uploaded successfully.'
+form = cgi.FieldStorage()
+if "file" not in form:
+    # Show upload form
+    message = """
+        The uploaded file must be a plain-text file.  The first line will be used as the subject of the
+        email message and the remainder as the message body.  The encoding of the file must be UTF-8 with
+        no leading BOM.
```
ruschein, line 19: Let's add "(byte-order marker)" in parentheses because we're usually dealing with muggles.

ruschein (changes requested):

PR #3596, opened 2020-09-21
zotero_cgi: Show docker container logs

mtrojan-ub:


mtrojan-ub:
... still causes right problems in CentOS 8, even if the apache user is added to the docker group.

PR #3600, opened 2020-09-22
zotero_cgi: Expose zotero translation server log file via syslog

mtrojan-ub:


cpp/installer.cc
```diff
@@ -492,10 +492,22 @@ void InstallUBTools(const bool make_install, const OSSystemType os_system_type)
         ExecUtil::ExecOrDie(ExecUtil::LocateOrDie("git"), { "clone", git_url, ZOTERO_ENHANCEMENT_MAPS_DIRECTORY });
     }
 
+    // logfile for zts docker container
+    const std::string ZTS_LOGFILE(UBTools::GetTueFindLogPath() + "/zts.log");
+    if (not FileUtil::Exists(ZTS_LOGFILE))
+        FileUtil::WriteStringOrDie(ZTS_LOGFILE, "");
```
ruschein, line 498: I think we need a FileUtil::Touch...

cpp/installer.cc
```diff
@@ -492,10 +492,22 @@ void InstallUBTools(const bool make_install, const OSSystemType os_system_type)
...
+    if (os_system_type == UBUNTU) {
+        // This is only necessary for UBUNTU since syslogd does not run with root privileges.
+        ExecUtil::ExecOrDie(ExecUtil::LocateOrDie("chown"), { "syslog:adm", ZTS_LOGFILE });
```
ruschein, line 501: Could be rolled into a possible "Touch" function.

docker/zts/systemd/start.sh
```diff
@@ -3,7 +3,10 @@
 LIB_FILE="$(dirname $(readlink --canonicalize "$0"))/../../lib.sh"
 . $LIB_FILE
 
-DOCKER_CMD="docker run --publish 1969:1969"
+# We use syslog as log driver because it's the easiest way to redirect output to a single known file
+# which can later be made accessible via zotero_cgi.
+# (Adding apache user to docker group & calling `docker logs` is not suitable under CentOS 8).
```
ruschein, line 8: "the" docker group...
Do you mean "appropriate" by "suitable"?

ruschein, line 8: "the" apache user...

ruschein (changes requested):

docker/zts/systemd/start.sh
mtrojan-ub: I mean that it is allowed, but it won't work anyway.
https://www.projectatomic.io/blog/2015/08/why-we-dont-let-non-root-users-run-docker-in-centos-fedora-or-rhel/

cpp/installer.cc
mtrojan-ub: Since "Touch" implicates that the function has a similar purpose like the bash command, i'd rather create a FileUtil::ChangeOwnerOrDie function.

cpp/installer.cc
```diff
@@ -673,8 +684,8 @@ void ConfigureApacheUser(const OSSystemType os_system_type, const bool install_s
         ExecUtil::ExecOrDie(ExecUtil::LocateOrDie("sed"),
             { "-i", "s/listen.acl_users = apache,nginx/listen.acl_users = apache,nginx," + username + "/", php_config_filename });
 
-        ExecUtil::ExecOrDie(ExecUtil::LocateOrDie("chown"), { "-R", username + ":" + username, "/var/log/httpd" });
-        ExecUtil::ExecOrDie(ExecUtil::LocateOrDie("chown"), { "-R", username + ":" + username, "/var/run/httpd" });
+        FileUtil::ChangeOwnerOrDie("/var/log/httpd", username, username, true);
```
ruschein, line 687: Please add inline comments before "true".  Also on the following line.

cpp/installer.cc
```diff
@@ -685,7 +696,7 @@ void ConfigureApacheUser(const OSSystemType os_system_type, const bool install_s
     ExecUtil::ExecOrDie(ExecUtil::LocateOrDie("find"),
                         { VUFIND_DIRECTORY + "/local", "-name", "cache", "-exec", "chown", "-R", username + ":" + username, "{}",
                           "+" });
-    ExecUtil::ExecOrDie(ExecUtil::LocateOrDie("chown"), { "-R", username + ":" + username, UBTools::GetTueFindLogPath() });
+    FileUtil::ChangeOwnerOrDie(UBTools::GetTueFindLogPath(), username, username, true);
```
ruschein, line 699: Ditto!

cpp/lib/include/FileUtil.h
```diff
@@ -270,6 +270,13 @@ inline std::string MakeAbsolutePath(const char * const relative_path)
 bool MakeEmpty(const std::string &path);
 
 
+/** \brief Create an empty file or change last modified timestamp of an existing file. */
+void TouchFileOrDie(const std::string &path);
+
+
+void ChangeOwnerOrDie(const std::string &path, const std::string &user, const std::string &group="", bool recursive=true);
```
ruschein, line 277: I'd make the "user" argument optional, too.  That way you can change either the owner or the group, or both.
"recursive" should be const, too!

cpp/lib/src/FileUtil.cc
```diff
@@ -514,6 +515,29 @@ bool MakeEmpty(const std::string &path) {
 }
 
 
+void TouchFileOrDie(const std::string &path) {
+    if (not Exists(path))
```
ruschein, line 519: Please switch the if- and else-branches in order to get rid of the "not".

cpp/lib/src/FileUtil.cc
```diff
@@ -514,6 +515,29 @@ bool MakeEmpty(const std::string &path) {
...
+        WriteStringOrDie(path, "");
+    else
+        OpenForAppendingOrDie(path);
```
ruschein, line 522: This is a nice try.  Implementing "touch" optimally is tricky.  See, for example, https://stackoverflow.com/questions/40630695/linux-modify-file-modify-access-change-time for a link to a man page with useful system calls.  (This may do for now.)

cpp/lib/include/FileUtil.h
```diff
@@ -270,6 +270,13 @@ inline std::string MakeAbsolutePath(const char * const relative_path)
 bool MakeEmpty(const std::string &path);
 
 
+/** \brief Create an empty file or change last modified timestamp of an existing file. */
```
ruschein, line 273: What about the "last accessed" timestamp?

ruschein (changes requested):

cpp/lib/include/FileUtil.h
mtrojan-ub: Well to be absolutely correct, touch changes all 3 timestamps (access, modify, change).
https://stuff-things.net/2015/06/17/unix-timestamps-explained/

cpp/lib/src/FileUtil.cc
```diff
@@ -514,6 +515,29 @@ bool MakeEmpty(const std::string &path) {
...
+    if (recursive)
+        args.emplace_back("-R");
+
+    std::string user_and_group(user);
+    if (not group.empty())
+        user_and_group += ":" + group;
+    args.emplace_back(user_and_group);
+
+    args.emplace_back(path);
+    ExecUtil::ExecOrDie(ExecUtil::LocateOrDie("chown"), args);
```
ruschein, line 537: Do you need to use "chgrp" if there is an empty user?

cpp/lib/src/FileUtil.cc
```diff
@@ -514,6 +515,29 @@ bool MakeEmpty(const std::string &path) {
...
+void TouchFileOrDie(const std::string &path) {
+    if (Exists(path))
+        OpenForAppendingOrDie(path);
+    else
+        WriteStringOrDie(path, "");
+}
+
+
+void ChangeOwnerOrDie(const std::string &path, const std::string &user, const std::string &group, const bool recursive) {
+    std::vector<std::string> args;
```
ruschein, line 527: You might want to add a LOG_ERROR here if user and group are both empty.

ruschein (changes requested):

cpp/lib/src/FileUtil.cc
```diff
@@ -514,6 +515,29 @@ bool MakeEmpty(const std::string &path) {
...
+    if (recursive)
+        args.emplace_back("-R");
```
ruschein, line 529: "--recursive"

cpp/lib/src/FileUtil.cc
```diff
@@ -514,6 +515,29 @@ bool MakeEmpty(const std::string &path) {
...
+
+    std::string user_and_group(user);
+    if (not group.empty())
+        user_and_group += ":" + group;
```
ruschein, line 533: No colon when the user is empty and "chgrp" instead of "chown".

cpp/lib/src/FileUtil.cc
mtrojan-ub: No, if the parameter given to chown starts with `:` it will only change the group.

PR #3603, opened 2020-09-23
Ruschein

ruschein:


cpp/lib/src/MBox.cc
```diff
@@ -33,6 +33,16 @@ MBox::Message &MBox::Message::swap(Message &other_message) {
 }
 
 
+std::string MBox::Message::toString() const {
+    return   "reception_time: " + TimeUtil::TimeTToString(reception_time_) + "\n"
+           + "original_host:  " + original_host_ + "\n"
+           + "sender:         " + sender_ + "\n"
+           + "subject:        " + subject_ + "\n\n"
+           + message_body_;
+
```
mtrojan-ub, line 42: superfluous newline

ruschein: Removed!

cpp/email_watcher.cc
```diff
@@ -0,0 +1,258 @@
...
+    std::set<std::string> matched_section_names;
+    for (const auto &email_message : mbox) {
+        ++email_message_count;
+        for (const auto &[section_name, email_description] : email_descriptions) {
+            if (email_description.subjectAndBodyMatched(email_message))
+                matched_section_names.emplace(section_name);
+            else
+                unmatched_emails.emplace_back(email_message.toString());
+        }
+    }
```
mtrojan-ub, line 203: I think it would be useful to always forward messages hat have either originally been sent with high priority or contain a pre-defined custom header.

cpp/data/email_watcher.conf
```diff
@@ -0,0 +1,20 @@
+# Each section must provide the following entries:
+# from_host           -- where the emails are originating
+# subject_patttern    -- a PCRE regex that must match the emails' subjects
```
mtrojan-ub, line 3: patttern with 3t

cpp/email_watcher.cc
```diff
@@ -0,0 +1,258 @@
...
+#include "UBTools.h"
+#include "util.h"
+
+
+namespace {
+
+
+struct EmailDescription {
+    std::string from_host_;
+    RegexMatcher *subject_pattern_;
```
mtrojan-ub, line 45: subject_pattern_ and email_body_pattern_ do not contain a pattern => they contain a matcher, so the variables should be named subject_matcher_ and email_body_matcher_

mtrojan-ub: and maybe they should be unified to subject_matcher_ + body_matcher_ or email_subject_matcher_ + email_body_matcher_

cpp/data/email_watcher.conf
```diff
@@ -0,0 +1,20 @@
...
+# email_body_pattern  -- a PCRE that must match the emails' bodies
```
mtrojan-ub, line 4: it might be useful to not only have patterns that must match, but also have patterns that must not match.

E.g. if we have a message with "pipeline completed successfully" which also contains a warning at the same time, we might want to forward that message anyway.

cpp/email_watcher.cc
```diff
@@ -0,0 +1,258 @@
...
+    const IniFile ini_file(argv[1]);
+
+    std::vector<std::string> notification_email_addresses;
+    if (StringUtil::SplitThenTrimWhite(ini_file.getString("", "notify"), ',', &notification_email_addresses) == 0)
+        LOG_ERROR("we need at least one notification email address in the \"notify\" entry in \""
+                  + ini_file.getFilename() + "\"!");
+
+    const std::string backup_dir_path(ini_file.getString("", "backup_dir_path") + "/");
+
+    const auto overdue_db(OpenOrCreateOverdueDatabase());
```
mtrojan-ub, line 187: Is this connection actually used anywhere?

cpp/email_watcher.cc
ruschein: I think dealing w/ custom headers is difficult and also not necessary.  There are several, possibly many, custom headers in use by email systems.

cpp/email_watcher.cc
ruschein: Apparently not.

cpp/data/email_watcher.conf
```diff
@@ -0,0 +1,26 @@
+# Each section must provide the following entries:
+# from_host                    -- where the emails are originating
+# subject_pattern              -- a PCRE regex that must match the emails' subjects
+# email_body_positive_pattern  -- a PCRE that must match the emails' bodies
```
mtrojan-ub, line 4: Same as variables:
One pattern is called "subject_pattern" so the others should be called e.g "body_positive_pattern" instead of "email_body_positive pattern"

cpp/data/email_watcher.conf
```diff
@@ -0,0 +1,26 @@
...
+# Where to move the mbox files after processing:
+backup_dir_path=/tmp
+
+# Forward all emails with a priority like this or higher
+forward_priority = 5
+
+[FoxHunt]
+from_host                    = goldweb.com.au
+subject_pattern              = .*
+email_body_positive_pattern  = .*
```
mtrojan-ub, line 25: same

cpp/email_watcher.cc
```diff
@@ -0,0 +1,263 @@
...
+    std::string error_message;
+    subject_matcher_ = RegexMatcher::RegexMatcherFactory(subject_pattern_regex, &error_message);
+    if (subject_matcher_ == nullptr)
+        LOG_ERROR("bad regex \"" + subject_pattern_regex + "\" in \"" + ini_file_section.getSectionName() + "\"!");
+
+    if (not ini_file_section.hasEntry("email_body_pattern"))
+        LOG_ERROR("ini file section \"" + ini_file_section.getSectionName()
+                  + "\" is missing an \"email_body_pattern\" entry!");
+
+    const auto email_body_positive_pattern_regex(ini_file_section.getString("email_body_positive_pattern", ""));
```
mtrojan-ub, line 77: same thing here with "email_"-Prefix

cpp/email_watcher.cc
mtrojan-ub: That's why i propose defining our own like e.g. "X-UB-ERROR" which could be added if any program exits with LOG_ERROR and then be automatically redirected whenever this header is found.

cpp/email_watcher.cc
mtrojan-ub: Seems you removed it now.

In which interval will this program be called? Is it necessary to prevent it from sending multiple overdue reminders for the same mail?

cpp/lib/src/MBox.cc
```diff
@@ -28,11 +29,22 @@ MBox::Message &MBox::Message::swap(Message &other_message) {
...
     other_message.sender_.swap(sender_);
     other_message.subject_.swap(subject_);
+    std::swap(other_message.priority_, priority_);
     other_message.message_body_.swap(message_body_);
     return *this;
 }
 
 
+std::string MBox::Message::toString() const {
+    return   "reception_time: " + TimeUtil::TimeTToString(reception_time_) + "\n"
```
mtrojan-ub, line 39: multiple spaces after return statement

cpp/data/email_watcher.conf
```diff
@@ -0,0 +1,26 @@
...
+#                                 matching the other criteria
+
+# The following is a comma-separated list of email address
+notify=johannes.ruscheinski@uni-tuebingen.de
+
+# Where to move the mbox files after processing:
+backup_dir_path=/tmp
+
+# Forward all emails with a priority like this or higher
+forward_priority = 5
```
mtrojan-ub, line 20: maybe 4 would be better (anything higher than 3=normal)

cpp/lib/src/MBox.cc
```diff
@@ -152,6 +164,7 @@ MBox::Message MBox::getNextMessage() const {
     }
 
     std::string sender, original_host, subject;
+    int priority(-1);
```
mtrojan-ub, line 167: since allowed priorities are 1-5 we could also use an unsigned with 0 as default

cpp/email_watcher.cc
ruschein: It will probably run every 10 minutes or so.  And no, because we update the last sent time when we complain.

cpp/lib/src/MBox.cc
ruschein: Intentional in order to line things up vertically.

cpp/data/email_watcher.conf
ruschein: This is not our actual config file but just one I use for testing.  I only committed it because I wanted you to have it for documentation purposes.

mtrojan-ub: Then maybe rename it to .conf.example?

cpp/lib/src/FileUtil.cc
```diff
@@ -1476,4 +1491,81 @@ bool GetMostRecentlyModifiedFile(const std::string &directory_path, std::string
...
+
+    const int fd(::open(path.c_str(), O_LARGEFILE|O_RDWR));
+    if (fd == -1)
+        LOG_ERROR("failed to open(2) \"" + path + "\"!");
+
+    void * const map_start(::mmap(nullptr, original_file_size, PROT_READ|PROT_WRITE, MAP_SHARED, fd, 0));
+    if (map_start == MAP_FAILED or map_start == nullptr)
+        LOG_ERROR("Failed to mmap(2) \"" + path + "\"!");
+
+    ::memmove(map_start, static_cast<void *>(static_cast<char *>(map_start) + no_of_bytes), original_file_size - no_of_bytes);
```
mtrojan-ub, line 1516: What happens if the log file is written between mmap and memmove?

cpp/log_rotate.cc
```diff
@@ -34,51 +35,14 @@ namespace {
...
-    std::unique_ptr<File> output(FileUtil::OpenOutputFileOrDie(temp_file.getFilePath()));
-    CopyLines(input.get(), output.get());
-    input->close(), output->close();
-    if (not FileUtil::RenameFile(temp_file.getFilePath(), filename, /* remove_target = */ true))
-        logger->error("in KeepLines: failed to rename \"" + temp_file.getFilePath() + "\" to \"" + filename + "\"!");
+[[noreturn]] void Usage() {
+    ::Usage("[--verbose] [--max-rotations=max_rotations|--no-of-lines-to-keep=max_line_count] [--recreate] directory file_regex\n"
+            "where the default for \"max_rotations\" is " + std::to_string(DEFAULT_MAX_ROTATIONS) + "\n"
+            "if \"--recreate\" has been specified the original filename will be recreated with same owner, group,\n"
+            "and, if appropriate, SELinux secucrity context.\n"
```
mtrojan-ub, line 42: secucrity

mtrojan-ub: And since "recreate" does not really describe what happens, the flag should be renamed

mtrojan-ub: How is this program meant to be used in the crontab? Do we need 2 calls? (1 for all logs except zts.log without --recreate, 1 just for zts.log with recreate?)

cpp/email_watcher.cc
ruschein: We can postpone this or just use priority 5.

cpp/lib/src/FileUtil.cc
ruschein: We're up shit creek or, I have to go back to the drawing board! :-(

cpp/log_rotate.cc
ruschein: Well, "recreate" is correct.  The original files are being renamed.  And yes, we would need two calls.

PR #3611, opened 2020-09-29
Mtrojan

mtrojan-ub:


cpp/zts_harvester_delivery_pipeline.sh
```diff
@@ -112,43 +112,43 @@ LOGGER_FORMAT=no_decorations,strip_call_site \
...
 EndPhase
 
 
 StartPhase "Validate Generated Records"
-cd $HARVESTER_OUTPUT_DIRECTORY
+cd "$HARVESTER_OUTPUT_DIRECTORY"
 counter=0
 shopt -s nullglob
 for d in */ ; do
     d=${d%/}
```
ruschein, line 128: Maybe the right hand side of this assignment should also be in double quotes?

cpp/zts_harvester_delivery_pipeline.sh
```diff
@@ -112,43 +112,43 @@ LOGGER_FORMAT=no_decorations,strip_call_site \
...
-    if [[ $d -ef $HARVESTER_OUTPUT_DIRECTORY ]]; then
+    if [[ $d -ef "$HARVESTER_OUTPUT_DIRECTORY" ]]; then
```
ruschein, line 129: $d should be in double quotes.

ruschein (changes requested):

PR #3612, opened 2020-09-29
Ruschein

ruschein:


cpp/lib/include/ProcessUtil.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include <string>
+#include <unordered_set>
+#include <sys/types.h>
+
+
+namespace ProcessUtil {
+
+
+/** \return  The list of PID's that have an open file descripto for "path".
```
mtrojan-ub, line 36: descripto

cpp/lib/include/ProcessUtil.h
```diff
@@ -0,0 +1,43 @@
+/** \file    ProcessUtil.h
+ *  \brief   The ProcessUtil interface.
+ *  \author  Dr. Johannes Ruscheinski
+ */
+
+/*
+ *  Copyright 2020 Universitätsbibliothek Tübingen
+ *
+ *  This file is part of the libiViaCore package.
```
mtrojan-ub, line 9: really?

mtrojan-ub: please also check the comment blocks below that also reference to libiViaCore

cpp/lib/include/ProcessUtil.h
```diff
@@ -0,0 +1,43 @@
+/** \file    ProcessUtil.h
+ *  \brief   The ProcessUtil interface.
+ *  \author  Dr. Johannes Ruscheinski
+ */
+
+/*
+ *  Copyright 2020 Universitätsbibliothek Tübingen
```
mtrojan-ub, line 7: University Library of Tübingen

cpp/lib/src/ProcessUtil.cc
```diff
@@ -0,0 +1,52 @@
+/** \file    ProcessUtil.c
+ *  \brief   The ProcessUtil implementation.
+ *  \author  Dr. Johannes Ruscheinski
+ */
+
+/*
+ *  Copyright 2020 Universitätsbibliothek Tübingen
```
mtrojan-ub, line 7: University Library of Tübingen

cpp/lib/src/ProcessUtil.cc
```diff
@@ -0,0 +1,52 @@
...
+ *
+ *  This file is part of the libiViaCore package.
```
mtrojan-ub, line 9: Also the whole libiViaCore stuff

mtrojan-ub: By the way, do we really need a license statement in each file? In the meantime it is possible to set a license for a github repository.

cpp/test/GetProcessIdsForPath_test.cc
```diff
@@ -0,0 +1,26 @@
+// Test harness for ProcessUtil::GetProcessIdsForPath.
+#include "FileUtil.h"
+#include "ProcessUtil.h"
+#include "util.h"
+
+
+int Main(int, char **) {
+    const std::string TEMPFILE_NAME("/tmp/GetProcessIdsForPath_test.temp");
```
mtrojan-ub, line 8: use FileUtil::AutoTempFile instead?

mtrojan-ub (changes requested):
see comments

cpp/test/GetProcessIdsForPath_test.cc
ruschein: I intentionally did not use it to make what's happening more explicit.

cpp/lib/src/ProcessUtil.cc
ruschein: Well, maybe we don't

cpp/lib/src/ProcessUtil.cc
```diff
@@ -0,0 +1,49 @@
...
+
+namespace ProcessUtil {
+
+
+std::unordered_set<pid_t> GetProcessIdsForPath(const std::string &path) {
+    std::unordered_set<pid_t> pids;
+
+    FileUtil::Directory proc("/proc", "\\d+"); // Only PID's
+    for (const auto &pid_dir : proc) {
+        FileUtil::Directory fd_dir("/proc/" + pid_dir.getName() + "/fd");
```
mtrojan-ub, line 35: due to pid_dir.getName() documentation, it returns the full name of the entry including the directory path, so prefixing it with "/proc" should not be necessary.

cpp/lib/src/ProcessUtil.cc
```diff
@@ -0,0 +1,49 @@
...
+        for (const auto &link : fd_dir) {
+            std::string link_target;
+            if (FileUtil::ReadLink("/proc/" + pid_dir.getName() + "/fd/" + link.getName(), &link_target) and link_target == path) {
```
mtrojan-ub, line 38: Same here

cpp/log_rotate.cc
```diff
@@ -117,6 +127,9 @@ int Main(int argc, char *argv[]) {
                                                FileUtil::GroupnameFromGID(gid));
                     if (SELinuxUtil::IsEnabled())
                         SELinuxUtil::FileContext::ApplyChanges(filename);
+
+                    for (const auto pid : pids)
+                        ::kill(pid, SIGHUP);
```
mtrojan-ub, line 132: Is it necessary to send SIGHUP also to ourself?
If not, it might be useful if ProcessUtil::GetProcessIdsForPath would take an additional parameter to include or exclude the process itself from the result.

cpp/log_rotate.cc
```diff
@@ -98,10 +101,17 @@ int Main(int argc, char *argv[]) {
...
-            if (max_line_count > 0)
-                FileUtil::OnlyKeepLastNLines(directory_path + "/" + entry.getName(), max_line_count);
-            else {
-                const std::string filename(directory_path + "/" + entry.getName());
+
+            const std::string filename(directory_path + "/" + entry.getName());
+            if (max_line_count > 0) {
+                const int fd(::open(filename.c_str(), O_RDWR));
+                { // New scope to ensure that the FileLocker instance goes out of scope before we close fd!
+                    FileLocker file_locker(fd, FileLocker::WRITE_ONLY);
```
mtrojan-ub, line 109: FileUtil::OnlyKeepLastNLines opens it with O_RDONLY to calculate no_of_reading_bytes_to_remove, then it passes it on to ::truncate later which needs write access.

FileLocker only supports READ_ONLY and WRITE_ONLY, maybe we need a READ_WRITE flag?

cpp/lib/src/FileLocker.cc
```diff
@@ -34,22 +35,36 @@
...
     struct flock lock_struct;
     lock_struct.l_type   = static_cast<short>((lock_type == READ_ONLY) ? F_RDLCK : F_WRLCK); /* F_RDLCK, F_WRLCK, F_UNLCK    */
     lock_struct.l_whence = SEEK_SET;                                                         /* SEEK_SET, SEEK_CUR, SEEK_END */
     lock_struct.l_start  = 0;                                                                /* Offset from l_whence         */
     lock_struct.l_len    = 0;                                                                /* length, 0 = to EOF           */
     lock_struct.l_pid    = ::getpid();                                                       /* our PID                      */
 
-    if (::fcntl(lock_fd_, F_SETLKW, &lock_struct) == -1)
-        throw std::runtime_error("in FileLocker::FileLocker: fcntl(2) failed! (" + std::to_string(errno) + ")!");
+    if (::fcntl(lock_fd_, F_SETLKW, &lock_struct) == -1) {
```
mtrojan-ub, line 53: From the manpage:
F_SETLKW: As for F_SETLK, but if a conflicting lock is held on the file, then wait for that lock to be released.

I don't see our timeout variable being used here?

cpp/log_rotate.cc
mtrojan-ub: Or maybe WRITE_ONLY would be correct and should be renamed to READ_WRITE?

cpp/lib/src/FileLocker.cc
```diff
@@ -34,22 +35,36 @@
...
+        const int last_errno(errno);
+        ::alarm(0); // Cancel our alarm!
+        if (last_errno != EINTR)
+            throw std::runtime_error("fcntl(2) failed! (" + std::to_string(last_errno) + ")!");
+        errno = 0;
+        interrupted_ = true;
```
mtrojan-ub, line 59: So this will be used later to prevent an unlock.

But if we get here this also means that the locking has never been complete, so how do we guarantee a file is locked after this function ends? Shouldn't it be called in a loop while the locking hasn't been successful?

cpp/lib/src/util.cc
```diff
@@ -25,15 +25,37 @@
...
+    ::close(log_fd);
+    const int new_fd(::open(log_path.c_str(), O_WRONLY|O_APPEND));
+    if (new_fd == -1) { // We're screwed!
+        const std::string error_message("in HUPHandler(util.cc): can't open(2) \"" + log_path + "\"!\n");
+        ::write(STDERR_FILENO, error_message.c_str(), error_message.size());
+        ::exit(EXIT_FAILURE);
+    }
+
+    // Make sure we're using the same file descriptor as before!
+    ::dup2(new_fd, log_fd);
```
mtrojan-ub, line 55: what if ::dup2 fails?

cpp/lib/src/util.cc
```diff
@@ -64,6 +89,13 @@ Logger::Logger()
 }
 
 
+void Logger::redirectOutput(const int new_fd) {
```
mtrojan-ub, line 92: In the current version in ruschein branch, redirectOutput is directly implemented in util.h which is not part of this pull request. Did you forget to commit it?

cpp/lib/src/util.cc
```diff
@@ -25,15 +25,37 @@
...
+#include "FileLocker.h"
+#include "FileUtil.h"
 #include "MiscUtil.h"
+#include "SignalUtil.h"
 #include "StringUtil.h"
 #include "TimeUtil.h"
 
 
+static std::string log_path;
+static int log_fd;
```
mtrojan-ub, line 42: 1) What's the difference between log_fd and fd_?
2) It might be clearer to either move these variables to util.h (next to fd_) declaration, or to convert fd_ to a static variable as well

cpp/lib/src/util.cc
mtrojan-ub: and what happens with fd_? maybe just call redirectOutput(new_fd) for consistency?

cpp/lib/src/util.cc
```diff
@@ -44,11 +66,14 @@ char *progname; // Must be set in main() with "progname = argv[0];";
 const std::string Logger::FUNCTION_NAME_SEPARATOR(" --> ");
 
 
-
 Logger::Logger()
     : fd_(STDERR_FILENO), log_process_pids_(false), log_no_decorations_(false), log_strip_call_site_(false),
       min_log_level_(LL_INFO)
 {
+    log_fd = fd_;
```
mtrojan-ub, line 73: How about calling redirectOutput (fd_) instead?

cpp/lib/src/util.cc
ruschein: I did indeed! :-(

cpp/lib/src/util.cc
ruschein: Good point.  I eliminated fd_.

cpp/lib/src/ProcessUtil.cc
ruschein: That documentation is incorrect! :-( (I fixed it.)

cpp/log_rotate.cc
ruschein: FileUtil::OnlyKeepLastNLines calls RemoveLeadingBytes which opens the file with O_RDWR before calling ftruncate.
As far as the locking is concerned that is my reading of what the underlying fcntl documentation states.

cpp/lib/src/FileLocker.cc
ruschein: That's the alarm call.

cpp/lib/include/FileUtil.h
```diff
@@ -127,7 +127,7 @@ class Directory {
     public:
         Entry(const Entry &other);
 
-        /** \return the full name of the entry, i.e., including the directory path. */
+        /** \return the name of the entry w/o the directory path. */
```
mtrojan-ub, line 130: Then it might be good if we had a GetFullName() function, so we do not need to build the full name again by hand.

git-config/hooks/post-merge
```diff
@@ -46,3 +46,9 @@ if [[ $compare_against_installer_schema && -r /tmp/new_ub_tools.schema ]]; then
         echo "Verification successful! (The current DB schema was successfully compared to the installer's DB schema.)"
     fi
 fi
+
+
+SYSTEM_UPDATER=/usr/local/bin/system_updater
+if [ -e "$SYSTEM_UPDATER" ]; then
+    "$SYSTEM_UPDATER"
```
mtrojan-ub, line 53: MYSQL_PATCHER takes the directory $MYSQL_SCHEMA_UPDATES_PATH as argument.
How about implementing SYSTEM_UPDATER in the same way?

cpp/log_rotate.cc
ruschein: Fixed.

cpp/lib/src/util.cc
ruschein: Obsolete because fd_ does not exist any longer.

cpp/lib/include/FileUtil.h
ruschein: It exist now.

git-config/hooks/post-merge
ruschein: Done!

cpp/lib/src/ProcessUtil.cc
```diff
@@ -0,0 +1,50 @@
...
+std::unordered_set<pid_t> GetProcessIdsForPath(const std::string &path) {
+    std::unordered_set<pid_t> pids;
+
+    FileUtil::Directory proc("/proc", "\\d+"); // Only PID's
+    for (const auto &pid_dir : proc) {
+        const std::string FD_DIR_PATH(pid_dir.getFullName() + "/fd");
+        FileUtil::Directory fd_dir(FD_DIR_PATH);
+        for (const auto &link : fd_dir) {
+            std::string link_target;
+            if (FileUtil::ReadLink(FD_DIR_PATH + "/" + link.getName(), &link_target) and link_target == path) {
```
mtrojan-ub, line 39: getFullName() could be used here as well

cpp/system_updater.cc
```diff
@@ -0,0 +1,95 @@
...
+    }
+    if (script_names.empty()) {
+        LOG_INFO("nothing to be done!");
+        return EXIT_SUCCESS;
+    }
+
+    std::sort(script_names.begin(), script_names.end(), ScriptLessThan);
+
+    for (const auto &script_name : script_names)
+        ExecUtil::ExecOrDie(SYSTEM_UPDATES_DIR + "/" + script_name);
```
mtrojan-ub, line 89: Maybe some more LOG_INFO calls might be useful
- when executing a script
- when completing the operation / writing highest_version_number to VERSION_PATH

cpp/lib/include/SysLog.h
```diff
@@ -0,0 +1,57 @@
...
+    // \brief Write a log entry.
+    void log(const LogLevel level, const std::string &message);
+
+    // \return An or'rd together set of the current log levels.
+    int getLogLevels() const { return ::setlogmask(0); }
+
+    // \brief  Sets a new set of levels that will be logged unless "new_levels" is 0.
+    // \param  new_levels An or'rd together set of the levels that wile be logged after the call.
+    // \return An or'rd together set of the current log levels..
+    int setLogLevsls(const int new_levels) { return ::setlogmask(new_levels); }
```
mtrojan-ub, line 56: typo => setLogLevsls

cpp/data/rsyslog.conf
```diff
@@ -0,0 +1 @@
+local0.*  /usr/local/var/log/tuefind/syslog.log
```
mtrojan-ub, line 1: please move conf to cpp/data/installer + rename e.g. to syslog.ub_tools.conf

cpp/data/system_updates/1.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+install --owner=root --group=root --mode=644 /usr/local/ub_tools/cpp/data/rsyslog.conf /etc/rsyslog.d/ub_tools.conf
```
mtrojan-ub, line 2: this should also be performed by installer.cc & system_version should be detected by installer

cpp/lib/src/SysLog.cc
```diff
@@ -0,0 +1,38 @@
+/** \file    SysLog.h
```
mtrojan-ub, line 1: This is SysLog.cc, not SysLog.h

What's the meaning of the \file tag except redundant information? Guess i will never understand.

cpp/lib/src/SysLog.cc
```diff
@@ -0,0 +1,38 @@
...
+ *  \brief   Declaration of class SysLog.
```
mtrojan-ub, line 2: I guess this should be "Implementation" and not "Declaration", which is also redundant because we know it due to the file extension...

cpp/syslog.cc
```diff
@@ -0,0 +1,123 @@
+/** \brief A tool for logging to syslogd in a shell script.
```
mtrojan-ub, line 1: Could there be conflicts with any other program named syslog in PATH?

cpp/log_rotate.cc
mtrojan-ub: Again: Why is the flag called WRITE_ONLY if, in fact, we use it to both reading + writing?

cpp/log_rotate.cc
```diff
@@ -117,6 +127,12 @@ int Main(int argc, char *argv[]) {
                                                FileUtil::GroupnameFromGID(gid));
                     if (SELinuxUtil::IsEnabled())
                         SELinuxUtil::FileContext::ApplyChanges(filename);
+
+                    const pid_t our_pid(::getpid());
+                    for (const auto pid : pids) {
+                        if (pid != our_pid)
```
mtrojan-ub, line 133: GetProcessIdsForPath() sould take an additional bool to exclude the current pid, then we wouldn't have to test for our_pid here explicitly

cpp/syslog.cc
ruschein: No, if at all the demon is called syslogd and, I checked, there is no system tool called syslog.

cpp/merge_print_and_online.cc
```diff
@@ -711,7 +711,9 @@ bool MergeFieldPair245(const MARC::Record::Field &merge_field, const MARC::Recor
         if (FuzzyEqual(varying_title_field.getFirstSubfieldWithCode('a'), import_title))
             return true;
     }
-    merge_record->insertField("246", { { 'a', import_title },
+
+    const bool import_title_is_newer(import_record.getPublicationYear() > merge_record->getPublicationYear());
+    merge_record->insertField("246", { { 'a', import_title_is_newer? import_title : merge_title },
```
mtrojan-ub, line 716: Missing space before '?'

PR #3615, opened 2020-10-06
CGI based restart of zts service

jriedl:


cpp/cgi_progs/restart_zts.cc
```diff
@@ -0,0 +1,112 @@
...
+void RestartZTS() {
+    SendHeaders();
+    std::cout << "<h2>Attempting restart of ZTS...</h2>\n" << std::endl;
+    bool log_no_decorations_old(logger->getLogNoDecorations());
+    bool log_strip_call_site_old(logger->getLogStripCallSite());
+    logger->setLogNoDecorations(true);
+    logger->setLogStripCallSite(true);
+    logger->redirectOutput(STDOUT_FILENO);
+    try {
+        ExecUtil::ExecOrDie("/usr/bin/sudo", { "systemctl", "restart", "zts" });
```
mtrojan-ub, line 84: - Do we need apache user be enabled in sudoers for this? 
- (Is this really a good way regarding security?)
- Are there any additional SELinux settings that need to be set?
- Does the installer need to be changed as well?

cpp/cgi_progs/restart_zts.cc
```diff
@@ -0,0 +1,112 @@
...
+    logger->setLogStripCallSite(log_strip_call_site_old);
+}
+
+
+} // end unnamed namespace
+
+int Main(int /*argc*/, char */*argv*/[]) {
+    std::multimap<std::string, std::string> cgi_args;
+    if (RestartActionPresent(&cgi_args)) {
+        DisplayRestartButton();
```
mtrojan-ub, line 102: I guess that DisplayRestartButton() and RestartZTS() calls should be swapped

cpp/cgi_progs/restart_zts.cc
jriedl: ad 1) Yes, because restart has to be run with root privileges.
ad 2) Security is higher than using SUID, especially since it is restricted to a specific user with specific commands
ad 3) Runs on Ubuntu. We must see whether this poses further problems on CentOS.
ad 4) Already Included.

cpp/data/installer/sudo.zts-restart
```diff
@@ -0,0 +1,4 @@
+# Allow apache to restart the zts service
+Cmnd_Alias ZTS_RESTART = /bin/systemctl restart zts
+Cmnd_Alias ZTS_STATUS = /bin/systemctl status zts
+vufind  ALL = NOPASSWD: ZTS_RESTART, ZTS_STATUS
```
mtrojan-ub, line 4: Apache will only run as user `vufind` if we have a vufind installation. On ub-tools-only under CentOS the username is `apache`, if we might have any constellation where ub-tools-only is installed under Ubuntu it might also be "www-data".

jriedl: Additional potential users have been included.

PR #3616, opened 2020-10-06
Implemented new strategy for the elimination of dangling links.

ruschein:


cpp/merge_print_and_online.cc
```diff
@@ -205,47 +205,55 @@ void EliminateDanglingOrUnreferencedCrossLinks(const bool debug, const std::unor
...
-                    if (ppn_to_offset_map.find(ppn) == ppn_to_offset_map.end()) {
-                        LOG_INFO("Don't merge group around PPN " + ppn + " because the PPN is missing in our data! All PPNs in group: "
-                                 + StringUtil::Join(group_ppns, ','));
-                        drop_group = true;
-                        break;
-                    }
-                }
-            }
+        if (ppn_to_offset_map.find(canonical_ppn_and_ppn->second) != ppn_to_offset_map.end())
+            canonical_ppn_and_ppn = std::next(canonical_ppn_and_ppn, 1);
```
mtrojan-ub, line 216: 1 is default and can be omitted
https://en.cppreference.com/w/cpp/iterator/next

cpp/merge_print_and_online.cc
```diff
@@ -205,47 +205,55 @@ void EliminateDanglingOrUnreferencedCrossLinks(const bool debug, const std::unor
...
+    //
 
-                dropped_count += group_ppns.size() + 1;
-                while (canonical_ppn_to_ppn_map->find(canonical_ppn) != canonical_ppn_to_ppn_map->end())
-                    canonical_ppn_and_ppn = canonical_ppn_to_ppn_map->erase(canonical_ppn_to_ppn_map->find(canonical_ppn));
-            }
+    // 2a, determine what the new canical PPN's are in those cases where we have to
+    //     swap them out.
+    std::unordered_map<std::string, std::string> old_canonical_ppn_to_new_canonical_ppn_map;
+    for (const auto &[current_canonical_ppn, non_canonical_ppn] : *canonical_ppn_to_ppn_map) {
```
mtrojan-ub, line 230: Do we need some kind of recursion here?

ruschein: No!

PR #3637, opened 2020-10-27
Riedl

jriedl:


cpp/lib/src/Downloader.cc
```diff
@@ -315,7 +315,7 @@ const std::string &Downloader::getLastErrorMessage() const {
 
 unsigned Downloader::getResponseCode() {
     std::string err_msg;
-    const std::string regex_pattern("HTTP(/\\d\\.\\d)?\\s*(\\d{3})\\s*");
+    const std::string regex_pattern("HTTP(?:/\\d(?:\\.(?:\\d)?)?)?\\s+(\\d{3})");
```
mtrojan-ub, line 318: This pattern would also support something like this: `HTTP/1. 200`
I guess we could exclude this case & therefore shorten the pattern a little bit: `HTTP(?:/\\d(?:\\.\\d)?)?\\s+(\\d{3})`

jriedl: But then `HTTP/1 200` would match.

mtrojan-ub: with your first pattern it would match too

jriedl: Yes, but since apparently there is no easy approach to only match correct version strings and the practical impact is negligible it would just go with the current approach.

PR #3639, opened 2020-10-27
zotaut changes

mtrojan-ub:
i will update wiki after PR is merged

### DIFF
--- a/cpp/lib/src/ZoteroHarvesterConversion.cc
+++ b/cpp/lib/src/ZoteroHarvesterConversion.cc
@@ -1007,6 +1007,24 @@ std::string CalculateMarcRecordHash(const MARC::Record &marc_record) {
 }
 
 
+const std::vector<std::string> UNDESIRED_ITEM_TYPES {
+    "webpage"
+};
+
+
+bool ExcludeUnsesiredItemTypes(const MetadataRecord &metadata_record) {
+    if (std::find(UNDESIRED_ITEM_TYPES.begin(),
+                  UNDESIRED_ITEM_TYPES.end(),
+                  metadata_record.item_type_) != UNDESIRED_ITEM_TYPES.end())
+    {
+        LOG_DEBUG("Skipping: undesired item type \"" + metadata_record.item_type_ + "\"");
+        return true;
+    }
+
+    return false;
+}
+
+
 const std::vector<std::string> VALID_ITEM_TYPES_FOR_ONLINE_FIRST {
     "journalArticle", "magazineArticle", "review"
 };
@@ -1089,7 +1107,9 @@ void ConversionTasklet::run(const ConversionParams &parameters, ConversionResult
             if (new_metadata_record.url_.empty())
                 throw std::runtime_error("no URL set");
 
-            if (ExcludeOnlineFirstRecord(new_metadata_record, parameters)) {
+            if (ExcludeUnsesiredItemTypes(new_metadata_record)) {
+                ++result->num_skipped_since_undesired_item_type_;
+            } else if (ExcludeOnlineFirstRecord(new_metadata_record, parameters)) {
                 ++result->num_skipped_since_online_first_;
                 continue;
             } else if (ExcludeEarlyViewRecord(new_metadata_record, parameters)) {
--- a/cpp/lib/src/ZoteroHarvesterConversion.cc
+++ b/cpp/lib/src/ZoteroHarvesterConversion.cc
@@ -997,7 +997,7 @@ bool MarcRecordMatchesExclusionFilters(const Util::HarvestableItem &download_ite
 }
 
 
-const std::set<MARC::Tag> EXCLUDED_FIELDS_DURING_CHECKSUM_CALC {
+const std::set<MARC::Tag> EXCLUDED_FIELDS_DURING_CHECKSUM_CALC{
     "001", "URL", "ZID", "JOU",
 };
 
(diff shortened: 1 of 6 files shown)


cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -1007,6 +1007,24 @@ std::string CalculateMarcRecordHash(const MARC::Record &marc_record) {
 }
 
 
+const std::vector<std::string> UNDESIRED_ITEM_TYPES {
```
ruschein, line 1010: There should be no space before the opening curly bracket.

cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -1007,6 +1007,24 @@ std::string CalculateMarcRecordHash(const MARC::Record &marc_record) {
...
+    "webpage"
+};
+
+
+bool ExcludeUnsesiredItemTypes(const MetadataRecord &metadata_record) {
```
ruschein, line 1015: This could probably be static.

cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -1089,7 +1107,9 @@ void ConversionTasklet::run(const ConversionParams &parameters, ConversionResult
             if (new_metadata_record.url_.empty())
                 throw std::runtime_error("no URL set");
 
-            if (ExcludeOnlineFirstRecord(new_metadata_record, parameters)) {
+            if (ExcludeUnsesiredItemTypes(new_metadata_record)) {
+                ++result->num_skipped_since_undesired_item_type_;
```
ruschein, line 1111: Please do not use blocks for one-liners.

cpp/lib/src/ZoteroHarvesterZederInterop.cc
```diff
@@ -92,14 +92,16 @@ static std::string ResolveHarvesterOperation(const Zeder::Entry &zeder_entry, co
 }
 
 
-static std::string ResolveUploadOperation(const Zeder::Entry &zeder_entry, const Zeder::Flavour /*unused*/) {
-    const auto &prodf(zeder_entry.getAttribute("prodf", ""));
-    if (prodf == "zotat")
+std::string ResolveUploadOperation(const Zeder::Entry &zeder_entry, const Zeder::Flavour /*unused*/) {
+    const auto &prode(zeder_entry.getAttribute("prode", ""));
```
ruschein, line 96: It seems like you don't need a variable here, also below regarding prodf.

cpp/zeder_to_zotero_importer.cc
```diff
@@ -62,7 +62,7 @@ using namespace ZoteroHarvester;
             "\tfields_to_update                Comma-separated list of the following fields to update: \n"
             "\t                                \tONLINE_PPN, PRINT_PPN, ONLINE_ISSN, PRINT_ISSN, EXPECTED_LANGUAGES, ENTRY_POINT_URL, UPLOAD_OPERATION, UPDATE_WINDOW, SSGN, LICENSE.\n"
             "\t                                Ignored when importing entries (all importable fields will be imported).\n"
-            "\t                                If mode is IMPORT and zeder_ids is '*', new journals will only be added if \"prodf\" is \"zota\" or \"zotat\".\n\n");
+            "\t                                If mode is IMPORT and zeder_ids is '*', new journals will only be added if UPLOAD_OPERATION would not be NONE.\n\n");
```
ruschein, line 65: "is not NONE" instead of the clunky sounding "would not be NONE", please.

cpp/zeder_to_zotero_importer.cc
```diff
@@ -334,9 +335,9 @@ unsigned ImportZederEntries(const Zeder::EntryCollection &zeder_entries, Harvest
             LOG_WARNING("couldn't import Zeder entry " + std::to_string(zeder_id) + " (" + title + "): already exists with different zeder id " + existing_title_section->getString("zeder_id"));
             continue;
         } else if (existing_journal_section == nullptr and autodetect_new_datasets) {
-            const std::string prodf(zeder_entry.getAttribute("prodf", ""));
-            if (prodf != "zota" && prodf != "zotat") {
-                LOG_DEBUG("Skipping Zeder entry " + std::to_string(zeder_id) + " (" + title + "): prodf is neither \"zota\" nor \"zotat\"");
+            if (upload_operation == Config::UPLOAD_OPERATION_TO_STRING_MAP.at(Config::UploadOperation::NONE)) {
```
ruschein, line 338: Why not fold this condition into the one above and save one more level of indentation?

cpp/zeder_to_zotero_importer.cc
```diff
@@ -363,14 +363,16 @@ unsigned ImportZederEntries(const Zeder::EntryCollection &zeder_entries, Harvest
             Config::JournalParams::LICENSE,
         };
 
-        // special-case the Zeder ID and modified timestamp fields
+        // special-cases: Zeder ID, modified timestamp, UPLOAD_OPERATION
```
ruschein, line 366: "to special-case" is a verb.

ruschein (changes requested):

cpp/lib/src/ZoteroHarvesterConversion.cc
mtrojan-ub: OK, but then i will also change this in other occurrences i see

cpp/lib/src/ZoteroHarvesterConversion.cc
mtrojan-ub: you're right, but the problem is that the `continue` statement is missing

PR #3640, opened 2020-10-27
zotaut: global ZoteroMetadataParams & MarcMetadataParams

mtrojan-ub:
implementation is now complete but untested

### DIFF
--- a/cpp/lib/src/ZoteroHarvesterConfig.cc
+++ b/cpp/lib/src/ZoteroHarvesterConfig.cc
@@ -132,6 +132,10 @@ GlobalParams::GlobalParams(const IniFile::Section &config_section) {
     rss_harvester_operation_params_.harvest_interval_ = config_section.getUnsigned(GetIniKeyString(RSS_HARVEST_INTERVAL));
     rss_harvester_operation_params_.force_process_feeds_with_no_pub_dates_ = config_section.getBool(GetIniKeyString(RSS_FORCE_PROCESS_FEEDS_WITH_NO_PUB_DATES));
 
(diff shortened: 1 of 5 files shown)


cpp/lib/src/ZoteroHarvesterConfig.cc
```diff
@@ -62,6 +62,55 @@ std::string GetHostTranslationServerUrl() {
...
+            fields_to_suppress_.insert(std::make_pair(field_name,
+                                       std::unique_ptr<ThreadSafeRegexMatcher>(new ThreadSafeRegexMatcher(entry.value_))));
+        } else if (StringUtil::StartsWith(entry.name_, PREFIX_EXCLUDE_JSON_FIELD)) {
+            const auto metadata_name(entry.name_.substr(__builtin_strlen(PREFIX_EXCLUDE_JSON_FIELD)));
+            exclusion_filters_.insert(std::make_pair(metadata_name,
+                                      std::unique_ptr<ThreadSafeRegexMatcher>(new ThreadSafeRegexMatcher(entry.value_))));
+        }
+    }
+}
+
```
ruschein, line 85: Q: Do we get an error message if we misspell one of the global parameters?

mtrojan-ub: No, but therefore it would be necessary to have a register of all valid parameter names for a section and that does neither exist globally nor section-specific.

mtrojan-ub: Well i guess my last statement wasn't true... i guess that there is a list which isn't complete due to the dynamic fields which are just an allowed prefix, so this could be achieved pretty easily, but i guess we could do this in a follow-up PR.

PR #3643, opened 2020-10-29
ZoteroHarvesterConfig: validation

mtrojan-ub:


cpp/lib/src/ZoteroHarvesterConfig.cc
```diff
@@ -143,49 +180,60 @@ GlobalParams::GlobalParams(const IniFile::Section &config_section) {
...
     name_ = group_section.getSectionName();
     user_agent_ = group_section.getString(GetIniKeyString(USER_AGENT));
     isil_ = group_section.getString(GetIniKeyString(ISIL));
     output_folder_ = group_section.getString(GetIniKeyString(OUTPUT_FOLDER));
     author_swb_lookup_url_ = group_section.getString(GetIniKeyString(AUTHOR_SWB_LOOKUP_URL));
     author_lobid_lookup_query_params_ = group_section.getString(GetIniKeyString(AUTHOR_LOBID_LOOKUP_QUERY_PARAMS), "");
-}
 
+    // Warnings for unknown entries
+    for (const auto &entry : group_section) {
```
ruschein, line 221: This looks exactly like some other new code further up.  Please encapsulate both in a new function.

mtrojan-ub: There are 3 validation mechanisms looking similar but they are not exactly the same (additional functions are called for validation and the error messages differ as well)

mtrojan-ub: i see what i can do

mtrojan-ub: I could only achieve that with a template function. How do you define a generic parameter for a map of any enum to a std::string?

ruschein: Yes, but it is a trivial template.

cpp/lib/include/ZoteroHarvesterConfig.h
```diff
@@ -59,8 +60,44 @@ struct MarcMetadataParams {
 public:
     MarcMetadataParams() = default;
     MarcMetadataParams(const IniFile::Section &config_section);
+    static bool IsValidIniEntry(const IniFile::Entry &entry);
 };
```
ruschein, line 65: Missing blank line.

cpp/lib/include/ZoteroHarvesterConfig.h
```diff
@@ -59,8 +60,44 @@ struct MarcMetadataParams {
...
+        bool valid(false);
+        for (const auto &allowed_value : allowed_values) {
+            if (entry.name_ == allowed_value.second) {
+                valid = true;
+                break;
+            }
+        }
+
+        if (not valid) {
+            for (unsigned i(0); i<callbacks.size(); ++i) {
```
ruschein, line 84: You can go back to the for-in loop.  Also, this version is missing spaces.

ruschein (changes requested):

PR #3647, opened 2020-11-03
Added explicit support for 936rv.

ruschein:


cpp/merge_print_and_online.cc
```diff
@@ -837,8 +837,12 @@ void MergeRecordPair(MARC::Record * const merge_record, MARC::Record * const imp
         if (GetFuzzyIdenticalField(*merge_record, import_field, &merge_field_pos, compare_indicators, compare_subfields))
             continue;
 
-        // Non-existing fields in the merge_record can be inserted unconditionally
-        if (merge_record->findTag(import_field.getTag()) == merge_record->end()) {
+        // Non-existing or 936rv fields in the merge_record can be inserted unconditionally.
```
jriedl, line 840: Isn't the "Non-existing fields in the merge_record can be inserted unconditionally" rule unconditionally valid and this logic belongs to the MergeFieldPair936-function below?

cpp/merge_print_and_online.cc
```diff
@@ -837,8 +837,12 @@ void MergeRecordPair(MARC::Record * const merge_record, MARC::Record * const imp
...
+        // (936rv are keyword fields.)
+        const auto &import_tag(import_field.getTag());
+        if ((import_tag.toString() == "936" and import_field.getIndicator1() == 'r' and import_field.getIndicator2() == 'v')
```
jriedl, line 843: From the Krimdok title data and https://wiki.k10plus.de/display/K10PLUS/Exportformate?preview=/27361336/267288632/Pica_Marc_Titeldaten_K10plus_20200727_Wiki.xlsx indicators "bk" must also be handled seperatately. Indicators "uw" seem to be our standard 936 fields.

PR #3653, opened 2020-11-06
Use a different downloader.

ruschein:


cronjobs/util.py
```diff
@@ -52,7 +56,9 @@ def RetrieveFileByURL(url: str, timeout: int, accepted_content_types: List[str]
                 content_type: str = headers["Content-type"].lower()
                 for accepted_content_type in accepted_content_types:
                     if accepted_content_type.lower() == content_type:
+                        print(str(headers))
```
mtrojan-ub, line 59: Is this a debug statement?

ruschein: It was and I removed it.

PR #3659, opened 2020-11-11
zotaut: Show error messages in web interface

mtrojan-ub:


cpp/data/zotero_cgi/delivered.html
```diff
@@ -27,7 +27,7 @@ <h2>Delivered records for Zeder ID {zeder_id}</h2>
                     <td align="center" style="vertical-align: middle;">
                         {IF delivery_states == "automatic"}<span class="dot" style="background-color: green;" title="This record has been delivered automatically."></span>{ENDIF}
                         {IF delivery_states == "manual"}<span class="dot" style="background-color: turquoise;" title="This record has been delivered manually and will no longer be harvested automatically."></span>{ENDIF}
-                        {IF delivery_states == "error"}<span class="dot" style="background-color: yellow;" title="An error has occurred when trying to deliver this record."></span>{ENDIF}
+                        {IF delivery_states == "error"}<span class="dot" style="background-color: yellow;" title="An error has occurred when trying to deliver this record:&#013;{error_messages}"></span>{ENDIF}
```
ruschein, line 30: I think "while" instead of "when" sounds more natural.

cpp/lib/include/DbConnection.h
```diff
@@ -153,10 +153,13 @@ class DbConnection {
     /** \note Converts the binary contents of "unescaped_string" into a form that can used as a string.
      *  \note This probably breaks for Sqlite if the string contains binary characters.
      */
-    std::string escapeString(const std::string &unescaped_string, const bool add_quotes = false);
+    std::string escapeString(const std::string &unescaped_string, const bool add_quotes = false, const bool return_null_on_empty_string=false);
```
ruschein, line 156: Missing spaces.  Also it might be nice to document this parameter.  I guess it if it is true and we would have o/w returned an empty string we return "NULL" instead?

cpp/lib/include/DbConnection.h
```diff
@@ -153,10 +153,13 @@ class DbConnection {
...
     inline std::string escapeAndQuoteString(const std::string &unescaped_string) {
         return escapeString(unescaped_string, /* add_quotes = */true);
     }
+    inline std::string escapeAndQuoteStringOrNull(const std::string &unescaped_string) {
```
ruschein, line 160: I'd add a blank line above this.  We only omit those between one-liners w/o comments.

ruschein, line 160: I don't think this name is particularly good. I know it's a monstrosity, but how about escapeAndQuoteNonEmptyStringOrReturnNull?

cpp/lib/include/ZoteroHarvesterUtil.h
```diff
@@ -578,7 +579,7 @@ class UploadTracker {
 
     // Saves the record blob and its associated metadata in the host's database.
     // Returns true on success, false otherwise.
-    bool archiveRecord(const MARC::Record &record, const DeliveryState delivery_state);
+    bool archiveRecord(const MARC::Record &record, const DeliveryState delivery_state, const std::string &error_message="");
```
ruschein, line 582: Missing spaces.  Maybe what gets you is that in Python we tend to omit the spaces around the equal signs for default value assignments in function parameters?  Or maybe in PHP?

ruschein (changes requested):

PR #3661, opened 2020-11-12
Renamed column.

ruschein:


cpp/data/installer/ub_tools.sql
```diff
@@ -40,10 +40,10 @@ CREATE TABLE zeder_journals (
 -- Table to be used w/ our validate_harvested_records tool:
 CREATE TABLE metadata_presence_tracer (
     zeder_journal_id INT(11) UNSIGNED,
-    metadata_field_name CHAR(4) NOT NULL,
+    marc_field_tag CHAR(4) NOT NULL,
```
mtrojan-ub, line 43: 3

ruschein: Fixed

cpp/data/sql_updates/ub_tools.22
```diff
@@ -0,0 +1 @@
+ALTER TABLE metadata_presence_tracer RENAME COLUMN metadata_field_name TO marc_field_tag
```
mtrojan-ub, line 1: length should be changed to 3 here as well

ruschein: Done!

mtrojan-ub: Multiple Problems:
- Please use MODIFY COLUMN syntax so it works on Ubuntu + CentOS
- Don't forget NOT NULL in ub_tools.22
- Why did you add record_type in this pull request? (it is used nowhere)
- If you add a new column, please also use the AFTER keyword to add it at the right place (else diffs will get more and more confusing)

mtrojan-ub: And please use semicolon after each statement!

ruschein: Done.

PR #3662, opened 2020-11-13
Use syntax that also works with MariaDB.

ruschein:


mtrojan-ub:
i will fix this myself...

PR #3667, opened 2020-11-17
Fix problem w/ strange ZWI fields in record 828842965 on ptah etc.

ruschein:


cpp/merge_print_and_online.cc
```diff
@@ -493,7 +493,8 @@ void UpdateMergedPPNs(MARC::Record * const record, const std::set<std::string> &
     if (not max_publication_year.empty())
         zwi_subfields.addSubfield('y', max_publication_year);
 
-    record->replaceField("ZWI", zwi_subfields);
+    record->deleteFields("ZWI");
+    record->appendField("ZWI", zwi_subfields);
```
mtrojan-ub, line 497: Do we lose information this way? (e.g. can it happen that the existing ZWI contains information about formerly merged PPNs that are not included in the given merged_ppns array?)

cpp/lib/src/MARC.cc
```diff
@@ -1643,6 +1643,21 @@ std::unordered_set<std::string> Record::getTagSet() const {
...
+    std::vector<size_t> field_indices;
+    for (auto field(fields_.cbegin()); field != fields_.cend(); ++field) {
+        if (field->getTag() == field_tag)
+            field_indices.emplace_back(field - fields_.cbegin());
+    }
+
+    if (field_indices.empty())
+        return 0;
+
+    fields_.erase(fields_.begin() + field_indices.front(), fields_.begin() + field_indices.back());
```
mtrojan-ub, line 1656: why is begin() used here if cbegin() is used on line 1650?

cpp/lib/src/MARC.cc
```diff
@@ -1643,6 +1643,21 @@ std::unordered_set<std::string> Record::getTagSet() const {
 }
 
 
+size_t Record::deleteFields(const Tag &field_tag) {
+    std::vector<size_t> field_indices;
+    for (auto field(fields_.cbegin()); field != fields_.cend(); ++field) {
+        if (field->getTag() == field_tag)
+            field_indices.emplace_back(field - fields_.cbegin());
```
mtrojan-ub, line 1650: isn't fiends_.cbegin() == 0?

cpp/lib/src/MARC.cc
mtrojan-ub: and also: isn't begin() == 0?

PR #3678, opened 2020-11-24
Ruschein2

ruschein:


cpp/data/installer/ub_tools.sql
```diff
@@ -39,8 +39,9 @@ CREATE TABLE zeder_journals (
 
 -- Table to be used w/ our validate_harvested_records tool:
 CREATE TABLE metadata_presence_tracer (
-    zeder_journal_id INT(11) UNSIGNED,
+    journal_id INT(11) UNSIGNED,
```
mtrojan-ub, line 42: careful about renaming this column. There are about 55 other potential references to this column in the whole code that should be renamed as well, also you forgot the reference in the CONSTRAINT in this table, and there also needs to be an sql update.

cpp/data/installer/ub_tools.sql
```diff
@@ -39,8 +39,9 @@ CREATE TABLE zeder_journals (
...
     marc_field_tag CHAR(3) NOT NULL,
+    marc_subfield_code CHAR(1) NOT NULL,
```
mtrojan-ub, line 44: any DEFAULT? (what happens to old values? if we throw all of them away, there needs to be a DELETE statement in the sql_update before adding the column)

cpp/data/sql_updates/ub_tools.23
```diff
@@ -0,0 +1 @@
+ALTER TABLE metadata_presence_tracer ADD COLUMN marc_subfield_code CHAR(1) AFTER marc_field_tag;
```
mtrojan-ub, line 1: should be the same as ub_tools.sql (e.g. NOT NULL missing)

cpp/validate_harvested_records.cc
```diff
@@ -39,274 +40,195 @@
...
 
-    return false;
+    LOG_ERROR("unknown field presence \"" + field_presence_str + "\"!");
 }
 
 
-struct FieldInfo {
-    std::string name_;
-    FieldPresence presence_;
+class FieldRule {
```
mtrojan-ub, line 58: This looks like a container to me that may contain multiple rules, so it might be better to call the class `FieldRules` instead of `FieldRule`

cpp/validate_harvested_records.cc
```diff
@@ -39,274 +40,195 @@
...
-                ++lhs_iter;
-            } else if (compare_result > 0) {
-                combined_info.addField(*rhs_iter);
-                ++rhs_iter;
-            }
+    for (const auto &subfield_code_and_field_presence : subfield_code_to_field_presence_map_) {
+        const auto iter(found_subfield_codes.find(subfield_code_and_field_presence.first));
+        if (iter == found_subfield_codes.end() and subfield_code_and_field_presence.second == ALWAYS) {
+            if (not reason_for_being_invalid->empty())
+                reason_for_being_invalid->append("; ");
```
mtrojan-ub, line 88: Personal taste: i would rather use a `std::vector<std::string>` for `reason_for_being_invalid` and leave the way things are formatted for output to the caller (e.g. to implode it with `;`, `\n` or anything else)

cpp/validate_harvested_records.cc
```diff
@@ -39,274 +40,195 @@
...
-    const std::string &getZederId() const { return zeder_id_; }
-    const std::string &getZederInstance() const { return zeder_instance_; }
-    bool isInDatabase() const { return not not_in_database_yet_; }
+    ~Validator() {}
+
+    /** \return True if we found rules for all subfields in "field" o/w false.
+     *  \note   If a rule violation was found, "reason_for_being_invalid" w/ be non-empty after the call and
+     *          we will return true.
+     */
+    virtual bool foundRuleMatch(const std::string &journal_id, const MARC::Record::Field &field,
```
mtrojan-ub, line 104: Maybe we can find a better name for the function. The name suggests that this function only checks for rules that match against the field, but it's more about validating a Field against these rules, so i guess `validateField` would be better

mtrojan-ub: Regarding the return value, i would also suggest that a `validate` function might be clearer. A field should be treated as valid (true) if for each subfield, either the rule is fulfilled or there is no rule. Only if at least 1 existing rule is violated, the validation should fail.

cpp/validate_harvested_records.cc
```diff
@@ -322,126 +244,122 @@ void SendEmail(const std::string &email_address, const std::string &message_subj
...
-                              + "AND zeder_instance=" + db_connection->escapeAndQuoteString(zeder_instance) + ") "
-                              + "AND field_name='" + field_name + "'");
-    if (db_connection->getNoOfAffectedRows() == 0)
-        LOG_ERROR("can't update non-existent database entry: " + zeder_id + "(" + zeder_instance + ")"
-                  + ", field_name: \"" + field_name + "\"");
+void CheckGenericRequirements(const MARC::Record &record, std::string * const reasons_for_being_invalid) {
+    if (not record.hasTag("001"))
+        reasons_for_being_invalid->append("control field 001 is missing\n");
+
+    if (not record.hasTag("008"))
```
mtrojan-ub, line 251: We do not generate 008. Instead, 003 and 007 should be checked (in addition to 001).

mtrojan-ub:
zotero_cgi changes missing

cpp/validate_harvested_records.cc
```diff
@@ -322,126 +244,122 @@ void SendEmail(const std::string &email_address, const std::string &message_subj
...
+        reasons_for_being_invalid->append("control field 008 is missing\n");
+
+    const auto _245_field(record.findTag("245"));
+    if (_245_field == record.end())
+        reasons_for_being_invalid->append("field 245 is missing\n");
+    else if (_245_field->getFirstSubfieldWithCode('a').empty())
+        reasons_for_being_invalid->append("subfield 245$a is missing\n");
+
+    if (not record.hasTag("100") and not record.hasTag("700"))
```
mtrojan-ub, line 260: I am not sure if it is really mandatory to have at least 1 author

cpp/validate_harvested_records.cc
```diff
@@ -322,126 +244,122 @@ void SendEmail(const std::string &email_address, const std::string &message_subj
...
-        LOG_ERROR("can't update non-existent database entry: " + zeder_id + "(" + zeder_instance + ")"
-                  + ", field_name: \"" + field_name + "\"");
+void CheckGenericRequirements(const MARC::Record &record, std::string * const reasons_for_being_invalid) {
+    if (not record.hasTag("001"))
+        reasons_for_being_invalid->append("control field 001 is missing\n");
+
+    if (not record.hasTag("008"))
+        reasons_for_being_invalid->append("control field 008 is missing\n");
+
+    const auto _245_field(record.findTag("245"));
```
mtrojan-ub, line 254: 245a could be checked via a "normal" rule from the DB which is visible in the frontend to keep things transparent to the use

cpp/validate_harvested_records.cc
```diff
@@ -322,126 +244,122 @@ void SendEmail(const std::string &email_address, const std::string &message_subj
...
 
 
 } // unnamed namespace
 
 
 int Main(int argc, char *argv[]) {
-    if (argc != 5 and argc != 6)
-        Usage();
+    if (argc != 5)
+        ::Usage("marc_input marc_output missed_expectations_file email_address");
```
mtrojan-ub, line 318: It's OK if you remove the update_db mode, but what happened to the [--update-db-errors] flag for the regular mode?

cpp/validate_harvested_records.cc
```diff
@@ -322,126 +244,122 @@ void SendEmail(const std::string &email_address, const std::string &message_subj
...
-    unsigned total_record_count(0), new_record_count(0), missed_expectation_count(0);
-    while (const auto record = reader->read()) {
+    GeneralValidator general_regular_article_validator, general_review_article_validator;
+    JournalSpecificValidator journal_specific_regular_article_validator, journal_specific_review_article_validator;
+    LoadRules(&db_connection, &general_regular_article_validator, &journal_specific_regular_article_validator,
+              &general_review_article_validator, &journal_specific_review_article_validator);
+    std::vector<const Validator *> regular_article_validators{ &journal_specific_regular_article_validator,
+                                                               &general_regular_article_validator },
+                                   review_article_validators{ &journal_specific_review_article_validator,
+                                                              &journal_specific_regular_article_validator,
```
mtrojan-ub, line 335: Shouldn't `general_review_article_validator` have a higher priority than `journal_specific_regular_article_validator`?

cpp/validate_harvested_records.cc
ruschein: According to my conversation w/ C.W. last week that is a hard requirement.

cpp/validate_harvested_records.cc
```diff
@@ -39,274 +40,195 @@
...
-
-    return combined_info;
 }
 
 
-class JournalInfo : public GeneralInfo {
-    std::string zeder_id_;
-    std::string zeder_instance_;
-    bool not_in_database_yet_;
+class Validator {
```
mtrojan-ub, line 96: `FieldValidator` might suit better

cpp/validate_harvested_records.cc
```diff
@@ -39,274 +40,195 @@
...
+               JournalSpecificValidator * const journal_specific_regular_article_validator,
+               GeneralValidator * const general_review_article_validator,
+               JournalSpecificValidator * const journal_specific_review_article_validator)
+{
+    db_connection->queryOrDie(
+        "SELECT marc_field_tag,marc_subfield_code,field_presence,record_type FROM metadata_presence_tracer");
+    DbResultSet result_set(db_connection->getLastResultSet());
+    while (const auto row = result_set.getNextRow()) {
+        if (row["record_type"] == "regular_article") {
+            if (row.isNull("journal_id"))
```
mtrojan-ub, line 216: the `journal_id` column is not included in your sql statement

mtrojan-ub: ... Also it might be useful to add `ORDER BY marc_field_tag, marc_subfield_code ASC` so that error messages will be in the correct order later

cpp/validate_harvested_records.cc
ruschein: True.  We can discuss this when we will discuss the rule-inheritance hierarchy.

cpp/validate_harvested_records.cc
ruschein: Done

cpp/data/installer/ub_tools.sql
mtrojan-ub: ... and if i say 55 other potential references, i also mean the rest of ub_tools which is not included in this pull request so far (e.g. ZoteroHarvesterUtil and others)

cpp/data/sql_updates/ub_tools.24
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE metadata_presence_tracer ADD COLUMN marc_subfield_code CHAR(1) NOT NULL AFTER marc_field_tag;
+ALTER TABLE metadata_presence_tracer CHANGE COLUMN zeder_journal_id journal_id INT(11) UNSIGNED;
```
mtrojan-ub, line 2: What about the change to the metadata_presence_tracer_zeder_journal_id CONSTRAINT? Will this be changed implicitly as well?

cpp/validate_harvested_records.cc
mtrojan-ub: Then maybe we should put a documentation into the zotero_cgi webinterface (where the rules are listed) and tell the user which additional hardcoded checks are performed and where there is no need to add explicit rules e.g. for 100 and 700.

cpp/validate_harvested_records.cc
```diff
@@ -322,126 +245,123 @@ void SendEmail(const std::string &email_address, const std::string &message_subj
...
-        JournalInfo new_journal_info;
-        LoadFromDatabaseOrCreateFromScratch(db_connection, zeder_id, zeder_instance, &new_journal_info);
-        (*journal_name_to_info_map)[journal_name] = new_journal_info;
-        journal_name_and_info = journal_name_to_info_map->find(journal_name);
+    reasons_for_being_invalid->clear();
+
+    const auto zid_field(record.findTag("ZID"));
+    if (unlikely(zid_field == record.end()))
+        LOG_ERROR("record is missing a ZID field!");
+    else if (zid_field->getFirstSubfieldWithCode('a').empty())
```
mtrojan-ub, line 276: Is it better to use a variable for the subfield lookup and re-using it instead of executing the lookup twice?

cpp/validate_harvested_records.cc
mtrojan-ub: Also we should put a comment somewhere here that if somebody makes any changes here the frontend documentation should be changed as well

cpp/validate_harvested_records.cc
ruschein: Done

cpp/validate_harvested_records.cc
mtrojan-ub: ah, now i understand why it's needed this way without validation, sorry for bothering.

cpp/validate_harvested_records.cc
```diff
@@ -322,58 +250,70 @@ void SendEmail(const std::string &email_address, const std::string &message_subj
...
+    for (const auto &field : record) {
+        const auto current_tag(field.getTag());
+        if (current_tag == last_tag and not field.isRepeatableField())
+            reasons_for_being_invalid->append(current_tag.toString() + " is not a repeatable field\n");
+        last_tag = current_tag;
+        present_tags.emplace(current_tag.toString());
+
+        for (const auto field_validator : field_validators) {
+            std::string reason_for_being_invalid;
+            if (field_validator->foundRuleMatch(ZID, field, &reason_for_being_invalid)) {
```
mtrojan-ub, line 302: I guess this should not only be performed for every field, but for every single subfield.

cpp/validate_harvested_records.cc
ruschein: Done

cpp/validate_harvested_records.cc
ruschein: Done

cpp/validate_harvested_records.cc
ruschein: No! (Please talk to me about why.)

cpp/data/sql_updates/ub_tools.24
ruschein: I would hope so but am not 100% certain.

cpp/data/installer/ub_tools.sql
ruschein: Done

cpp/data/installer/ub_tools.sql
ruschein: DONE

cpp/data/sql_updates/ub_tools.23
ruschein: Done

cpp/validate_harvested_records.cc
ruschein: Done

cpp/data/sql_updates/ub_tools.24
ruschein: Taken care of.

cpp/validate_harvested_records.cc
ruschein: Done

cpp/validate_harvested_records.cc
mtrojan-ub: As discussed, if we want to keep the behaviour like this, we should definitely speak to @socheres about it before merging.

cpp/data/installer/ub_tools.sql
mtrojan-ub: OK, this should also be discussed with @socheres when merging

cpp/validate_harvested_records.cc
```diff
@@ -322,58 +251,70 @@ void SendEmail(const std::string &email_address, const std::string &message_subj
...
-
-    db_connection->queryOrDie("UPDATE metadata_presence_tracer SET field_presence='" + field_presence_str + "' WHERE zeder_journal_id="
-                              + "(SELECT id FROM zeder_journals WHERE zeder_id=" + db_connection->escapeAndQuoteString(zeder_id) + " "
-                              + "AND zeder_instance=" + db_connection->escapeAndQuoteString(zeder_instance) + ") "
-                              + "AND field_name='" + field_name + "'");
-    if (db_connection->getNoOfAffectedRows() == 0)
-        LOG_ERROR("can't update non-existent database entry: " + zeder_id + "(" + zeder_instance + ")"
-                  + ", field_name: \"" + field_name + "\"");
+void CheckGenericRequirements(const MARC::Record &record, std::string * const reasons_for_being_invalid) {
+    if (not record.hasTag("001"))
```
mtrojan-ub, line 255: ... might be more elegant to define a std::set<std::string> of the mandatory tags and iterate over it rather than copy/pasting code

cpp/validate_harvested_records.cc
mtrojan-ub: You added `ORDER BY ...`, but `journal_id` is still missing

cpp/data/sql_updates/ub_tools.24
```diff
@@ -0,0 +1,7 @@
+DELETE * FROM metadata_presence_tracer;
+ALTER TABLE metadata_presence_tracer ADD COLUMN marc_subfield_code CHAR(1) NOT NULL AFTER marc_field_tag;
+ALTER TABLE metadata_presence_tracer CHANGE COLUMN zeder_journal_id journal_id INT(11) UNSIGNED;
+ALTER TABLE metadata_presence_tracer DROP CONSTRAINT metadata_presence_tracer_zeder_journal_id;
```
mtrojan-ub, line 4: The `DROP CONSTRAINT` statements must be executed before the `CHANGE COLUMN` statement

cpp/data/sql_updates/ub_tools.24
```diff
@@ -0,0 +1,7 @@
+DELETE * FROM metadata_presence_tracer;
```
mtrojan-ub, line 1: I guess you should add `WHERE 1` add the end, because some MySQL installations my refuse to execute a `DELETE` statement without `WHERE` (especially under CentOS).

cpp/data/installer/ub_tools.sql
```diff
@@ -39,11 +39,12 @@ CREATE TABLE zeder_journals (
...
     record_type ENUM('regular_article', 'review') DEFAULT 'regular_article' NOT NULL,
     field_presence ENUM('always', 'sometimes', 'ignore') NOT NULL,
-    CONSTRAINT metadata_presence_tracer_zeder_journal_id FOREIGN KEY (zeder_journal_id) REFERENCES zeder_journals (id) ON DELETE CASCADE ON UPDATE CASCADE,
+    CONSTRAINT metadata_presence_tracer_zeder_journal_id FOREIGN KEY (journal_id) REFERENCES zeder_journals (id) ON DELETE CASCADE ON UPDATE CASCADE,
```
mtrojan-ub, line 47: The name of this constraint should be changed as well (also in ub_tools.24)

mtrojan-ub: Also, the second constraint still contains a reference to zeder_journal_id

cpp/data/installer/ub_tools.sql
mtrojan-ub: i hope you're right

cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -578,7 +578,7 @@ bool ProcessShowQASubActionAdd(const std::multimap<std::string, std::string> &cg
     if (add_type == "global")
         journal_id_to_insert = "NULL";
 
-    db_connection->queryOrDie("INSERT INTO metadata_presence_tracer (zeder_journal_id, marc_field_tag, field_presence) "
+    db_connection->queryOrDie("INSERT INTO metadata_presence_tracer (journal_id, marc_field_tag, field_presence) "
```
mtrojan-ub, line 581: This won't work anymore with the changes to ub_tools.sql (missing mandatory subfield code in insert statement).

Multiple things need to be changed in zotero_cgi.cc to, and also qa.html (which is missing from the pull request so far).

cpp/validate_harvested_records.cc
ruschein: Should be fixed now.

cpp/validate_harvested_records.cc
ruschein: I don't think so.

cpp/data/installer/ub_tools.sql
ruschein: Fixed

cpp/cgi_progs/zotero_cgi.cc
ruschein: Fixed

cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -578,9 +579,9 @@ bool ProcessShowQASubActionAdd(const std::multimap<std::string, std::string> &cg
     if (add_type == "global")
         journal_id_to_insert = "NULL";
 
-    db_connection->queryOrDie("INSERT INTO metadata_presence_tracer (zeder_journal_id, marc_field_tag, field_presence) "
-                              " VALUES (" + journal_id_to_insert + ", " + db_connection->escapeAndQuoteString(add_tag) + ", "
-                              " " + db_connection->escapeAndQuoteString(add_presence) + ")");
+    db_connection->queryOrDie("INSERT INTO metadata_presence_tracer (journal_id, marc_field_tag, marc_subfield_code, field_presence) "
+                              " VALUES (" + journal_id_to_insert + ", " + db_connection->escapeAndQuoteString(add_tag) + ", '"
```
mtrojan-ub, line 583: escapeAndQuoteString could be used here as well (consistency)

cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -590,16 +591,18 @@ bool ProcessShowQASubActionDelete(const std::multimap<std::string, std::string>
...
     if (delete_type == "global")
         journal_id_to_delete = "IS NULL";
 
-    db_connection->queryOrDie("DELETE FROM metadata_presence_tracer "
-                             "WHERE zeder_journal_id " + journal_id_to_delete + " "
-                             "AND marc_field_tag = " + db_connection->escapeAndQuoteString(delete_tag));
+    db_connection->queryOrDie("DELETE FROM metadata_presence_tracer"
+                              " WHERE journal_id " + journal_id_to_delete +
+                              " AND marc_field_tag = " + db_connection->escapeAndQuoteString(delete_tag) +
+                              " AND subfield_code = " + db_connection->escapeAndQuoteString(delete_subfield_code));
```
mtrojan-ub, line 605: `subfield_code` => `marc_subfield_code`

cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -628,34 +631,36 @@ void ProcessShowQAAction(const std::multimap<std::string, std::string> &cgi_args
             submitted = "true";
 
     // display current settings
-    db_connection->queryOrDie("SELECT * FROM metadata_presence_tracer WHERE zeder_journal_id IS NULL "
-                             "OR zeder_journal_id IN (SELECT id FROM zeder_journals WHERE zeder_id=" + db_connection->escapeAndQuoteString(zeder_id) +
+    db_connection->queryOrDie("SELECT * FROM metadata_presence_tracer WHERE journal_id IS NULL "
+                             "OR journal_id IN (SELECT id FROM zeder_journals WHERE zeder_id=" + db_connection->escapeAndQuoteString(zeder_id) +
                              " AND zeder_instance=" + db_connection->escapeAndQuoteString(zeder_instance) + ")"
-                             " ORDER BY marc_field_tag ASC, zeder_journal_id ASC");
+                             " ORDER BY marc_field_tag ASC, journal_id ASC");
```
mtrojan-ub, line 637: also order by `marc_subfield_code ASC`

cpp/data/sql_updates/ub_tools.24
mtrojan-ub: ... the ADD COLUMN statement could have been stayed at the top because it's semantically not depending on the others, but well... OK anyway

cpp/data/installer/ub_tools.sql
```diff
@@ -39,12 +39,13 @@ CREATE TABLE zeder_journals (
...
-    zeder_journal_id INT(11) UNSIGNED,
+    journal_id INT(11) UNSIGNED,
     marc_field_tag CHAR(3) NOT NULL,
+    marc_subfield_code CHAR(1) NOT NULL,
     record_type ENUM('regular_article', 'review') DEFAULT 'regular_article' NOT NULL,
     field_presence ENUM('always', 'sometimes', 'ignore') NOT NULL,
-    CONSTRAINT metadata_presence_tracer_zeder_journal_id FOREIGN KEY (zeder_journal_id) REFERENCES zeder_journals (id) ON DELETE CASCADE ON UPDATE CASCADE,
-    CONSTRAINT metadata_presence_tracer_journal_id_and_marc_field_tag UNIQUE (zeder_journal_id, marc_field_tag)
+    CONSTRAINT metadata_presence_tracer_journal_id FOREIGN KEY (journal_id) REFERENCES zeder_journals (id) ON DELETE CASCADE ON UPDATE CASCADE,
+    CONSTRAINT metadata_presence_tracer_journal_id_and_marc_field_tag UNIQUE (journal_id, marc_field_tag)
```
mtrojan-ub, line 48: I just noticed that we need to change this unique constraint anyway by adding the subfield_code to it, else we will have a major problem! (please also change the name)

cpp/data/sql_updates/ub_tools.24
```diff
@@ -0,0 +1,10 @@
+-- We rename metadata_presence_tracer.zeder_journal_id to journal_id
+-- and add the new column metadata_presence_tracer.marc_subfield_code.
+-- Constraints affected by the rename are also being updated.
+TRUNCATE TABLE metadata_presence_tracer;
+ALTER TABLE metadata_presence_tracer DROP CONSTRAINT metadata_presence_tracer_zeder_journal_id;
+ALTER TABLE metadata_presence_tracer DROP CONSTRAINT metadata_presence_tracer_journal_id_and_marc_field_tag;
+ALTER TABLE metadata_presence_tracer ADD COLUMN marc_subfield_code CHAR(1) NOT NULL AFTER marc_field_tag;
+ALTER TABLE metadata_presence_tracer CHANGE COLUMN zeder_journal_id journal_id INT(11) UNSIGNED;
+ALTER TABLE metadata_presence_tracer ADD CONSTRAINT metadata_presence_tracer_journal_id FOREIGN KEY (journal_id) REFERENCES zeder_journals (id) ON DELETE CASCADE ON UPDATE CASCADE;
+ALTER TABLE metadata_presence_tracer ADD CONSTRAINT metadata_presence_tracer_journal_id_and_marc_field_tag UNIQUE (journal_id, marc_field_tag);
```
mtrojan-ub, line 10: Also add `marc_subfield_code` to the UNIQUE constraint here & change the name

cpp/data/zotero_cgi/qa.html
```diff
@@ -52,6 +52,7 @@ <h1>QA Settings</h1>
                 <p>Add new journal setting:</p>
                 <!-- The quantifier inside the pattern needs to be written as "{{3}" due to Template.h escaping rules! -->
                 <input name="add_tag" placeholder="3-digit tag number" required="required" minlength="3" maxlength="3" pattern="[0-9]{{3}">
+                <input name="add_subfield_code" placeholder="3-character subfield code" required="required" minlength="1" maxlength="1" pattern="[0-9a-z]{{1}">
```
mtrojan-ub, line 55: copy/paste error in placeholder attribute: should be `1`-character subfield code

cpp/data/sql_updates/ub_tools.24
ruschein: Done.  Please note that constraint names can be at most 64 characters long.

cpp/mysql_database_patcher.cc
```diff
@@ -92,7 +92,7 @@ void ApplyUpdate(DbConnection * const db_connection, const std::string &update_d
     SplitIntoDatabaseAndVersion(update_filename, &database, &update_version);
 
     if (not db_connection->mySQLDatabaseExists(database)) {
-        LOG_INFO("database \"" + database + "\" does not exist, skipping file " + update_filename);
+        LOG_WARNING("database \"" + database + "\" does not exist, skipping file " + update_filename);
```
mtrojan-ub, line 95: This shouldn't be a warning... there are many servers where this is totally normal, e.g. if no VuFind or IxTheo database exists on servers like nu, benu or ub28

cpp/data/zotero_cgi/qa.html
```diff
@@ -33,11 +33,11 @@ <h1>QA Settings</h1>
                 </tr>
             </thead>
             <tbody>
-                {LOOP tags, global_settings, journal_settings}
+                {LOOP tags, subfield_codes, global_settings, journal_settings}
```
mtrojan-ub, line 36: There should be a help text somewhere in this file that tells the user which checks are hardcoded (e.g. 001, 003, 007)

cpp/mysql_database_patcher.cc
ruschein: Reverted.

cpp/data/sql_updates/ub_tools.24
```diff
@@ -0,0 +1,10 @@
+-- We rename metadata_presence_tracer.zeder_journal_id to journal_id
+-- and add the new column metadata_presence_tracer.marc_subfield_code.
+-- Constraints affected by the rename are also being updated.
+TRUNCATE TABLE metadata_presence_tracer;
+ALTER TABLE metadata_presence_tracer DROP CONSTRAINT metadata_presence_tracer_zeder_journal_id;
```
mtrojan-ub, line 5: Just tested this... DROP CONSTRAINT only works with MariaDB but not with MySQL 😭

cpp/validate_harvested_records.cc
ruschein: Done

cpp/cgi_progs/zotero_cgi.cc
ruschein: I don't think that's any more readable.

cpp/cgi_progs/zotero_cgi.cc
ruschein: Done

cpp/cgi_progs/zotero_cgi.cc
ruschein: Done

cpp/data/zotero_cgi/qa.html
mtrojan-ub, line 36: Also, the difference of regular_article settings to review_article settings must be added to this view!

cpp/data/installer/ub_tools.sql
```diff
@@ -39,12 +39,13 @@ CREATE TABLE zeder_journals (
...
-    zeder_journal_id INT(11) UNSIGNED,
+    journal_id INT(11) UNSIGNED,
     marc_field_tag CHAR(3) NOT NULL,
+    marc_subfield_code CHAR(1) NOT NULL,
     record_type ENUM('regular_article', 'review') DEFAULT 'regular_article' NOT NULL,
     field_presence ENUM('always', 'sometimes', 'ignore') NOT NULL,
-    CONSTRAINT metadata_presence_tracer_zeder_journal_id FOREIGN KEY (zeder_journal_id) REFERENCES zeder_journals (id) ON DELETE CASCADE ON UPDATE CASCADE,
-    CONSTRAINT metadata_presence_tracer_journal_id_and_marc_field_tag UNIQUE (zeder_journal_id, marc_field_tag)
+    CONSTRAINT metadata_presence_tracer_journal_id FOREIGN KEY (journal_id) REFERENCES zeder_journals (id) ON DELETE CASCADE ON UPDATE CASCADE,
+    CONSTRAINT metadata_presence_tracer_journal_id_and_marc_field_tag UNIQUE (journal_id, marc_field_tag, marc_subfield_code)
```
mtrojan-ub, line 48: The name of this constraint doesn't fully represent the contents anymore

cpp/data/zotero_cgi/qa.html
mtrojan-ub: problem still exists

cpp/data/installer/ub_tools.sql
ruschein: See newest change.

cpp/email_watcher.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+fetchmail
+/usr/local/bin/email_watcher /usr/local/var/lib/tuelib/email_watcher.conf /home/qubjr01/mbox
```
mtrojan-ub, line 3: i guess this should not run with the qubjr01 user/directory

cpp/data/installer/ub_tools.sql
```diff
@@ -39,12 +39,14 @@ CREATE TABLE zeder_journals (
 
 -- Table to be used w/ our validate_harvested_records tool:
 CREATE TABLE metadata_presence_tracer (
-    zeder_journal_id INT(11) UNSIGNED,
+    journal_id INT(11) UNSIGNED,
     marc_field_tag CHAR(3) NOT NULL,
+    marc_subfield_code CHAR(1) NOT NULL,
+    optional_regex VARCHAR(200),
```
mtrojan-ub, line 45: - I don't remember any other field that contains the prefix "optional_" in its name, just because it is not mandatory. Is this really necessary?
- How about a DEFAULT value for an optional field?

cpp/merge_print_and_online.cc
```diff
@@ -272,13 +272,12 @@ void EliminateDanglingOrUnreferencedCrossLinks(const bool debug, const std::unor
 }
 
 
-const std::set<MARC::Tag> LINK_TAGS{ "800", "810", "830", "773", "775", "776" };
+const std::set<MARC::Tag> LINK_TAGS{ "800", "810", "830", "773", "830" };
```
mtrojan-ub, line 275: Duplicate entry "830". Also, is it corect that 775 and 776 have been removed?

ruschein: Done

cpp/data/zotero_cgi/qa.html
```diff
@@ -33,17 +44,17 @@ <h1>QA Settings</h1>
...
                         <td>{global_settings}</td>
-                        {IF journal_name != ""}<td>{journal_settings}{IF journal_settings != ""}<a href="?action=show_qa&zeder_id={zeder_id}&zeder_instance={zeder_instance}&delete_tag={tags}&delete_type=local" title="Delete this rule" onclick="return confirm('Are you sure?')"><sup>x</sup></a>{ENDIF}</td>{ENDIF}
+                        {IF journal_name != ""}<td>{journal_settings}{IF journal_settings != ""}<a href="?action=show_qa&zeder_id={zeder_id}&zeder_instance={zeder_instance}&delete_tag={tags}&delete_subfield_code={subfield_codes}&delete_type=local" title="Delete this rule" onclick="return confirm('Are you sure?')"><sup>x</sup></a>{ENDIF}</td>{ENDIF}
                     </tr>
                 {ENDLOOP}
             </tbody>
         </table>
 
-        <form id="add_form" method="post" action="?" autocomplete="off" onsubmit="return confirm('Are you sure?')">
+        <form id="add_form" method="post" action="?" autocomplete="off" onclick="ValidateRegex()" onsubmit="return confirm('Are you sure?')">
```
mtrojan-ub, line 57: If you use the onclick event of the form element, this function will trigger if the user clicks anywhere in the form (any input field, any button, even the background). So you should either move it to the input button or, even better, combine it with the existing onsubmit event.

cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -570,17 +570,22 @@ bool ProcessShowQASubActionAdd(const std::multimap<std::string, std::string> &cg
...
     if (add_type.empty() or add_tag.empty() or add_presence.empty())
         return false;
 
-    std::string journal_id_to_insert = db_connection->escapeAndQuoteString(zeder_journal_id);
-    if (add_type == "global")
-        journal_id_to_insert = "NULL";
+    const std::string regex_to_insert(add_regex.empty() ? "NULL"
+                                                        : db_connection->escapeAndQuoteString(add_regex));
+    const std::string journal_id_to_insert(add_type == "global" ? "NULL"
+                                                                : db_connection->escapeAndQuoteString(zeder_journal_id));
```
mtrojan-ub, line 582: the variable should  be renamed to journal_id as well

cpp/data/zotero_cgi/qa.html
ruschein: There is now.

ruschein: Done

cpp/data/installer/ub_tools.sql
ruschein: Renamed

cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -570,17 +570,22 @@ bool ProcessShowQASubActionAdd(const std::multimap<std::string, std::string> &cg
...
 
-    db_connection->queryOrDie("INSERT INTO metadata_presence_tracer (zeder_journal_id, marc_field_tag, field_presence) "
-                              " VALUES (" + journal_id_to_insert + ", " + db_connection->escapeAndQuoteString(add_tag) + ", "
-                              " " + db_connection->escapeAndQuoteString(add_presence) + ")");
+    db_connection->queryOrDie("INSERT INTO metadata_presence_tracer (journal_id, marc_field_tag, marc_subfield_code,"
```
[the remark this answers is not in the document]
ruschein: Done

cpp/cgi_progs/zotero_cgi.cc
ruschein: Done

cpp/data/zotero_cgi/qa.html
ruschein: Done

cpp/data/zotero_cgi/qa.html
```diff
@@ -33,17 +44,17 @@ <h1>QA Settings</h1>
                 </tr>
             </thead>
             <tbody>
-                {LOOP tags, global_settings, journal_settings}
+                {LOOP tags, subfield_codes, regexes, record_types, global_settings, journal_settings}
```
mtrojan-ub, line 47: Maybe put rules for different record types into separate columns?

cpp/data/installer/ub_tools.sql
mtrojan-ub: optional_regex was renamed to regex in the schema, but i can still find at least 5 occurrences in the code (e.g. when building sql statements) where it's still named optional_regex.

cpp/data/zotero_cgi/qa.html
```diff
@@ -33,17 +44,17 @@ <h1>QA Settings</h1>
...
                         <td>{global_settings}</td>
-                        {IF journal_name != ""}<td>{journal_settings}{IF journal_settings != ""}<a href="?action=show_qa&zeder_id={zeder_id}&zeder_instance={zeder_instance}&delete_tag={tags}&delete_type=local" title="Delete this rule" onclick="return confirm('Are you sure?')"><sup>x</sup></a>{ENDIF}</td>{ENDIF}
+                        {IF journal_name != ""}<td>{journal_settings}{IF journal_settings != ""}<a href="?action=show_qa&zeder_id={zeder_id}&zeder_instance={zeder_instance}&delete_tag={tags}&delete_subfield_code={subfield_codes}&delete_record_type={record_types}&delete_type=local" title="Delete this rule" onclick="return confirm('Are you sure?')"><sup>x</sup></a>{ENDIF}</td>{ENDIF}
                     </tr>
                 {ENDLOOP}
             </tbody>
         </table>
 
-        <form id="add_form" method="post" action="?" autocomplete="off" onsubmit="return confirm('Are you sure?')">
+        <form id="add_form" method="post" action="?" autocomplete="off" onsubmit="ValidateForm">
```
mtrojan-ub, line 57: `ValidateForm` => `return ValidateForm()`

cpp/data/installer/ub_tools.sql
ruschein: Fixed

cpp/data/zotero_cgi/qa.html
ruschein: Done

cpp/data/zotero_cgi/qa.html
mtrojan-ub: We talked about this on the phone => i will change the table myself after the PR has been merged.

PR #3679, opened 2020-11-25
Address maliciously scattered null bytes in "Boletín criminológico"

jriedl:


cpp/lib/src/ZoteroHarvesterDownload.cc
```diff
@@ -345,7 +345,14 @@ bool Crawler::getNextPage(CrawlResult * const crawl_result) {
                                            | WebUtil::ATTEMPT_TO_EXTRACT_JAVASCRIPT_URLS);
 
     std::vector<WebUtil::UrlAndAnchorTexts> urls_and_anchor_texts;
-    WebUtil::ExtractURLs(download_result.response_body_, next_url, WebUtil::ABSOLUTE_URLS, &urls_and_anchor_texts, EXTRACT_URL_FLAGS);
+
+    // Some pages are evil and include null bytes in strange places...
```
ruschein, line 349: Would it be better to put this inside ExtractURLs?

jriedl: Maybe, but I thought due to performance it reasons should execute as seldom as possible.

PR #3682, opened 2020-11-30
ZoteroHarvesterConversion: Min difference for language detection

mtrojan-ub:


cpp/lib/src/NGram.cc
```diff
@@ -316,9 +317,9 @@ void ClassifyLanguage(std::istream &input, std::vector<std::string> * const top_
 
     // Select the top scoring language and anything that's close (as defined by alternative_cutoff_factor):
     const double high_score(languages_and_scores[0].second);
-    top_languages->push_back(languages_and_scores[0].first);
-    for (unsigned i(1); i < languages_and_scores.size() and (languages_and_scores[i].second < alternative_cutoff_factor * high_score); ++i)
-        top_languages->push_back(languages_and_scores[i].first);
+    top_languages->emplace_back(languages_and_scores[0].first, languages_and_scores[0].second);
```
ruschein, line 320: Why not make top_languages be a vector of DetectedLanguage and add an operator< member function to it?

cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -546,9 +545,21 @@ void DetectLanguage(MetadataRecord * const metadata_record, const Config::Journa
...
-                                NGram::DEFAULT_NGRAM_NUMBER_THRESHOLD);
-        detected_language = top_languages.front();
+        std::vector<NGram::DetectedLanguage> detected_languages;
+        NGram::ClassifyLanguage(record_text, &detected_languages, journal_params.language_params_.expected_languages_,
+                                /*alternative_cutoff_factor = */ 0);
+        const auto top_language(detected_languages.front());
+        detected_language = top_language.language_;
+        if (detected_languages.size() > 1) {
+            const auto second_language(detected_languages[1]);
+            if (top_language.score_ - second_language.score_ < 0.01) {
```
ruschein, line 555: Not sure if this makes sense.  Let's discuss this briefly.

cpp/lib/src/NGram.cc
mtrojan-ub: not necessary

PR #3683, opened 2020-11-30
Riedl

jriedl:


cpp/data/installer/scripts/install_centos_packages.sh
```diff
@@ -61,6 +61,10 @@ dnf --assumeyes install llvm-toolset
 # Make Johannes happy :-)
 dnf --assumeyes install tmux emacs
 
+# Integrate custom certificates
+cp /usr/local/ub_tools/docker/zts/extra_certs/extra_certs.pem /etc/pki/ca-trust/source/anchors/eguzkilore.crt
```
mtrojan-ub, line 65: There should be an additional script in cpp/data/system_updates to install this certificate on all already installed machines.

PR #3689, opened 2020-12-03
rss_aggregator: Limit item description to 500 characters

mtrojan-ub:
... not sure if it is necessary to improve the algorithm (e.g. skip entities and multibyte characters)

### DIFF
--- a/cpp/lib/include/HtmlUtil.h
+++ b/cpp/lib/include/HtmlUtil.h
@@ -108,7 +108,7 @@ std::string StripHtmlTags(const std::string &text_with_optional_tags, const bool
 
 
 /** \brief Shorten a text but keep HTML structure. */
-std::string ShortenText(const std::string &html_document, const unsigned max_length);
+std::string ShortenText(const std::string &html_document, const size_t max_length);
 
 
 } // namespace HtmlUtil
--- a/cpp/lib/include/HtmlUtil.h
+++ b/cpp/lib/include/HtmlUtil.h
@@ -107,4 +107,8 @@ size_t ExtractAllLinks(const std::string &html_document, std::vector<std::string
 std::string StripHtmlTags(const std::string &text_with_optional_tags, const bool replace_entities = true);
 
 
+/** \brief Shorten a text but keep HTML structure. */
+std::string ShortenText(const std::string &html_document, const unsigned max_length);
+
+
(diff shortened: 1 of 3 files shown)


cpp/lib/include/HtmlUtil.h
```diff
@@ -107,4 +107,8 @@ size_t ExtractAllLinks(const std::string &html_document, std::vector<std::string
 std::string StripHtmlTags(const std::string &text_with_optional_tags, const bool replace_entities = true);
 
 
+/** \brief Shorten a text but keep HTML structure. */
+std::string ShortenText(const std::string &html_document, const unsigned max_length);
```
ruschein, line 111: size_t would be better because that's what std::string::size returns.

cpp/lib/src/HtmlUtil.cc
```diff
@@ -1019,4 +1019,32 @@ std::string StripHtmlTags(const std::string &text_with_optional_tags, const bool
...
+std::string ShortenText(const std::string &html_text, const unsigned max_length) {
+    std::string shortened_text;
+    shortened_text.reserve(html_text.size());
+
+    bool in_tag(false);
+    unsigned length(0);
+
+    static const std::string placeholder("...");
+    unsigned max_length_without_placeholder(max_length - placeholder.length());
+    for (auto ch(html_text.cbegin()); ch != html_text.cend(); ++ch) {
```
ruschein, line 1031: Any particular reason why you didn't use an implicit for-loop?

ruschein (changes requested):

cpp/lib/src/HtmlUtil.cc
mtrojan-ub: I tried to find an example for a similar implementation and copied code from HtmlUtil::StripHtmlTags.

cpp/lib/include/HtmlUtil.h
mtrojan-ub: OK, i changed it (and the length type inside the function as well)

cpp/lib/src/HtmlUtil.cc
ruschein: That is ancient code.  Please use "for (const char ch : html_text)" and replace "*ch" with "ch" everywhere in the loop body.

PR #3697, opened 2020-12-08
zotero_harvester: Allow domain-specific overrides for download_delay_time

mtrojan-ub:


cpp/lib/include/MiscUtil.h
```diff
@@ -243,4 +243,23 @@ enum class PreferredPathLocation { LEADING, TRAILING };
 bool AddToPATH(const std::string &new_directory_path, const PreferredPathLocation preferred_path_location);
 
 
+/** Search for key in container and return its value, or default value if not found. */
+template<typename ContainerType, typename KeyType, typename ValueType> ValueType GetContainerValueOrDefault(const ContainerType &container,
+                                                                                                            const KeyType &key,
+                                                                                                            const ValueType &default_value,
+                                                                                                            bool * const default_returned = nullptr)
+{
+    const auto iterator(container.find(key));
```
ruschein, line 252: I would name the iterator key_and_value.

cpp/lib/include/ZoteroHarvesterConfig.h
```diff
@@ -40,6 +42,27 @@ namespace Config {
 static constexpr unsigned DEFAULT_ZEDER_ID(0);
 
 
+
+struct DownloadDelayParams {
+    unsigned default_delay_;
+    unsigned max_delay_;
+    std::map<std::string,unsigned> domain_to_default_delay_map_;
```
ruschein, line 49: If we anticipate that the number of entries in these maps could go above 20, I think we should prefer hash maps, i.e. std::unordered_map.
Also, a space is missing after the comma.

cpp/lib/include/ZoteroHarvesterConfig.h
```diff
@@ -40,6 +42,27 @@ namespace Config {
 static constexpr unsigned DEFAULT_ZEDER_ID(0);
 
 
+
+struct DownloadDelayParams {
+    unsigned default_delay_;
```
ruschein, line 47: I'd add a comment to indicate the units or, even better, rename the field to something like default_delay_in_seconds_ or whatever is appropriate.

cpp/lib/include/ZoteroHarvesterConfig.h
```diff
@@ -40,6 +42,27 @@ namespace Config {
...
+    unsigned max_delay_;
+    std::map<std::string,unsigned> domain_to_default_delay_map_;
+    std::map<std::string,unsigned> domain_to_max_delay_map_;
```
ruschein, line 50: Missing space.

cpp/lib/include/ZoteroHarvesterConfig.h
```diff
@@ -40,6 +42,27 @@ namespace Config {
...
+public:
+    DownloadDelayParams() = default;
```
ruschein, line 52: I think you may want to initialise max_delay_ to 0.

ruschein (changes requested):

PR #3698, opened 2020-12-08
zotero_cgi: Minor fixes for qa.html

mtrojan-ub:


mtrojan-ub:
... and i am about to change the view.
The old way it looks is like this

grafik

In the future it should look more like this:
grafik

PR #3701, opened 2020-12-09
Tagging of superior works w/ subsystem tags.

ruschein:


java/solrmarc_mixin/src/de/unituebingen/ub/ubtools/solrmarcMixin/TuelibMixin.java
```diff
@@ -2208,6 +2208,20 @@ public String isSuperiorWork(final Record record) {
         return Boolean.toString(sprField.getSubfield('a') != null);
     }
 
+    public Set getSubsystemsForSuperiorWork(final Record record) {
```
mtrojan-ub, line 2211: just `Set`? not `Set<String>`?

PR #3705, opened 2020-12-10
New phase for SPR field w/ subsystem information.

ruschein:


cpp/pipelines/ixtheo_marc_pipeline_fifo.sh
```diff
@@ -361,7 +361,7 @@ wait
 
 StartPhase "Tags Which Subsystems have Inferior Records in Superior Works Records"
 (add_is_superior_work_for_subsystems GesamtTiteldaten-post-phase"$((PHASE-1))"-"${date}".mrc \
-    GesamtTiteldaten-post-pipeline-"${date}".mrc >> "${log}" 2>&1 && \
+    GesamtTiteldaten-post-phase"$PHASE"-"${date}".mrc >> "${log}" 2>&1 && \
```
mtrojan-ub, line 364: This is the last phase in the pipeline before marc_check, so post-pipeline would be correct here (didn't you just correct that this morning?)
But there is another phase called add_subsystem_tags, how about moving this phase next to the other or even merging the phases?

ruschein: I was paranoid and that's why I put it this late in the pipeline.  I should probably improve the logging on this phase to help us trouble shoot any issues having to do with it being in the wrong position.

PR #3718, opened 2020-12-18
Ruschein

ruschein:


cpp/get_dangling_title_references.cc
```diff
@@ -27,106 +27,73 @@
...
-                            LOG_WARNING("Could not find K10plus-PPN for referenced PPN " + referenced_ppn + " [PPN: " + ppn + "]");
-                            *dangling_log << new_ppn->second << ", K10+ PPN DOES NOT EXIST FOR \"" << referenced_ppn + "\"\n";
-                        }
-                        *dangling_log << new_ppn->second << "," << new_referenced_ppn->second << '\n';
-                    } else
-                        *dangling_log << ppn << "," << referenced_ppn << '\n';
-
-                    ++unreferenced_ppns;
-                }
+            if ((not consider_only_reviews or record.isReviewArticle())
```
mtrojan-ub, line 61: the first condition could be checked once for the whole record instead of checking it again and again for every single field

ruschein: Done

PR #3721, opened 2021-01-07
new subdir cpp/zotaut

mtrojan-ub:
- as discussed with @ruschein in november
- Makefile similar to cpp/pipelines subdir


cpp/zotaut/Makefile
```diff
@@ -0,0 +1,30 @@
+LIB := ../lib
+INC = $(LIB)/include
```
ruschein, line 2: Shouldn't this be the other way 'round?  I mean using "=" on the first line an ":=" on the second line?

ruschein (changes requested):

cpp/zotaut/Makefile
mtrojan-ub: The makefile was copied from the pipelines subdir. The other makefile (e.g. cpp/Makefile) also start with `LIB := ../lib`

ruschein: I was afraid of that!  According to what I found on Stackoverflow line 2 seems suspicious.  After reading https://www.gnu.org/software/make/manual/html_node/Setting.html it seems that this works but is probably not optimal.

PR #3723, opened 2021-01-12
zotero_cgi: Use fewer GET parameters

mtrojan-ub:


cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -76,6 +76,31 @@ std::string GetMinElementOrDefault(const std::vector<std::string> &elements, con
 }
 
 
+struct Journal {
+    unsigned id_;
```
ruschein, line 80: What is this ID?  Either a name change or a comment might be nice.

cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -76,6 +76,31 @@ std::string GetMinElementOrDefault(const std::vector<std::string> &elements, con
...
+    Zeder::Flavour zeder_flavour_;
+    std::string name_;
+};
+
+
+Journal GetJournalById(const unsigned id, DbConnection * const db_connection) {
+    db_connection->queryOrDie("SELECT id, zeder_id, zeder_instance, journal_name FROM zeder_journals WHERE id="
+                              + db_connection->escapeAndQuoteString(std::to_string(id)));
+    auto result_set(db_connection->getLastResultSet());
+    while (const auto &row = result_set.getNextRow()) {
```
ruschein, line 91: row should not be a reference.

cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -76,6 +76,31 @@ std::string GetMinElementOrDefault(const std::vector<std::string> &elements, con
...
+        Journal journal;
+        journal.id_ = id;
+        journal.zeder_id_ = StringUtil::ToUnsigned(row["zeder_id"]);
+        journal.zeder_flavour_ = Zeder::GetFlavourByString(row["zeder_instance"]);
+        journal.name_ = row["journal_name"];
+        return journal;
+    }
+
+    LOG_ERROR("Journal with id \"" + std::to_string(id) + "\" does not exist in database!");
```
ruschein, line 100: "id" is usually written as "ID" in English. https://english.stackexchange.com/questions/101248/how-should-the-abbreviation-for-identifier-be-capitalized#:~:text=In%20English%2C%20both%20letters%20should,or%20be%20in%20all%20caps.  https://www.dictionary.com/browse/id?s=t

cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -544,15 +569,16 @@ void ProcessShowDownloadedAction(const std::multimap<std::string, std::string> &
...
+            ResetDeliveredRecordsForJournal(journal.id_, db_connection);
         else
             UpdateRecordDeliveryStateAndTimestamp(id_to_reset, ZoteroHarvester::Util::UploadTracker::DeliveryState::RESET, db_connection);
         at_least_one_action_done = "true";
     }
 
-    names_to_values_map.insertScalar("zeder_id", zeder_id);
-    names_to_values_map.insertScalar("zeder_instance", zeder_instance);
-    names_to_values_map.insertScalar("group", group);
+    names_to_values_map.insertScalar("id", std::to_string(journal.id_));
```
ruschein, line 578: I don't think you need to convert journal.id_ to a std::string.  Template::Map:: insertScalar is overloaded on the second parameter.

cpp/data/zotero_cgi/delivered.html
```diff
@@ -24,11 +24,11 @@
     </head>
     <body>
         <h1>Zotero Harvesting</h1>
-        <h2>Delivered records for Zeder ID {zeder_id} ({zeder_instance})</h2>
+        <h2>Delivered records for &quot;{journal_name}&quot;</h2>
```
ruschein, line 27: Why are you not using literal quotes here?

ruschein (changes requested):

cpp/cgi_progs/zotero_cgi.cc
mtrojan-ub: Its the journal id => primary key in the database. I normally don't write this as prefix if it belongs to the main object. That's why i also called the other property "name_" and not "journal_name_". Usually you access this on a variable which is e.g. defined as `Journal journal;` and then you only need to write `journal.id_` instead of `journal.journal_id_` so it means less typing. Only in case of e.g. `zeder_journal_` it is useful to add a prefix, to make it clear that this is something special or external and no default property.

cpp/data/zotero_cgi/delivered.html
mtrojan-ub: Because i always try to use encoded entities when writing HTML text, to avoid later copy/paste errors e.g. when copying text into an attribute.

ruschein: I don't think that is very readable.  How often do you really copy text into an attribute and when does the reviewer not catch that?

mtrojan-ub: https://stackoverflow.com/a/46000063/13923163

PR #3731, opened 2021-01-15
Various sanity improvements.

ruschein:


cpp/lib/src/MapUtil.cc
```diff
@@ -86,9 +86,10 @@ void DeserialiseMap(const std::string &input_filename, std::unordered_map<std::s
         if (key.empty() or value.empty())
             LOG_ERROR("Bad input in \"" + input_filename + "\" on line " + std::to_string(line_no) + "!");
         if (revert_keys_and_values)
-            (*map)[value] = key;
-        else
-            (*map)[key] = value;
+            key.swap(value);
+        if (unlikely(map->find(key) != map->end()))
```
mtrojan-ub, line 90: There might be a problem when revert_keys_and_values is set and multiple different keys would have the same values, so we might want to check this before swapping as well

cpp/lib/src/DbConnection.cc
```diff
@@ -840,7 +840,9 @@ std::unordered_set<DbConnection::MYSQL_PRIVILEGE> DbConnection::mySQLGetUserPriv
             RegexMatcher::RegexMatcherFactory("GRANT (.+) ON [`']?" + database_name + "[`']?.* TO ['`]?" + user
                                               + "['`]?@['`]?" + host + "['`]?"));
 
-        if (mysql_privileges_matcher->matched(row[0])) {
+        if (not mysql_privileges_matcher->matched(row[0]))
+            LOG_WARNING("unexpectedly, no privileges were extracted from " + row[0]);
```
mtrojan-ub, line 844: sure it is unexpected, else it would not be a warning

PR #3734, opened 2021-01-18
zotaut: Renamed "...alreadyDelivered" to "...alreadyInDatabase"

mtrojan-ub:


cpp/lib/src/ZoteroHarvesterDownload.cc
```diff
@@ -808,7 +808,7 @@ std::unique_ptr<Util::Future<DirectDownload::Params, DirectDownload::Result>>
     // check if we have already delivered this URL
     if (not global_params_.force_downloads_
         and operation == DirectDownload::Operation::USE_TRANSLATION_SERVER
-        and upload_tracker_.urlAlreadyDelivered(source.url_.toString(), Util::UploadTracker::DELIVERY_STATES_TO_RETRY))
+        and upload_tracker_.urlAlreadyInDatabase(source.url_.toString(), /*delivery_states_to_ignore=*/Util::UploadTracker::DELIVERY_STATES_TO_RETRY))
```
jriedl, line 811: Sorry, don't quite understand the comment: Does it mean that `DELIVERY_STATES_TO_RETRY` can be ignored in this context?

PR #3756, opened 2021-02-03
One more approach for Spanish names

jriedl:


jriedl:
The lower two lines are a reinsertion and can be omitted in the review.

PR #3757, opened 2021-02-03
Address https://github.com/ubtue/DatenProbleme/issues/96

jriedl:


cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -890,6 +890,11 @@ std::string TruncateAbstractField(const std::string &abstract_field) {
 }
 
 
+std::string GetArticleNumIndicator() {
```
ruschein, line 893: The name of the function sounds like it would return a number.

ruschein (changes requested):

cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -1017,8 +1017,14 @@ void GenerateMarcRecordFromMetadataRecord(const MetadataRecord &metadata_record,
         _936_subfields.appendSubfield('d', issue);
 
     const std::string pages(metadata_record.pages_);
-    if (not pages.empty())
-        _936_subfields.appendSubfield('h', pages);
+    const std::string ARTICLE_NUM_INDICATOR("article");
```
ruschein, line 1020: static

jriedl: https://softwareengineering.stackexchange.com/questions/350501/is-there-any-benefit-to-to-define-constant-local-variables-as-static-c

ruschein: See the second comment on Stackexchange.  Also, this is an instance of a class and not a simple constant.  There is a difference.  If it were an integer constant for example, the compiler could optimise it away.  For a string constant that is probably not possible.  In general the compiler cannot know if it is permissible to hoist the constant out of the function and has to recreate it on every function invocation.  Basically that is because the string constructor can do anything it likes and the compiler cannot make assumptions as to what it does on repeated invocations.

PR #3759, opened 2021-02-04
Added support for inline attachments.

ruschein:


cpp/email_watcher.cc
```diff
@@ -198,21 +189,62 @@ std::map<std::string, time_t> LoadSectionNamesToLastSeenTimeWindowsMap() {
     return section_names_to_last_seen_time_map;
 }
 
-
+#if 0
```
mtrojan-ub, line 192: the `#if 0` section should be removed before merging

cpp/cgi_progs/translator.cc
```diff
@@ -720,8 +720,8 @@ void MailMyTranslations(DbConnection &db_connection, const IniFile &ini_file, co
     if (recipient.empty())
         return;
 
-    if (unlikely(not SendEmail("no-reply@ub.uni-tuebingen.de", recipient, "Your IxTheo Translations",
-                               mail_content.str(), EmailSender::DO_NOT_SET_PRIORITY, EmailSender::HTML)))
+    if (unlikely(not EmailSender::SimpleSendEmail("no-reply@ub.uni-tuebingen.de", { recipient }, "Your IxTheo Translations",
```
mtrojan-ub, line 723: Do we really need sender, priority and format in the simplest version?

cpp/email_watcher.cc
```diff
@@ -198,21 +189,62 @@ std::map<std::string, time_t> LoadSectionNamesToLastSeenTimeWindowsMap() {
...
+            LOG_DEBUG("\tSection " + section_name + " not found in section_names_to_last_seen_time_map!");
+            matched_section_names->emplace(section_name);
+            continue;
+        }
+        LOG_DEBUG("\tLast seen " + TimeUtil::TimeTToString(section_name_and_last_seen_time->second) + ".");
+
+        const unsigned overdue_time_window(section.getUnsigned("overdue_time_window")); // in hours
+        if ((NOW - section_name_and_last_seen_time->second) > overdue_time_window * 3600u) {
+            overdue_list += "No email from " + section.getString("from_host") + " with subject pattern "
+                            + section.getString("subject_pattern") + " within the last "
```
mtrojan-ub, line 229: subject_pattern should be quoted somehow

cpp/email_watcher.cc
```diff
@@ -286,11 +291,11 @@ int Main(int argc, char *argv[]) {
         else
             section_name_and_last_seen_time->second = NOW;
     }
+#if 0
```
mtrojan-ub, line 294: Another `#if 0` section

cpp/email_watcher.cc
ruschein: No because I am still working on it.  I had to adapt it to the new API o/w the submitted code wouldn't compile.

cpp/cgi_progs/translator.cc
ruschein: Yes, we do as you will see when you continue reading the rest of the PR.  I still think that "Simple" instead of "Simpler" is the better prefix.

cpp/email_watcher.cc
ruschein: Done!

cpp/email_watcher.cc
ruschein: Please see my comment above.

cpp/find_missing_references.cc
```diff
@@ -108,10 +108,10 @@ int Main(int argc, char *argv[]) {
         archive_writer.write(missing_references_text);
         archive_writer.close();
 
-        const auto status_code(EmailSender::SendEmail("nobody@nowhere.com", email_address, "Missing PPN's",
-                                                      "Attached is the new list of " + std::to_string(new_missing_ppns.size()) + " missing PPN('s).",
-                                                      EmailSender::DO_NOT_SET_PRIORITY, EmailSender::PLAIN_TEXT, /* reply_to = */"",
-                                                      /* use_ssl = */true, /* use_authentication = */true, { ZIP_FILENAME }));
+        const auto status_code(EmailSender::SendEmailWithFileAttachments(
+            "nobody@nowhere.com", { email_address }, "Missing PPN's",
```
mtrojan-ub, line 112: This should also be changed to `no_reply@ub.uni-tuebingen.de` when time permits

mtrojan-ub: How about setting and re-using a constant for this? EmailSender::DEFAULT_EMAIL_SENDER or something

PR #3762, opened 2021-02-08
HMAC

mtrojan-ub:
This should be merged & pulled before ubtue/tuefind#1400

### DIFF
--- a/cpp/data/system_updates/5.sh
+++ b/cpp/data/system_updates/5.sh
@@ -4,3 +4,4 @@ if [ ! -f ${HMAC_FILE} ]; then
     HASH=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
     echo "HMACkey = $HASH" > $HMAC_FILE
 fi
+
--- a/cpp/data/system_updates/5.sh
+++ b/cpp/data/system_updates/5.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 HMAC_FILE=/usr/local/vufind/local/tuefind/local_overrides/hmac.conf
 if [ ! -f ${HMAC_FILE} ]; then
-    HASH=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
+    HASH=$(cat /dev/urandom | tr --delete --complement 'a-zA-Z0-9' | fold --width=32 | head --lines=1)
     echo "HMACkey = $HASH" > $HMAC_FILE
 fi
-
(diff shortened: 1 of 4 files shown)


cpp/lib/src/StringUtil.cc
```diff
@@ -2830,4 +2832,16 @@ unsigned RomanNumeralToDecimal(const std::string &s) {
 }
 
 
+std::string GenerateRandom(const size_t length, const std::string &alphabet) {
+    std::default_random_engine random_engine(std::random_device{}());
+    std::uniform_int_distribution<> distribution(0, alphabet.size()-1);
```
ruschein, line 2837: Please always add spaces around operators.

cpp/data/system_updates/5.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+HMAC_FILE=/usr/local/vufind/local/tuefind/local_overrides/hmac.conf
+if [ ! -f ${HMAC_FILE} ]; then
+    HASH=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
```
ruschein, line 4: Please use the long option names for tr and fold.

cpp/lib/src/StringUtil.cc
```diff
@@ -2830,4 +2832,16 @@ unsigned RomanNumeralToDecimal(const std::string &s) {
...
+    auto get_random_character = [ alphabet, &distribution, &random_engine ](){
```
ruschein, line 2838: I assume that this can be const.

ruschein: Also, I am not a huge fan of declaring functions this way, but never mind...

cpp/lib/src/StringUtil.cc
```diff
@@ -2830,4 +2832,16 @@ unsigned RomanNumeralToDecimal(const std::string &s) {
...
+        return alphabet[ distribution(random_engine) ];
```
ruschein, line 2839: We don't put spaces around the index of a vector or string.

cpp/lib/src/StringUtil.cc
```diff
@@ -2830,4 +2832,16 @@ unsigned RomanNumeralToDecimal(const std::string &s) {
...
+    };
+    std::string generated_string(length, 0);
```
ruschein, line 2841: It is prettier to write '\0' instead of simply 0.  (No, I am *not* kidding!)

ruschein (changes requested):

PR #3766, opened 2021-02-12
email sender normalization

mtrojan-ub:


cpp/zotaut/update_zotero_enhancement_maps.sh
```diff
@@ -59,14 +59,14 @@ function SendEmail {
 
     if [ "$no_problems_found" = true ]; then
         if [[ "$MODE" = "LIVE" ]]; then
-            send_email --priority=low --sender="update_zotero_enhancement_maps@uni-tuebingen.de" --recipients="$EMAIL_ADDRESS" \
+            send_email --priority=low --sender="no-reply@ub.uni-tuebingen.de" --recipients="$EMAIL_ADDRESS" \
```
ruschein, line 62: How often do we call send_email with this sender email address?  Would it make sense to make this the default for send_email?

mtrojan-ub: You're absolutely right!

PR #3767, opened 2021-02-12
First feature-complete version. (rss)

ruschein:
see also ubtue/tuefind#1403

- don't forget to update crontab after merge!
- Migration problems when dropping old rss_aggregator table?


cpp/rss_aggregator.cc
```diff
@@ -176,56 +203,71 @@ unsigned ProcessSection(const std::string &flavour, const IniFile::Section &sect
...
-                              "WHERE pub_date >= '" + SqlUtil::TimeTToDatetime(now - HARVEST_TIME_WINDOW * 86400) + "' "
-                              "AND flavour=" + db_connection->escapeAndQuoteString(flavour) + " "
-                              "ORDER BY pub_date DESC");
-    DbResultSet result_set(db_connection->getLastResultSet());
-    while (const DbRow row = result_set.getNextRow())
-        harvested_items->emplace_back(SyndicationFormat::Item(row["item_title"], row["item_description"], row["item_url"], row["item_id"],
-                                                              SqlUtil::DatetimeToTimeT(row["pub_date"])),
-                                                              row["serial_name"], row["feed_url"]);
-    return result_set.size();
+    db_connection->queryOrDie("SELECT id,feed_name,feed_url FROM tuefind_rss_feeds WHERE subsystem_types LIKE '%"
```
mtrojan-ub, line 220: Why do you use LIKE %...% if we have the exact subsystem type given? (simply use = without %)

cpp/convert_rss_db_tables.cc
```diff
@@ -0,0 +1,116 @@
...
+            continue;
+        if (unlikely(not feed_info.isCompatibleWith(row["flavour"])))
+            LOG_ERROR("Item w/ item_id \"" + row["item_id"] + " has a flavour \"" + row["flavour"] +
+                      "\" which is incompatible with the subsystem_types \""
+                      + StlHelpers::ContainerToString(feed_info.subsystem_types_.cbegin(),
+                                                      feed_info.subsystem_types_.cend(), ","));
+        CopyItem(db_writer.get(), feed_info.id_, row["item_id"], row["item_url"], row["item_title"],
+                 row["item_description"], row["pub_date"], row["insertion_time"]);
+    }
+    db_reader.queryOrDie("DROP TABLE IF EXISTS vufind.rss_aggregator");
```
mtrojan-ub, line 113: I guess this line is just for testing purposes and should be removed later

mtrojan-ub: .... and: This is executed on the db_reader (ub_tools database) so it should never have any effect since the reader has the connection to the ub_tools and not the vufind database.

ruschein: Good catch.  It should have been ub_tools.rss_aggregator.  In any case there is no table named rss_aggregator in the vufind database.

cpp/rss_aggregator.cc
ruschein: I don't think that works with sets.

cpp/data/installer/ixtheo.cronjobs
```diff
@@ -33,5 +33,6 @@ LOG_DIR=/usr/local/var/log/tuefind
 0 22 * * * "$BIN/new_journal_alert" ixtheo {ixtheo_host} "IxTheo Team<no-reply@ub.uni-tuebingen.de>" "IxTheo Subscriptions" > "$LOG_DIR/new_journal_alert_ixtheo.log" 2>&1
 30 22 * * * "$BIN/new_journal_alert" relbib {relbib_host} "RelBib Team<no-reply@ub.uni-tuebingen.de>" "RelBib Subscriptions" > "$LOG_DIR/new_journal_alert_relbib.log" 2>&1
 0 23 * * * "$BIN/update_tad_email_acl.sh" "$EMAIL"
-30 23 * * * "$BIN/rss_aggregator" "/usr/local/var/lib/tuelib/rss_aggregator_relbib.conf" "$EMAIL" "/usr/local/vufind/public/relbib_docs/news.rss" > "$LOG_DIR/rss_aggregator_relbib.log" 2>&1
+30 23 * * * "$BIN/rss_aggregator" relbib "$EMAIL" "/usr/local/vufind/public/relbib_docs/news.rss" > "$LOG_DIR/rss_aggregator_relbib.log" 2>&1
+40 23 * * * "$BIN/rss_aggregator" ixtheo "$EMAIL" "/usr/local/vufind/public/ixtheo_docs/news.rss" > "$LOG_DIR/rss_aggregator_ixtheo.log" 2>&1
```
mtrojan-ub, line 37: In fact, there is no directory called "ixtheo_docs", this one is just called "docs".

cpp/data/system_updates/6.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
mtrojan-ub, line 1: This file needs to be renamed to 7.sh due to conflicts with the master branch.

cpp/data/installer/ixtheo.cronjobs
mtrojan-ub: And be aware that of course the feed file can already be generated for ixtheo, but keep in mind that if you would like to use the whole functionality on tuefind side, lots of things need to be changed (e.g. the templates are made for tuefind2 and specialized for relbib2, which has no effect on ixtheo yet and might also need some changes for krimdok).

Also, if you enable the ixtheo cronjob you should also enable the krimdok cronjob.

cpp/lib/include/TimeUtil.h
```diff
@@ -133,6 +133,13 @@ std::string TimeTToString(const time_t &the_time, const std::string &format = DE
                           const std::string &time_locale = UB_DEFAULT_LOCALE);
 
 
+bool StringToTimeT(const std::string &time_str, time_t * const unix_time);
+
+
+/** \note This function will abort if "time_str" cannot be converted to a valid time_t! */
```
mtrojan-ub, line 139: Please add information about which kinds of strings are allowed.

cpp/installer.cc
```diff
@@ -552,6 +552,12 @@ void InstallUBTools(const bool make_install, const OSSystemType os_system_type)
         FileUtil::MakeDirectoryOrDie(UBTools::GetTueLocalTmpPath(), /* recursive = */true);
     }
 
+    // ...then create /var/www/custom_rss_feeds
+    if (not FileUtil::Exists(UBTools::GetTuelibPath())) {
+        Echo("creating /var/www/custom_rss_feeds");
```
mtrojan-ub, line 557: I don't think we should put the feeds into a new subdir in /var/www.

As we discussed so far, there will be a VuFind Controller which calls the program every time a feed needs to be generated. The Controller could either take the target path as an argument (which would be more likely somewhere in /tmp and the VuFind Controller will then deliver it), or write it directly to stdout.

ruschein: We now write the XML output to /dev/stdout.

cpp/lib/include/TimeUtil.h
ruschein: Done!

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,311 @@
...
+
+        const auto description(HtmlUtil::ShortenText(harvested_item.item_.getDescription(), /*max_length = */500));
+        if (not description.empty())
+            xml_writer->writeTagsWithData("description", description);
+
+        xml_writer->writeTagsWithData("pubDate",
+                                      TimeUtil::TimeTToString(harvested_item.item_.getPubDate(), TimeUtil::RFC822_FORMAT,
+                                                                     TimeUtil::UTC));
+        xml_writer->writeTagsWithData("guid", harvested_item.item_.getId());
+        xml_writer->writeTagsWithData("tuefind:rss_title", harvested_item.feed_title_);
```
mtrojan-ub, line 115: The entire `xmlns:tuefind` namespace and all related entries may be removed, since TueFind will be able to get the related values from the db table in the future, see ubtue/tuefind#1412

ruschein: Then what would be a good strategy here?

cpp/data/installer/ixtheo.cronjobs
ruschein: We need different directories for ixtheo and relbib.  I only wanted to keep the option of also using this mechanism for KrimDok.  We don't need to activate it yet, or do we?

ruschein: Removed the IxTheo entry.

cpp/rss_subset_aggregator.cc
ruschein: I have eliminated any references to the tuefind namespace.

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,315 @@
...
+#include "util.h"
+#include "VuFind.h"
+#include "XmlWriter.h"
+
+
+namespace {
+
+
+[[noreturn]] void Usage() {
+    ::Usage("email_address [user_id]\n"
```
mtrojan-ub, line 46: 1. if a rss feed is generated for a single user to show him in the frontend, why would we want to pass an email address here?
2. If the aggregation takes place for all users + email is sent to them, the email addresses from the vufind db should be used, so why would we want to pass an email address here? For cc? And do we want to put these mails into several cronjobs for each instance?

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,319 @@
...
+    std::unordered_map<std::string, UserInfo> ids_to_user_infos_map;
+    while (const auto user_row = user_result_set.getNextRow())
+        ids_to_user_infos_map[user_row["id"]] =
+            UserInfo(user_row["id"], user_row["firstname"], user_row["lastname"], user_row["email"],
+                     user_row["tuefind_rss_feed_last_notification"]);
+
+    unsigned feed_generation_count(0), email_sent_count(0);
+    for (const auto &[user_id, user_info] : ids_to_user_infos_map) {
+        if (vufind_user_id.empty() and not EmailSender::IsValidEmailAddress(user_info.email_)) {
+            LOG_WARNING("no valid email address for vfind.user.id " + user_id + "!");
```
mtrojan-ub, line 297: typo vfind

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,319 @@
...
+    auto user_result_set(db_connection->getLastResultSet());
+    std::unordered_map<std::string, UserInfo> ids_to_user_infos_map;
+    while (const auto user_row = user_result_set.getNextRow())
+        ids_to_user_infos_map[user_row["id"]] =
+            UserInfo(user_row["id"], user_row["firstname"], user_row["lastname"], user_row["email"],
+                     user_row["tuefind_rss_feed_last_notification"]);
+
+    unsigned feed_generation_count(0), email_sent_count(0);
+    for (const auto &[user_id, user_info] : ids_to_user_infos_map) {
+        if (vufind_user_id.empty() and not EmailSender::IsValidEmailAddress(user_info.email_)) {
```
mtrojan-ub, line 296: vufind_user_id.empty() should be removed from this line

cpp/rss_subset_aggregator.cc
ruschein: Fixed.

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,319 @@
...
+            LOG_WARNING("no valid email address for vfind.user.id " + user_id + "!");
+            continue;
+        }
+
+        db_connection->queryOrDie("SELECT appellation,language,user_type FROM ixtheo_user");
```
mtrojan-ub, line 301: You cannot use the ixtheo_user table. This functionality will also be used for KrimDok!

cpp/rss_subset_aggregator.cc
ruschein: The user's email address is only needed if we don't generate a custom RSS feed.

mtrojan-ub: Maybe `and not` should be changed to `or not` then?

And how can you use the vufind user id in the corresponding warning if it is empty?

ruschein: No, we only want to check if the user's email address is valid if vufind_user_id.empty() is true.  If vufind_user_id is non-empty we don't need a valid user email.
As to the warning message: I am using "user_id" and not "vufind_user_id" in the error message.

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,315 @@
...
+#include "util.h"
+#include "VuFind.h"
+#include "XmlWriter.h"
+
+
+namespace {
+
+
+[[noreturn]] void Usage() {
+    ::Usage("--mode=(email|rss_xml) (user_id|error_email_address) subsystem_type]\n"
```
mtrojan-ub, line 46: 1) A user id is unique for all subsystem, so it shouldn't be necessary to add a subsystem_type if a user_id is given (only if --mode=email)
2) If i call it right now with --mode=rss_xml, the following messages are shown:

```
free(): double free detected in tcache 2
Abgebrochen (Speicherabzug geschrieben)
```

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,315 @@
...
+                max_insertion_time = insertion_time;
+        }
+    }
+    if (harvested_items.empty())
+        return false;
+
+    if (send_email) {
+        if (not SendEmail(subsystem_type, email_sender, user_email, user_address, language, harvested_items))
+            return true;
+        db_connection->queryOrDie("INSERT INTO user SET tuefind_rss_feed_last_notification='" + max_insertion_time
```
mtrojan-ub, line 225: This should be an UPDATE statement and no INSERT statement

cpp/rss_aggregator.cc
```diff
@@ -176,56 +203,72 @@ unsigned ProcessSection(const std::string &flavour, const IniFile::Section &sect
...
-                                                              row["serial_name"], row["feed_url"]);
-    return result_set.size();
+    db_connection->queryOrDie("SELECT id,feed_name,feed_url FROM tuefind_rss_feeds WHERE FIND_IN_SET('" + subsystem_type
+                              + "', subsystem_types) > 0");
+    DbResultSet feeds_result_set(db_connection->getLastResultSet());
+    std::unordered_map<std::string, FeedNameAndURL> feed_ids_to_names_and_urls_map;
+    while (const auto row = feeds_result_set.getNextRow())
+        feed_ids_to_names_and_urls_map[row["id"]] = FeedNameAndURL(row["feed_name"], row["feed_url"]);
+
+    const std::string NOW_AS_SQL_DATETIME(SqlUtil::TimeTToDatetime(std::time(nullptr) - HARVEST_TIME_WINDOW * 86400));
```
mtrojan-ub, line 227: Incorrect variable naming => value does not equal NOW

cpp/rss_aggregator.cc
```diff
@@ -37,21 +38,23 @@
...
+    ::Usage("[--use-web-proxy] subsystem_type email_address xml_output_path\n"
+            "where subsystem_type must be one of {ixtheo,relbib,krimdok}");
 }
 
 
-// These must be in sync with the sizes in data/ub_tools.sql (rss_aggregator table)
-const size_t MAX_ITEM_ID_LENGTH(100);
-const size_t MAX_ITEM_URL_LENGTH(512);
-const size_t MAX_ITEM_TITLE_LENGTH(200);
+// These must be in sync with the sizes in the vufind.rss_aggregator table!
```
mtrojan-ub, line 54: The correct new name is vufind.tuefind_rss_items

cpp/data/sql_updates/vufind.2
```diff
@@ -0,0 +1,42 @@
...
+    downloader_time_limit INT NOT NULL DEFAULT 30,
+    CONSTRAINT id_constraint UNIQUE (id),
+    CONSTRAINT feed_name_constraint UNIQUE (feed_name),
+    CONSTRAINT feed_url_constraint UNIQUE (feed_url(768)),
+    CONSTRAINT website_url_constraint UNIQUE (website_url(768))
+) DEFAULT CHARSET=utf8mb4 COLLATE utf8mb4_bin;
+
+
+CREATE TABLE IF NOT EXISTS vufind.tuefind_rss_items (
+    rss_feeds_id INT UNSIGNED NOT NULL,
```
mtrojan-ub, line 18: Please add missing id columns to tuefind_rss_items and tuefind_rss_subscriptions, see ubtue/tuefind#1403

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,318 @@
...
+    std::string link_;
+public:
+    ChannelDesc(const std::string &title, const std::string &link)
+        : title_(title), link_(link) { }
+};
+
+
+const std::map<std::string, ChannelDesc> subsystem_type_to_channel_desc_map = {
+    { "relbib",  ChannelDesc("RelBib RSS Aggregator",  "https://relbib.de/")                },
+    { "ixtheo",  ChannelDesc("IxTheo RSS Aggregator",  "https://itheo.de/")                 },
```
mtrojan-ub, line 72: ixtheo.de

cpp/rss_subset_aggregator.cc
ruschein: Done

cpp/rss_aggregator.cc
ruschein: Done

cpp/rss_aggregator.cc
ruschein: Done

cpp/rss_subset_aggregator.cc
ruschein: Done

cpp/data/sql_updates/vufind.2
ruschein: vufind.tuefind_rss_feeds has an id column already.  I have added an id column to vufind.tuefind_rss_items.

mtrojan-ub:
There seems to be similar logic in many places regarding rss_aggregator.cc and rss_subset_aggregator.cc. Could this be put into lib and be re-used somehow? examples:
- Structs to represent rss feeds / items
- Generate these structs by using the DB as data source
- Generating XML out of these structs

cpp/lib/src/LocalDataDB.cc
```diff
@@ -47,11 +47,10 @@ static void CreateTables(DbConnection * const db_connection) {
 LocalDataDB::LocalDataDB(const OpenMode open_mode): single_transaction_(open_mode == READ_WRITE) {
     db_connection_ = new DbConnection(UBTools::GetTuelibPath() + "local_data.sq3" /* must be the same path as in fetch_marc_updates.py */,
                                       (open_mode == READ_WRITE) ? DbConnection::CREATE : DbConnection::READONLY);
+    CreateTables(db_connection_);
```
mtrojan-ub, line 50: Is this technically possible, even if the sq3 file has been opened in READONLY mode?

cpp/data/system_updates/8.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -o errexit
```
mtrojan-ub, line 2: maybe add -o nounset just for consistency with other scripts? (+pipefail?)

cpp/lib/src/DbConnection.cc
```diff
@@ -348,9 +348,17 @@ bool DbConnection::queryFile(const std::string &filename) {
...
+    if (type_ == T_MYSQL) {
+        const auto individual_statements(SplitMySQLStatements(statements));
+        for (const auto &statement : individual_statements) {
+            if (::mysql_query(&mysql_, statement.c_str()) != 0) {
+                LOG_WARNING("Could not successfully execute statement \"" + statement + "\": SQL error code:"
+                            + std::to_string(::mysql_errno(&mysql_)));
+                return false;
+            }
+        }
+	return true;
```
mtrojan-ub, line 360: strange indentation here

cpp/data/system_updates/8.sh
ruschein: Done

cpp/lib/src/DbConnection.cc
ruschein: Should be fixed now.

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,318 @@
...
+               const std::vector<HarvestedRSSItem> &harvested_items)
+{
+    const auto template_filename_prefix(UBTools::GetTuelibPath() + "rss_email.template");
+    std::string template_filename(template_filename_prefix + "." + language);
+    if (not FileUtil::Exists(template_filename))
+        template_filename = template_filename_prefix + ".en";
+    static const std::string email_template(FileUtil::ReadStringOrDie(template_filename));
+
+    Template::Map names_to_values_map;
+    names_to_values_map.insertScalar("user_email", user_email);
```
mtrojan-ub, line 147: The template contains neither `user_email` nor `user_address`, but `user_name`, which is missing in `names_to_values_map`

cpp/data/sql_updates/vufind.2
```diff
@@ -0,0 +1,46 @@
...
+CREATE TABLE IF NOT EXISTS vufind.tuefind_rss_subscriptions (
+    id INT UNSIGNED NOT NULL AUTO_INCREMENT,
+    rss_feeds_id INT UNSIGNED NOT NULL,
+    user_id INT NOT NULL,
+    CONSTRAINT id_constraint UNIQUE (id),
+    FOREIGN KEY (rss_feeds_id) REFERENCES tuefind_rss_feeds(id) ON DELETE CASCADE,
+    FOREIGN KEY (user_id) REFERENCES vufind.user(id) ON DELETE CASCADE
+) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin;
+
+ALTER TABLE vufind.user ADD tuefind_rss_feed_send_emails BOOLEAN DEFAULT FALSE;
```
mtrojan-ub, line 45: This should be NOT NULL so it is similar to tuefind_subscribed_to_newsletter

cpp/data/system_updates/8.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -o errexit -o nounset -o pipefail
+cd /usr/local/ub_tools/cpp
+make convert_rss_db_tables
+./convert_rss_db_tables
```
mtrojan-ub, line 5: import_rss_aggregator_conf must be called first, or we won't have any feeds in the db which we can insert the items for!

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,318 @@
...
+                     user_row["tuefind_rss_feed_last_notification"]);
+
+    unsigned feed_generation_count(0), email_sent_count(0);
+    for (const auto &[user_id, user_info] : ids_to_user_infos_map) {
+        if (vufind_user_id.empty() and not EmailSender::IsValidEmailAddress(user_info.email_)) {
+            LOG_WARNING("no valid email address for vufind.user.id " + user_id + "!");
+            continue;
+        }
+
+        const std::string language("en");
```
mtrojan-ub, line 304: language is hardcoded here... you might wanna use data from the user.last_language column instead

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,318 @@
...
+};
+
+
+int Main(int argc, char *argv[]) {
+    if (argc != 4)
+        Usage();
+
+    std::string error_email_address, vufind_user_id;
+    if (std::strcmp(argv[1], "--mode=email") == 0)
+        vufind_user_id = argv[2];
```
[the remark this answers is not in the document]
ruschein: Done

cpp/rss_subset_aggregator.cc
ruschein: Done

cpp/data/system_updates/8.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -o errexit -o nounset -o pipefail
+cd /usr/local/ub_tools/cpp
+make import_rss_aggregator_conf
+./import_rss_aggregator_conf
```
mtrojan-ub, line 5: 2 arguments missing

ruschein: Done!

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,321 @@
...
+}
+
+
+// Yes, this function has a confusing name but I could not think of a better one. :-(
+// What is meant is how to address a user!
+std::string GenerateUserAddress(const std::string &first_name, const std::string &last_name) {
+    if (last_name.empty())
+        return first_name;
+
+    return first_name + " " + last_name;
```
mtrojan-ub, line 241: both first_name and last_name are not mandatory fields in the db. What should happen if both are empty? `Dear user`?

mtrojan-ub: what happens in new_journal_alert.cc here? can we find a unified solution?

cpp/data/rss_email.template.en
```diff
@@ -0,0 +1,25 @@
...
+                <th>Description</th>
+                <th>URL</th>
+            </tr>
+        </thead>
+        <tbody>
+        {LOOP titles,descriptions,links}
+            <tr>
+                <td>{titles}</td>
+                <td>{descriptions}</td>
+                <td><a href="{links}"/></td>
```
mtrojan-ub, line 20: if the a has only a href and no content, you can neither see nor click it in the mail.

cpp/data/rss_email.template.en
```diff
@@ -0,0 +1,25 @@
...
+<html>
+    <body>
+Dear ${user_name},
+
+The following items from your selected RSS feeds are new:</br>
+    <table id="table_delivered">
+        <thead>
+            <tr>
+                <th>Title</th>
+                <th>Description</th>
```
mtrojan-ub, line 11: Descriptions may contain lots of html tags which might screw up the table layout. I'd rather recommend to remove the descriptions column entirely and replace it with the name of the RSS feed instead (maybe including a hyperlink to it).

cpp/data/rss_email.template.en
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE HTML>
+<html>
+    <body>
+Dear ${user_name},
```
mtrojan-ub, line 4: The $ should be removed from this line

mtrojan-ub:
We should add an sql_update for ub_tools to delete old feed entries just to make sure we dont run into any errors before migrating:

```
DELETE FROM rss_aggregator WHERE feed_url = 'https://killingthebuddha.com/category/ktblog/feed/';
DELETE FROM rss_aggregator WHERE feed_url = 'https://religionnews.com/feed/';
```

cpp/rss_aggregator.cc
```diff
@@ -235,34 +278,33 @@ int ProcessFeeds(const std::string &flavour, const std::string &xml_output_filen
...
-    DbConnection db_connection;
-    IniFile ini_file(config_file_path);
-
     Downloader::Params params;
-    const std::string PROXY(ini_file.getString("", "proxy", ""));
-    if (not PROXY.empty()) {
-        LOG_INFO("using proxy: " + PROXY);
-        params.proxy_host_and_port_ = PROXY;
+    if (argc == 5) {
+        if (std::strcmp(argv[0], "--use-web-proxy") != 0)
```
mtrojan-ub, line 286: This should be argv[1]

cpp/rss_aggregator.cc
```diff
@@ -176,56 +203,72 @@ unsigned ProcessSection(const std::string &flavour, const IniFile::Section &sect
...
-        }
+    db_connection->queryOrDie("SELECT * FROM tuefind_rss_feeds WHERE FIND_IN_SET('" + subsystem_type
+                              + "', subsystem_types) > 0");
+    auto result_set(db_connection->getLastResultSet());
+    while (const auto row = result_set.getNextRow()) {
+        LOG_INFO("Processing feed \"" + row["feed_name"] + "\".");
+        const unsigned new_item_count(ProcessFeed(row["id"], row["feed_name"], row["feed_url"],
+                                                  row.getValue("title_suppression_regex"),
+                                                  row.getValue("strptime_format"), downloader, db_connection,
+                                                  StringUtil::ToUnsigned(row["downloader_time_limit"])));
```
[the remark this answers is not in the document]
ruschein: Done

cpp/rss_aggregator.cc
ruschein: Done

cpp/data/rss_email.template.en
```diff
@@ -0,0 +1,25 @@
...
+
+The following items from your selected RSS feeds are new:</br>
+    <table id="table_delivered">
```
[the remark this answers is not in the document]
ruschein: Copy-'n-paste error.  Removed!

cpp/data/rss_email.template.en
ruschein: Done

cpp/rss_subset_aggregator.cc
ruschein: I chose the same address for both now.

cpp/data/rss_email.template.en
ruschein: Fixed.

cpp/data/rss_email.template.en
```diff
@@ -0,0 +1,25 @@
...
+    <table>
+        <thead>
+            <tr>
+                <th>Item Title</th>
+                <th>Feed</th>
+                <th>Item URL</th>
+            </tr>
+        </thead>
+        <tbody>
+        {LOOP item_titles,feed_url,feed_name,item_urls}
```
mtrojan-ub, line 16: Loop variables should all be in plural form here as well as inside the loop

ruschein: Done

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,314 @@
...
+        feed_urls.emplace_back(harvested_item.feed_url_);
+        feed_names.emplace_back(harvested_item.feed_title_);
+    }
+    names_to_values_map.insertArray("item_titles", item_titles);
+    names_to_values_map.insertArray("item_urls", item_urls);
+    names_to_values_map.insertArray("feed_urls", feed_urls);
+    names_to_values_map.insertArray("feed_names", feed_names);
+
+    const auto email_body(Template::ExpandTemplate(email_template, names_to_values_map));
+    const auto retcode(EmailSender::SimplerSendEmail(email_sender, { user_email }, GetChannelDescEntry(subsystem_type, "title"),
```
mtrojan-ub, line 163: The log is pretty empty so far... do you think it's a good thing to call LOG_INFO for every mail address where SentEmail was successful?

ruschein: No, not really.

cpp/data/rss_email.template.en
```diff
@@ -0,0 +1,25 @@
...
+                <th>Item Title</th>
+                <th>Feed</th>
+                <th>Item URL</th>
+            </tr>
+        </thead>
+        <tbody>
+        {LOOP item_titles,feed_urls,feed_names,item_urls}
+            <tr>
+                <td>{item_titles}</td>
+                <td><a href="{feed_urls}">{feed_names></a></td>
```
mtrojan-ub, line 19: `2021-03-11T14:50:28 SEVERE ./rss_subset_aggregator: in int main(int, char **) --> caught exception: in Template::ProcessEndOfSyntax: error on line 19 expected '}' after variable expansion but found VARIABLE_NAME!`

cpp/data/rss_email.template.en
```diff
@@ -0,0 +1,25 @@
+<!DOCTYPE HTML>
+<html>
+    <body>
+Dear {user_name},
+
+The following items from your selected RSS feeds are new:</br>
```
mtrojan-ub, line 6: Couldn't test it due to the other SEVERE error yet, but the standalone </br> also looks wrong.

mtrojan-ub: Besides... the other newlines on line 4 and 5 normally won't have any effect, so it might be better if you do something like this:

<p>Dear {user_name},</p>
<p>The following items from your selected RSS feeds are new:</p>

cpp/rss_subset_aggregator.cc
```diff
@@ -0,0 +1,314 @@
...
+    static const std::string email_template(FileUtil::ReadStringOrDie(template_filename));
+
+    Template::Map names_to_values_map;
+    names_to_values_map.insertScalar("user_name", user_address);
+
+    std::vector<std::string> item_titles, item_urls, feed_urls, feed_names;
+    for (const auto &harvested_item : harvested_items) {
+        item_titles.emplace_back(HtmlUtil::HtmlEscape(harvested_item.item_.getTitle()));
+        item_urls.emplace_back(harvested_item.item_.getLink());
+        feed_urls.emplace_back(harvested_item.feed_url_);
```
mtrojan-ub, line 154: The end user may not have any RSS functionality installed in his browser, so instead of the feed_url we should add the website_url from the DB table.

cpp/data/system_updates/8.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -o errexit -o nounset -o pipefail
+cd /usr/local/ub_tools/cpp
```
mtrojan-ub, line 3: DB schema migration must be done anyway, but the data migration in this script should only be done if env variable $TUEFIND_FLAVOUR is ixtheo.

mtrojan-ub: Another problem might be that the rss_aggregator table in ub_tools will be dropped in `convert_rss_db_tables` but should be dropped for all systems. Maybe we only leave the lines 4 and 5 out (relbib-related) and leave `convert_rss_db_tables` in. On non-ixtheo systems, the table should be empty anyway.

=> i just checked this; for some reason we had old rows on ub15 (pub_date 2014-2019) but not on sobek, so i truncated the table on ub15 so the migration should not lead to any problems,

cpp/data/sql_updates/vufind.2
```diff
@@ -0,0 +1,46 @@
...
+CREATE TABLE IF NOT EXISTS vufind.tuefind_rss_subscriptions (
+    id INT UNSIGNED NOT NULL AUTO_INCREMENT,
+    rss_feeds_id INT UNSIGNED NOT NULL,
+    user_id INT NOT NULL,
+    CONSTRAINT id_constraint UNIQUE (id),
+    FOREIGN KEY (rss_feeds_id) REFERENCES tuefind_rss_feeds(id) ON DELETE CASCADE,
+    FOREIGN KEY (user_id) REFERENCES vufind.user(id) ON DELETE CASCADE
+) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin;
+
+ALTER TABLE vufind.user ADD tuefind_rss_feed_send_emails BOOLEAN NOT NULL DEFAULT FALSE;
```
mtrojan-ub, line 45: Since we use this column in a WHERE clause, it would be good to have an index on this field. (Please dont forget to add it to the tuefind PR as well!)

ruschein: Done

cpp/data/system_updates/8.sh
ruschein: Done

cpp/rss_subset_aggregator.cc
ruschein: Done

cpp/data/rss_email.template.en
ruschein: Done.

cpp/data/rss_email.template.en
ruschein: Fixed.

cpp/data/system_updates/8.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -o errexit -o nounset -o pipefail
+
+cd /usr/local/ub_tools/cpp
+
+if [[ "$TUEFIND_FLAVOUR" != "ixtheo" ]]; then
```
mtrojan-ub, line 6: This must be `==` !

ruschein: Fixed

PR #3768, opened 2021-02-12
VuFind 7

mtrojan-ub:
Should be merged & pulled together with ubtue/tuefind#1404.


cpp/installer.cc
```diff
@@ -796,6 +796,9 @@ void ConfigureSolrUserAndService(const VuFindSystemType system_type, const bool
                                              "solr soft nproc 65535\n");
     FileUtil::WriteString("/etc/security/limits.d/20-solr.conf", solr_security_settings);
 
+    if (SELinuxUtil::IsEnabled())
+        SELinuxUtil::Port::AddRecordIfMissing("http_port_t", "tcp", 8983);
```
ruschein, line 800: We might consider using the last 4 characters of Solr::DEFAULT_HOST_AND_PORT here or splitting up Solr::DEFAULT_HOST_AND_PORT.

mtrojan-ub: I'd rather split them up, because ports are not necessarily always 4 digits long

PR #3769, opened 2021-02-15
Riedl

jriedl:


cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -541,12 +553,7 @@ void DetectLanguage(MetadataRecord * const metadata_record, const Config::Journa
             record_text = metadata_record->title_ + " " + metadata_record->abstract_note_;
         else
             LOG_ERROR("unknown text field '" + journal_params.language_params_.source_text_fields_ + "' for language detection");
-
-        std::vector<NGram::DetectedLanguage> detected_languages;
-        NGram::ClassifyLanguage(record_text, &detected_languages, journal_params.language_params_.expected_languages_,
-                                /*alternative_cutoff_factor = */ 0);
-        const auto top_language(detected_languages.front());
-        detected_language = top_language.language_;
+        detected_language = TikaDetectLanguage(record_text);
```
mtrojan-ub, line 556: https://www.tutorialspoint.com/tika/tika_language_detection.htm

It seems like Tika only supports 18 languages, and some of the languages we already use in zotero_harvester.conf are not supported (afr, ara, may, ron, srp, tur). 

Maybe we need to use NGram as fallback if Tika cannot detect any language?

cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -511,6 +511,18 @@ void PostProcessAuthorName(std::string * const first_name, std::string * const l
 }
 
 
+const std::string TIKA_SERVER_DETECT_STRING_LANGUAGE_URL("http://localhost:9998/language/string");
```
mtrojan-ub, line 514: Installer Changes for Tika are missing.

I wonder - do we really need a server (either as running service or within a docker container), or would it be simpler to use the command line interface?

jriedl: Command line would mean spawning a java process on each call.

mtrojan-ub:
Urgent, changes to installer will be made later.

PR #3784, opened 2021-02-18
Riedl online first

jriedl:


cpp/data/sql_updates/ub_tools.27
```diff
@@ -0,0 +1 @@
+ALTER TABLE delivered_marc_records MODIFY COLUMN delivery_state ENUM('automatic', 'manual', 'error', 'ignore', 'reset', 'online_first') DEFAULT 'automatic' NOT NULL;
```
mtrojan-ub, line 1: This change should also be added to data/installer/ub_tools.sql

cpp/zotaut/validate_harvested_records.cc
```diff
@@ -42,7 +42,7 @@ namespace {
 
 
 [[noreturn]] void Usage() {
-    ::Usage("[--update-db-errors] marc_input marc_output missed_expectations_file email_address");
+    ::Usage("[--update-db-errors] marc_input marc_output online_first_files missed_expectations_file email_address");
```
mtrojan-ub, line 45: `online_first_files` should be renamed to `online_first_file` similar to `missed_expectations_file`

cpp/zotaut/validate_harvested_records.cc
```diff
@@ -466,13 +481,22 @@ int Main(int argc, char *argv[]) {
...
-            if (update_db_errors)
-                upload_tracker.archiveRecord(record, ZoteroHarvester::Util::UploadTracker::DeliveryState::ERROR,
-                                             error_messages);
-            delinquent_records_writer->write(record);
+            if (std::find(reasons_for_being_invalid.begin(), reasons_for_being_invalid.begin(), ONLINE_FIRST_OR_EARLY_VIEW_MESSAGE)
+                          != reasons_for_being_invalid.end())
+            {
+                online_first_records_writer->write(record);
+                ++online_first_record_count;
+                upload_tracker.archiveRecord(record, ZoteroHarvester::Util::UploadTracker::DeliveryState::ONLINE_FIRST,
```
mtrojan-ub, line 489: Please also have a close look at archiveRecord. Inside that function we should treat ONLINE_FIRST similar to the other DELIVERY_STATES_TO_RETRY states, so that the existing entry can be updated (and maybe set to DELIVERED)-

cpp/data/sql_updates/ub_tools.27
mtrojan-ub: And => did you implement any logic to skip the download for a certain timespan? I guess we also need an extra column for this.

jriedl: As to the extra column: see comment below.

cpp/zotaut/validate_harvested_records.cc
jriedl: I would intentionally locate the removal of online-first entries in ` journal_timeliness_checker` and entirely delete the entries from the DB after the timespan given there. Thus, they are either reloaded as valid articles or reinserted again if still present (r.g. in and RSS Feed). On the long run we avoid littering the DB with stale OF entries that never make it to an issue.

mtrojan-ub: journal_timeliness_checker is deprecated. The plan was to replace this program by pipelines/collect_journal_stats.cc together with logic inside of Zeder.

For transparency to the end user, it should be discussed with the librarians whether they want this information to stay in the database or not. Even for temporary listed information, there should be a change to zotero_cgi as well. If we decide to hide it from the frontend, there should at least be a comment there (and maybe in the wiki too).

cpp/lib/src/ZoteroHarvesterUtil.cc
```diff
@@ -710,6 +718,21 @@ void UploadTracker::registerZederJournal(const unsigned zeder_id, const std::str
 }
 
 
+void UploadTracker::deleteAllDeliveredOnlineFirstEntries(const unsigned zeder_id, const std::string &zeder_instance) {
```
mtrojan-ub, line 721: It might be useful to keep this function a bit more generic and pass the delivery_state as a parameter.

cpp/zotaut/journal_timeliness_checker.cc
```diff
@@ -36,22 +36,33 @@ namespace {
...
+     return max_delivered_at < ::time(nullptr) - update_window * 86400;
 }
 
 
 void ProcessJournal(ZoteroHarvester::Util::UploadTracker * const upload_tracker, const std::string &journal_name,
                     const std::string &zeder_id, const std::string &zeder_instance, const unsigned update_window,
                     std::string * tardy_list)
 {
-    const time_t max_delivered_at(upload_tracker->getLastUploadTime(StringUtil::ToUnsigned(zeder_id),
+    const time_t all_max_delivered_at(upload_tracker->getLastUploadTime(StringUtil::ToUnsigned(zeder_id),
```
mtrojan-ub, line 54: Under normal circumstances, we expect the journal to be updated within the update window. This would also mean that the online_first articles would never be removed. Shouldn't we compare each record's timestamp to a max timestamp instead?

cpp/zotaut/validate_harvested_records.cc
jriedl: As discussed: For the moment `journal_timeliness_checker` is still in use. Online first now shows up in the status with respective comments.

cpp/lib/src/ZoteroHarvesterUtil.cc
jriedl: Intentionally done like this. Other states already have logic to be removed/reset, so another approach seems dangerous.

cpp/zotaut/journal_timeliness_checker.cc
jriedl: Addressed in the current code.

cpp/zotaut/validate_harvested_records.cc
mtrojan-ub: I wonder if the logic in Zeder is implemented in the meantime... maybe we could rename journal_timeliness_checker and remove some deprecated functionality when time permits.

mtrojan-ub:
Seems like these changes should be tested on nu for a while before pulling them on ub28. Should we update ub28 with the changes from other PRs before this one is merged?

mtrojan-ub:
ub28 has been updated last week. @jriedl, do you still want to add something more (e.g. to zotero_cgi as you said this morning) or can the PR be merged now?

jriedl:
Adjusted OF handling in `zotero_cgi` and `zotero_harvester` now uploaded.

PR #3785, opened 2021-02-18
Allow line continuations with indentations

jriedl:


cpp/lib/src/IniFile.cc
```diff
@@ -585,13 +585,13 @@ void IniFile::processFile(const std::string &external_filename) {
             std::string buf;
             std::getline(ini_file, buf);
             ++getCurrentLineNo();
-            line += buf;
+            line += StringUtil::Trim(buf, " \t");
```
ruschein, line 588: Why not use TrimWhite?

jriedl: The `StringUtil::Trim()` approach is used in numerous positions in `IniFile.cc`. Use TrimWhite() anyway?

ruschein: Fair enough!

PR #3787, opened 2021-02-18
zotaut keywords 655

mtrojan-ub:


cpp/lib/src/FileUtil.cc
```diff
@@ -92,6 +92,19 @@ ReadLines::ReadLines(const std::string &path, const TrimMode trim_mode): trim_mo
 }
 
 
+std::vector<std::string> ReadLines::ReadOrDie(const std::string &path, const ReadLines::TrimMode trim_mode) {
+    std::vector<std::string> lines;
```
ruschein, line 96: This should be declared as close to its first use as possible.

cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -391,6 +391,12 @@ void StripBlacklistedTokensFromAuthorName(std::string * const first_name, std::s
 }
 
 
+bool Is655Keyword(const std::string &keyword) {
+    static const auto keyword_list(FileUtil::ReadLines::ReadOrDie(UBTools::GetTuelibPath() + "zotero-enhancement-maps/keywords_as_655.txt"));
```
ruschein, line 395: That seems like a strange name "keywords_as_655.txt".

cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -1088,8 +1094,13 @@ void GenerateMarcRecordFromMetadataRecord(const MetadataRecord &metadata_record,
         marc_record->insertField("773", _773_subfields);
 
     // Keywords
-    for (const auto &keyword : metadata_record.keywords_)
-        marc_record->insertField(MARC::GetIndexField(TextUtil::CollapseAndTrimWhitespace(keyword)));
+    for (const auto &keyword : metadata_record.keywords_) {
+        const std::string normalized_keyword(TextUtil::CollapseAndTrimWhitespace(keyword));
+        if (Is655Keyword(normalized_keyword))
+            marc_record->insertField("655", { { 'a', normalized_keyword } });
```
ruschein, line 1100: There is a variant of insertField that is simpler.

ruschein (changes requested):

cpp/lib/src/ZoteroHarvesterConversion.cc
mtrojan-ub: What would be a better name?

ruschein: marc655_keywords.txt?

PR #3788, opened 2021-02-19
AppArmorUtil - first version

mtrojan-ub:


cpp/lib/src/AppArmorUtil.cc
```diff
@@ -0,0 +1,99 @@
...
+        executable = "aa-audit";
+        break;
+    case COMPLAIN:
+        executable = "aa-complain";
+        break;
+    case ENFORCE:
+        executable = "aa-enforce";
+        break;
+    }
+    ExecUtil::ExecOrDie(ExecUtil::Which(executable), { profile_id });
```
ruschein, line 95: In general I am not a huge fan of using Which as it relies on a PATH being set. I assume that aa-audit and friends are in /sbin or /usr/sbin.

cpp/lib/src/AppArmorUtil.cc
```diff
@@ -0,0 +1,99 @@
...
+}
+
+
+}
```
ruschein, line 99: Missing comment.

ruschein (changes requested):

cpp/lib/src/AppArmorUtil.cc
```diff
@@ -0,0 +1,99 @@
...
+
+    return JSON::JSONNode::CastToObjectNodeOrDie("root", json_root);
+}
+
+
+std::vector<std::pair<std::string,PROFILE_MODE>> GetProfiles() {
+    const auto json_root(GetStatusAsJSON());
+    const auto json_profiles(json_root->getObjectNode("profiles"));
+    std::vector<std::pair<std::string,PROFILE_MODE>> profiles;
+    for (auto &profile_and_mode : *json_profiles) {
```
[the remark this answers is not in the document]
mtrojan-ub: No, because getObjectNode will abort if the node doesnt exist or cannot be converted.

cpp/lib/src/AppArmorUtil.cc
```diff
@@ -0,0 +1,99 @@
...
+    for (const auto &profile : profiles) {
+        if (profile_id == profile.first)
+            return profile.second;
+    }
+    LOG_ERROR("Profile not found: " + profile_id);
+}
+
+
+bool IsAvailable() {
+    return (ExecUtil::Which("aa-status") != "");
```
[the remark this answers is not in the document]
mtrojan-ub: Was copied from SELinuxUtil, i will change it there as well.

cpp/lib/src/AppArmorUtil.cc
mtrojan-ub: But if we rely too heavy on hardcoded paths, we might get problems if they change in future ubuntu distributions without noticing it.

ruschein: I believe that Ubuntu follows the FHS, see, for example, https://itectec.com/ubuntu/ubuntu-differences-between-bin-sbin-usr-bin-usr-sbin-usr-local-bin-usr-local-sbin/
Therefore it is extremely unlikely that paths for binaries are changed willy nilly.  There does seem to be a modern trend to shove everything from /usr/bin into /bin (and maybe also from /usr/sbin and /sbin) but even then there seem to be symlinks back to what used to be the old locations.

cpp/lib/src/SELinuxUtil.cc
```diff
@@ -55,7 +55,7 @@ Mode GetMode() {
 
 
 bool IsAvailable() {
-    return (ExecUtil::Which("getenforce") != "");
+    return (not ExecUtil::Which("getenforce").empty());
```
ruschein, line 58: Maybe we should cache the result here?

mtrojan-ub: ExecUtil::Which already uses a cache internally

PR #3792, opened 2021-02-23
Ruschein

ruschein:


cpp/data/installer/boot_notification.service
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Waity Service
+After=systend-user-sessions.service
```
mtrojan-ub, line 3: `systend` => This looks like a typo

cpp/data/installer/boot_notification.service
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Waity Service
```
mtrojan-ub, line 2: The Description sounds very generic

cpp/data/system_updates/6.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -o errexit -o nounset
+
+readonly SYSTEMD_SERVICE_DIR=/etc/systemd/system/
+mkdir --parents $(SYSTEMD_SERVICE_DIR)
+install --owner=root --group=root --mode=644 \
+        /usr/local/ub_tools/cpp/data/installer/boot_notification.service \
+        $(SYSTEMD_SERVICE_DIR)
+/usr/bin/systemctl start boot_notification.service
```
mtrojan-ub, line 9: When systemd service files are changed, you need to execute `systemctl daemon-reload` before accessing the service to make sure that the daemon has detected the new file

mtrojan-ub: Also, i guess you do not only want to start it, but also to enable it permanently. This might be even more important because if you start it, the mail will be sent after installation; more important is that we want to send it after each reboot.

cpp/installer.cc
```diff
@@ -578,6 +578,9 @@ void InstallUBTools(const bool make_install, const OSSystemType os_system_type)
     FileUtil::MakeDirectoryOrDie("/usr/local/run");
     RegisterSystemUpdateVersion();
 
+    // Install boot notification service:
+    SystemdUtil::InstallUnit(UB_TOOLS_DIRECTORY + "/cpp/data/installer/boot_notification.service");
```
mtrojan-ub, line 582: 1) This should not be related to a VuFind installation. It should rather be installed on all systems (e.g. zotaut servers)-
2) If you move it out of this function be careful, it should not be executed inside docker containers (if systemd is not available).

cpp/boot_notification.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -o errexit -o nounset
+
+
+# The following command does not return until the system is fully up and running!
+/usr/bin/systemctl is-system-running --wait
+
+/usr/local/bin/send_email --recipients=ixtheo-team@ub.uni-tuebingen.de --subject='One of our systems just booted up!' \
```
mtrojan-ub, line 8: Please put the hostname into the subject. It's just easier to keep the overview in our mailboxes.

ruschein: Done!

cpp/installer.cc
mtrojan-ub: Also, i am not sure if InstallUnit also enables the unit, please check!

ruschein: It is installed wherever we install ub_tools and not VuFind.  Also we only use it now if systemd is available.

cpp/data/installer/boot_notification.service
ruschein: Changed.

cpp/data/installer/boot_notification.service
ruschein: Fixed.

cpp/installer.cc
mtrojan-ub: Alright, sorry, the view on github was a bit irritating.
But still => Installing this service will not automatically enable it. This might also be incorrect for the other services!

cpp/data/system_updates/6.sh
mtrojan-ub: Also => do we really also want this to be installed on our development machines?

cpp/installer.cc
ruschein: Please have another look.

cpp/installer.cc
```diff
@@ -578,6 +580,13 @@ void InstallUBTools(const bool make_install, const OSSystemType os_system_type)
     FileUtil::MakeDirectoryOrDie("/usr/local/run");
     RegisterSystemUpdateVersion();
 
+    // Install boot notification service:
+    if (SystemdUtil::IsAvailable()) {
+        SystemdUtil::InstallUnit(UB_TOOLS_DIRECTORY + "/cpp/data/installer/boot_notification.service");
+        SystemdUtil::EnableUnit("boot_notification");
+        SystemdUtil::StartUnit("boot_notification");
```
mtrojan-ub, line 587: ... should not be started here, just enabled

cpp/data/system_updates/7.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -o errexit -o nounset
+
+readonly SYSTEMD_SERVICE_DIR=/etc/systemd/system/
+mkdir --parents $(SYSTEMD_SERVICE_DIR)
```
mtrojan-ub, line 5: i think mkdir should be removed.... if the directory doesnt exist, there will be no systemd installation and copying the file there will not have any effect

PR #3793, opened 2021-02-23
Truncating fields to prevent the program from crashing due to exceeding size. Added sed command for json data input file.

hlindeboom:


cpp/marc_tools/sed_command_convert_json_to_mrc.sed
```diff
@@ -0,0 +1,4 @@
+# sed command to use on json input data for convert_json_to_marc in order to unify data arrays
```
ruschein, line 1: json should be capitalized as JSON as it is an acronym standing for "JavaScript Object Notation".
Also, I would turn this into a bash script that converts the input file for convert_json_to_marc.  I think it is important to include CORE in the name and documentation for this file as we don't want to use this for any input for convert_json_to_marc but only for JSON files coming from CORE.

cpp/marc_tools/bash_script_convert_core_json_data
```diff
@@ -0,0 +1,20 @@
...
+# in order to use them as input for convert_json_to_mrc
+JSON_INPUT_FILE=$1
+JSON_OUTPUT_FILE=$2
+if [[ ! "$#" == 2 ]]; then
+    echo "Usage: $0 Input_File_Name Output_File_Name"
+    exit 1
+fi 
+if [ -e $JSON_OUTPUT_FILE ]; then
+  echo "File $JSON_OUTPUT_FILE already exists!"
+  read -p "Would you like to overwrite the File $JSON_OUTPUT_FILE ? [y/n] " -n 1 -r
```
ruschein, line 13: Usually we write our scripts to not require interactive usage.  You can either remove this test or test file descriptor 0 being interactive.

ruschein (changes requested):

cpp/marc_tools/bash_script_convert_core_json_data.sh
```diff
@@ -1,14 +1,14 @@
-#! /bin/bash
+#!/bin/bash
```
ruschein, line 1: Please rename this file and remove the bash_script_ prefix!  All our files ending in ".sh" are Bash scripts.

ruschein (changes requested):

PR #3796, opened 2021-02-25
sql_updates: Add new column "tuefind_uuid" to user table.

mtrojan-ub:
Can the sql_updater correctly handle the `DELIMITER` and `CREATE TRIGGER` statements?

For details how the new column is used, see ubtue/tuefind#1407


mtrojan-ub:
close+re-open as DRAFT

PR #3798, opened 2021-03-03
sql_updates: Add new column "tuefind_uuid" to user table.

mtrojan-ub:
Can the sql_updater correctly handle the `DELIMITER` and `CREATE TRIGGER` statements?
=> in the meantime it can, but the update needs to be rolled out on all servers before merging this PR into master!

For details how the new column is used, see ubtue/tuefind#1407

Make sure to reflect database changes in other PR before merging this one!


ruschein:
> Can the sql_updater correctly handle the `DELIMITER` and `CREATE TRIGGER` statements?
> 
> For details how the new column is used, see [ubtue/tuefind#1407](https://github.com/ubtue/tuefind/pull/1407)

It should be able to handle it.

mtrojan-ub:
I guess it should now.... but the new sql_updater version will not be installed on all systems before we rollout this one. On the other hand it might be useful to have these IDs available for the RSS feature as well => instead of IDs we could use the UUID for the url of the RSS feeds... but maybe we can change that later before the functionality is visible in the interface.

mtrojan-ub:
@ruschein: There still seems to be a problem. DbConnection::SplitMySQLStatements does not recognize the first DELIMITER statement. If you apply the contents through that function, you get the following output:

```
ALTER TABLE user ADD tuefind_uuid CHAR(36) DEFAULT NULL
ALTER TABLE user ADD CONSTRAINT tuefind_user_uuid UNIQUE (tuefind_uuid)
UPDATE user SET tuefind_uuid=(SELECT UUID())
ALTER TABLE user MODIFY COLUMN tuefind_uuid CHAR(36) NOT NULL
DELIMITER //CREATE TRIGGER before_user_insert BEFORE INSERT ON vufind.user FOR EACH ROW IF NEW.tuefind_uuid IS NULL THEN SET NEW.tuefind_uuid = UUID()
END IF
//DELIMITER
```

So, this means that DELIMITER // is valid sql syntax and after it the delimiter is a newline, not a semicolon. Can you please fix this?

PR #3800, opened 2021-03-03
normalize charset + collation of zeder_journals

mtrojan-ub:


cpp/data/sql_updates/ub_tools.28
```diff
@@ -0,0 +1 @@
+ALTER TABLE zeder_journals CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin;
```
ruschein, line 1: Should there be a corresponding change in ub_tools.sql for the collation?

ruschein: Apparently we inherit the character set and collation method from the ub_tools database and therefore there is no problem in ub_tools.sql.

PR #3802, opened 2021-03-03
zotaut: Re-allow force automatic language detection

mtrojan-ub:


cpp/lib/src/ZoteroHarvesterConfig.cc
```diff
@@ -339,7 +338,6 @@ JournalParams::JournalParams(const IniFile::Section &journal_section, const Glob
     ssgn_ = journal_section.getString(GetIniKeyString(SSGN), "");
     license_ = journal_section.getString(GetIniKeyString(LICENSE), "");
     selective_evaluation_ = journal_section.getBool(GetIniKeyString(SELECTIVE_EVALUATION), false);
-    force_language_detection_ = journal_section.getBool(GetIniKeyString(FORCE_LANGUAGE_DETECTION), false);
```
jriedl, line 342: This would stop the evaluation of `force_language_detection` in the config file. I thought these were two different aspects: Use `force_language_language_detection`to overrule an erroneous language code from the Translator and use the detected Tika result, whereas `*` is used to force a language known and set beforehand. If they are equal reintroducing `*` would add no further functionality as far as I see.

cpp/lib/src/ZoteroHarvesterConfig.cc
```diff
@@ -463,10 +461,12 @@ bool ParseExpectedLanguages(const std::string &expected_languages_string, Langua
...
         return true;
 
-    // force? (deprecated, treat as invalid)
+    // force
     std::string expected_languages(expected_languages_string);
-    if (expected_languages[0] == '*')
-        return false;
+    if (expected_languages[0] == '*') {
+        language_params->force_detection_ = true;
+        expected_languages = expected_languages.substr(1);
```
jriedl, line 468: Is a construct like `*eng` admissible? I thought `*` was a standanlone. In the last case I don't get the `substr` construct.

mtrojan-ub: `*` has never been allowed standalone, it is only allowed together with one or more languages, like  `*eng` or  `*eng,fre,ger`.

cpp/lib/src/ZoteroHarvesterConfig.cc
mtrojan-ub: The corresponding function is still called when evaluating the config file, see ParseExpectedLanguages on line 347. The `*` means that the language detection result will have priority over the language delivered by the Zotero Translator and therefore no error will be generated, see ZoteroHarvesterConversion.cc line 582 and following.

cpp/lib/src/ZoteroHarvesterConfig.cc
jriedl: Ah, ok - thanks for the explanation.

cpp/lib/src/ZoteroHarvesterConfig.cc
jriedl: But then there is no semantic difference to using `force_language_detection`?

mtrojan-ub: Thats correct - force_language_detection has just been moved to the LanguageSettings struct, because all parameters that need to be parsed by the ParseExpectedLanguages function belong there. The functionality is the same as with the old flag, except the old flag has been entirely disabled until now.

mtrojan-ub:
Sorry, i didn't notice that force_language_detection_ is a separate ini directive in the meantime! So as discussed with @jriedl i will close the PR, we will add the documentation to zotero_harvester.conf and discuss the matter via mail or in the next zotaut meeting.

PR #3805, opened 2021-03-09
ZoteroHarvesterConversion.cc: Changes to automatic language detection

mtrojan-ub:
Seems like the right thing was already done but the messages have been confusing. Variable names and log messages contained the words "detected language", even if the language was not detected at all, but instead chosen through configuration. Also, the messages have been in the wrong order. Also, the code has been changed very often in the past, so i made multiple changes:

- renamed `detected_language` to `configured_or_detected_language`. 
- added `configured_or_detected_info` which is added to the log messages, so the user can see if the language has been set due to configuration or due to detection
- changed the order of the checked conditions at the end, so that `force_language_detection` is checked first, before other conditions are checked. This way, the user will see that force always has priority.

(maybe it would also make things clearer if we rename `force_language_detection `to `force_language_override` => if only 1 language is configured, this parameter has an effect, but does not have anything to do with detection at all!)

### DIFF
--- a/cpp/lib/src/ZoteroHarvesterConversion.cc
+++ b/cpp/lib/src/ZoteroHarvesterConversion.cc
@@ -546,10 +546,12 @@ void DetectLanguage(MetadataRecord * const metadata_record, const Config::Journa
     if (journal_params.language_params_.expected_languages_.empty())
         return;
 
-    std::string detected_language;
-    if (journal_params.language_params_.expected_languages_.size() == 1)
-        detected_language = *journal_params.language_params_.expected_languages_.begin();
-    else {
+    std::string configured_or_detected_language, configured_or_detected_info;
+    if (journal_params.language_params_.expected_languages_.size() == 1) {
+        configured_or_detected_language = *journal_params.language_params_.expected_languages_.begin();
+        configured_or_detected_info = "configured";
+    } else {
+        configured_or_detected_info = "detected";
         std::string record_text;
         if (journal_params.language_params_.source_text_fields_.empty()
             or journal_params.language_params_.source_text_fields_ == "title")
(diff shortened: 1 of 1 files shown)


jriedl:
It has all got a bit confusing: From what I understood, the intention was to leave the current language logic as it is and (re-)introduce  the star operator that can only be used if one language is given and which forces the language setting no matter what. Besides specification of only one language (without star) is still possible and is treated the ordinary way, which means it can also fail or detect a mismatch. Is is possible to sort this out with @socheres directly?

mtrojan-ub:
So i talked to @socheres: The * and force_language_detection mean exactly the same. The * will only be re-introduced at the beginning of the sprz field in Zeder so that we do not need another field there, but the conf will stay splitted => * will only have an influence on the force_language_detection value, so the detection will stay the same. He also welcomes the clarified log messages.

I also told him that the librarians need to change the sprz value for the 16 journals in zeder (where force_language_detection is already set) at the same day we pull this change on nu, so that we do not lose any information in zotero_harvester.conf. (ub28 does not need to be updated the same day, because the meaning of the config entries stays the same).

Now i will start implementing the change to the Zeder sync process.

mtrojan-ub:
As discussed in our team meeting today, the functionality will be changed:
- `*` will **not** be re-introduced. Instead there will be 2 modes:
  - `!` will lead to the language from the configuration being forced. If multiple languages are set in the configuration, all of them will be written to generated marc records.
  - `?` will trigger language detection, and the (single language) result will be forced.
  - If neither of them are set, the old behaviour will stay as-is (detection will be triggered, no language will be set if the result differs from the zotero language).
- We also decided that the existing "force_language_detection" flag will be deprecated, but still be supported for backwards compatibility reasons. This will be treated same as `?` (if zotero_expected_languages doesnt start with `!`).

mtrojan-ub:
new PR will be opened as DRAFT as soon as changes are implemented.

socheres:
@mtrojan-ub these changes * > ! OR ? should be explained to the librarians in our next zotaut meeting

mtrojan-ub:
Of course! Sorry that we changed the definition again, but my colleagues insisted on a different solution (but which i also think is better than the original one right now), so we will explain on monday!

PR #3806, opened 2021-03-09
New one shot tool.

ruschein:


cpp/throwaways/check_archive_org_isbn_presence.cc
```diff
@@ -0,0 +1,133 @@
...
+
+} // unnamed namespace
+
+
+int Main(int argc, char *argv[]) {
+    if (argc != 4)
+        Usage();
+
+    const unsigned WORKER_THREAD_COUNT(StringUtil::ToUnsigned(argv[1]));
+    std::vector<std::thread> thread_pool((WORKER_THREAD_COUNT));
```
mtrojan-ub, line 117: double (()) truly needed here?

ruschein: Yes, this is because of the "most vexing parse": https://en.wikipedia.org/wiki/Most_vexing_parse

PR #3808, opened 2021-03-10
Riedl

jriedl:


cpp/cgi_progs/zotero_cgi.cc
```diff
@@ -832,7 +832,7 @@ int Main(int argc, char *argv[]) {
     const std::string include_online_first(GetCGIParameterOrDefault(cgi_args, "include_online_first", ""));
     std::string config_overrides(GetCGIParameterOrDefault(cgi_args, "config_overrides"));
     if (include_online_first.empty())
-         config_overrides.append(SKIP_ONLINE_FIRST_TRUE_DIRECTIVE);
+         config_overrides.append((config_overrides.empty() ? "" : " ") + SKIP_ONLINE_FIRST_TRUE_DIRECTIVE);
```
mtrojan-ub, line 835: The string will be parsed as INI section later. Maybe it would be better to use \n instead of a space here and to remove the \r from the SKIP_ONLINE_FIRST_TRUE_DIRECTIVE constant.

PR #3812, opened 2021-03-11
zotaut language detection '!' and '?'

mtrojan-ub:
- follow-up PR to #3805
- Allow multiple languages in MetadataRecord => multiple changes necessary to allow this
- Renamed+Splitted the very large DetectLanguage function into several smaller functions to make things clearer
- Added support for '!' and '?'
- Still keep backwards compatibility for non-merged force_language_detection entries

### DIFF
--- a/cpp/lib/src/ZoteroHarvesterConversion.cc
+++ b/cpp/lib/src/ZoteroHarvesterConversion.cc
@@ -535,24 +535,19 @@ std::string TikaDetectLanguage(const std::string &record_text) {
 
 void NormalizeGivenLanguages(MetadataRecord * const metadata_record) {
     // Normalize given languages
-    // We cant remove during iteration, so we use a multipass approach
-    std::set<std::string> unallowed_languages;
-    std::set<std::string> normalized_languages;
-    for (const auto &language : metadata_record->languages_) {
-        if (not Config::IsAllowedLanguage(language)) {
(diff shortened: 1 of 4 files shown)


cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -531,61 +533,90 @@ std::string TikaDetectLanguage(const std::string &record_text) {
...
+            LOG_INFO("Normalized language: " + language + " => " + normalized_language);
+            unallowed_languages.emplace(language);
+            normalized_languages.emplace(normalized_language);
+        }
+    }
+    for (const auto &language : unallowed_languages)
+        metadata_record->languages_.erase(language);
+    for (const auto &normalized_language : normalized_languages)
+        metadata_record->languages_.emplace(normalized_language);
+}
```
jriedl, line 556: Wouldn't it be more straighftorward to collect the results in a temporary object where the unallowed languages are just skipped and assign this object at the end?

mtrojan-ub: You're right, the code is really more complicated than it could be... i will simplify it. However, instead of copying the whole MetadataRecord, i'd rather prefer to copy only the languages_ property, so the amount of data touched is minimized to keep it efficient and as safe as possible (e.g. to avoid data loss if there is any bug in the copy constructor)

PR #3813, opened 2021-03-11
Be consistent!

ruschein:


cpp/lib/include/MARC.h
```diff
@@ -588,9 +588,16 @@ class Record {
...
 
     std::set<std::string> getDOIs() const;
+
+    /** Returns all ISSN's as normalised ISSN's, i.e. w/o hypens. */
     std::set<std::string> getISSNs() const;
+
+    /** Returns all superior ISSN's as normalised ISSN's, i.e. w/o hypens. */
     std::set<std::string> getSuperiorISSNs() const;
+
+    /** Returns all ISSN's as normalised ISSN's, i.e. w/o hypens. */
```
hlindeboom, line 598: Shouldn't this be ISBN? Otherwise the comment would be the same as the one in line 592.

ruschein: Fixed!

hlindeboom (approved):

PR #3815, opened 2021-03-11
keyword comparator

hlindeboom:


cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,145 @@
...
+        if (not record.getFirstSubfieldValue("150", 'a').empty())
+            gnd_keywords->emplace(std::make_pair(record.getFirstSubfieldValue("150", 'a'), record.getControlNumber()));
+        if (not record.getFirstSubfieldValue("150", 'g').empty())
+            gnd_keywords->emplace(std::make_pair(record.getFirstSubfieldValue("150", 'g'), record.getControlNumber()));
+        if (not record.getFirstSubfieldValue("150", 'x').empty())
+            gnd_keywords->emplace(std::make_pair(record.getFirstSubfieldValue("150", 'x'), record.getControlNumber()));
+    }
+
+    LOG_INFO("Processed " + std::to_string(record_count) + " MARC record(s).");
+
```
ruschein, line 63: There should be no blank line here.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,145 @@
...
+    for (const auto &keyword : *keywords_to_compare) {
+        const auto lookup(keywords_to_gnd->find(keyword));
+        if (lookup != keywords_to_gnd->end()) {
+            keyword_matches.insert(*lookup);
+            continue;
+        }
+        keywords_without_match.insert(keyword);
+    }
+    LOG_INFO("Found " + std::to_string(keyword_matches.size()) + " keyword matches.\n");
+    double percentage = (static_cast<double>(keyword_matches.size())/static_cast<double>(keywords_to_compare->size())) * 100;
```
ruschein, line 79: percentage should be const, also we use parentheses for variable initialisation and not equal signs.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,145 @@
...
+        if (not record.getFirstSubfieldValue("150", 'x').empty())
+            gnd_keywords->emplace(std::make_pair(record.getFirstSubfieldValue("150", 'x'), record.getControlNumber()));
+    }
+
+    LOG_INFO("Processed " + std::to_string(record_count) + " MARC record(s).");
+    
+}
+
+
+void FindEquivalentKeywords(std::unordered_map<std::string, std::string> const *keywords_to_gnd, std::unordered_set<std::string> const *keywords_to_compare, const std::string matches_output_file, const std::string no_matches_output_file) {
```
ruschein, line 67: It looks like you don't modify keywords_to_compare to compare in this function.  Therefore, please pass it in as a const reference.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,145 @@
...
+            continue;
+        }
+        keywords_without_match.insert(keyword);
+    }
+    LOG_INFO("Found " + std::to_string(keyword_matches.size()) + " keyword matches.\n");
+    double percentage = (static_cast<double>(keyword_matches.size())/static_cast<double>(keywords_to_compare->size())) * 100;
+    LOG_INFO("Which makes up for " + std::to_string(percentage) + "%\n");
+    LOG_INFO("Couldn't find a match for " + std::to_string(keywords_without_match.size()) + " keyword(s).\n");
+    std::ofstream output_file(matches_output_file);
+    for (const auto &[key, value] : keyword_matches) {
```
ruschein, line 83: We don't use braces for single statements in our C++ code.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,145 @@
...
+        output_file << "Keyword:  " << key << " PPN: " << value << "\n";
```
ruschein, line 84: For single characters, we use character constants and not string constants.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,145 @@
...
+    }
+    std::ofstream out_file(no_matches_output_file);
+    for (const auto &word : keywords_without_match) {
+        out_file << word << "\n";
```
ruschein, line 88: Ditto.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,145 @@
...
+        // read lines until newline character is not preceeded by a '\'
+        bool continued_line;
+        do {
+            std::string buf;
+            std::getline(input_file, buf);
+            line += StringUtil::Trim(buf, " \t");
+            if (line.empty())
+                continue;
+
+            continued_line = line[line.length() - 1] == '\\';
```
ruschein, line 114: You don't need a separate "continued_line" variable.  Also, just use line.back().

ruschein (changes requested):

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,167 @@
...
+            std::string key(StringUtil::Join(subfields, " "));
+            gnd_keywords->emplace(std::make_pair(key , record.getControlNumber()));
+        }
+    }
+    
+    LOG_INFO("Processed " + std::to_string(record_count) + " MARC record(s).");
+}
+
+
+void FindEquivalentKeywords(std::unordered_map<std::string, std::string> const *keywords_to_gnd, std::unordered_set<std::string> const &keywords_to_compare, const std::string matches_output_file, const std::string no_matches_output_file) {
```
ruschein, line 85: This seems fishy!  It seems that you should pass a reference and not a pointer here.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,167 @@
...
+    std::unordered_map<std::string, std::string> keyword_matches;
```
ruschein, line 86: We use the pattern keys_to_values_map to name all of our maps.  Of course you have to think of appropriate substitutions for "keys" and "values" here.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,167 @@
...
+    std::unordered_set<std::string> keywords_without_match;
+    for (const auto &keyword : keywords_to_compare) {
+        const auto lookup(keywords_to_gnd->find(keyword));
```
ruschein, line 89: I suspect that "keywords_to_gnd" should be named "keywords_to_gnd_numbers_map" and then "lookup" should be named "gnd_number" here.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,167 @@
...
+        if (lookup != keywords_to_gnd->end()) {
```
ruschein, line 90: If you reverse the logic you don't need a continue!  So, write:

if (lookup == keywords_to_gnd->end())
    keywords_without_match.insert(keyword);
else
       keyword_matches.insert(*lookup);

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,167 @@
...
+            keyword_matches.insert(*lookup);
+            continue;
+        }
+        keywords_without_match.insert(keyword);
+    }
+    LOG_INFO("Found " + std::to_string(keyword_matches.size()) + " keyword matches.\n");
+    const double percentage((static_cast<double>(keyword_matches.size())/static_cast<double>(keywords_to_compare.size())) * 100);
```
ruschein, line 97: Please put spaces on either side of all binary operators.  Here around the /.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,167 @@
...
+    LOG_INFO("Which makes up for " + std::to_string(percentage) + "%\n");
+    LOG_INFO("Couldn't find a match for " + std::to_string(keywords_without_match.size()) + " keyword(s).\n");
+    std::ofstream output_file(matches_output_file);
+    for (const auto &[key, value] : keyword_matches)
+        output_file << TextUtil::CSVEscape(key) << ',' << TextUtil::CSVEscape(value) << '\n';
+    std::ofstream out_file(no_matches_output_file);
+    for (const auto &word : keywords_without_match)
+        out_file <<TextUtil::CSVEscape(word) << '\n';
```
ruschein, line 105: Missing space.

cpp/marc_tools/single_marc_convert_keywords.cc
ruschein: Please don't forget this change.

ruschein (changes requested):

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,156 @@
...
+[[noreturn]] void Usage() {
+    ::Usage("gnd_input keyword_input keyword_matches_output keyword_without_matches_output"
+               "       Searches for keyword matches in the \"gnd_input\" file."
+               "       Returns a \"keyword_matches_output\" file with matching keywords and their PPN,"
+               "       as well as \"keywords_without_matches\" file containing keywords where no matches were found.\n");
+}
+
+
+void ExtractSubfieldsForTag(const MARC::Record &record, const std::string &subfield_tag, const std::string &subfield_codes, std::vector<std::string> * const subfields) {
+    const auto gnd_and_fields_iterator(record.findTag(subfield_tag));
```
ruschein, line 47: There are no such things as "subfield tags". Fields have tags and subfields have subfield codes!

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,156 @@
...
+    if (record.findTag(subfield_tag) == record.end())
```
ruschein, line 48: It is pointless and inefficient to call findTag again here.  Please just reuse the return value of the previous call to findTag.

cpp/marc_tools/single_marc_convert_keywords.cc
ruschein, line 47: We would typically call this XXX_field where XXX describes what kind of a field it is.  We don't add a _iterator suffix.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,156 @@
...
+        return;
+    const MARC::Subfields subfield_tag_subfields(gnd_and_fields_iterator->getSubfields());
```
ruschein, line 50: The name subfield_tag_subfields is wrong and confusing.  Just name it "subfields", please.

ruschein: Also, you could even replace MARC:Subfields with "auto" bit I won't insist on it.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,156 @@
...
+    for (const char subfield_code : subfield_codes) {
+        const std::vector<std::string> subfield_values(subfield_tag_subfields.extractSubfields(subfield_code));
+        if (subfield_values.empty()) {
```
ruschein, line 53: Please combine the conditions of the two if statements into a single condtion with "and".

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,156 @@
...
+            if (subfield_code == 'a')
+                LOG_WARNING("Entry has no Subfield 'a' for PPN " + record.getControlNumber());
+            continue;
+        }
+        std::string subfield_value(subfield_values[0]);
+        if (subfield_code == 'x')
+            subfield_value = '(' + subfield_value + ')';
+        subfields->emplace_back(subfield_value);
+        }
```
ruschein, line 62: Something is fishy with the indentation here.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,156 @@
...
+    }
+/*    if (not record.hasTag(subfield_tag))
```
ruschein, line 64: Please don't keep commented-out code in what you submit to GitHub.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,156 @@
...
+        else if (subfield_code == 'a')
+            LOG_WARNING("Entry has no Subfield 'a' for PPN " + record.getControlNumber());
+    } */
+
+
+void ReadInGndKeywords(MARC::Reader * const marc_reader, std::unordered_map<std::string, std::string> * const gnd_keywords)
+{
+    unsigned record_count(0);
+
+    while (MARC::Record record = marc_reader->read()) {
```
ruschein, line 82: I suspect that record can be const.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,167 @@
...
+        ++record_count;
+
+        std::vector<std::string> subfields;
+        extractSubfieldsForTag(record, "150", "agx", &subfields);
+        //entry for every single 'a' subfield
+        if (subfields.size() == 1)
+            gnd_keywords->emplace(std::make_pair(subfields[0], record.getControlNumber()));
+        //entry for all subfields combined
+        if (subfields.size() > 1) {
```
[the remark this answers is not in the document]
ruschein: Please use an "else if".

cpp/marc_tools/single_marc_convert_keywords.cc
ruschein: Please change the variable name.

cpp/marc_tools/single_marc_convert_keywords.cc
ruschein: Not fixed!

cpp/marc_tools/single_marc_convert_keywords.cc
ruschein: Not fixed.

ruschein (changes requested):

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -59,69 +59,65 @@ void ExtractSubfieldsForTag(const MARC::Record &record, const std::string &subfi
...
-        ExtractSubfieldsForTag(record, "110", "agx", &subfields);
-        ExtractSubfieldsForTag(record, "111", "agx", &subfields);
-        ExtractSubfieldsForTag(record, "130", "agx", &subfields);
-        ExtractSubfieldsForTag(record, "151", "agx", &subfields);
-        // second entry for all subfields combined
-        if (subfields.size() > 1) {
-            const std::string key(StringUtil::Join(subfields, " "));
-            gnd_keywords->emplace(std::make_pair(key , record.getControlNumber()));
-        }
+        AddMainSubfieldAndCombinationsToGndKeywords(record, keywords_to_gnd_numbers_map, "150", "agx", &subfields);
```
ruschein, line 90: You're not using "subfields" in this function so please take it out here and declare it as a local variable in AddMainSubfieldAndCombinationsToGndKeywords.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -59,69 +59,65 @@ void ExtractSubfieldsForTag(const MARC::Record &record, const std::string &subfi
...
+        AddMainSubfieldAndCombinationsToGndKeywords(record, keywords_to_gnd_numbers_map, "130", "agx", &subfields);
+        AddMainSubfieldAndCombinationsToGndKeywords(record, keywords_to_gnd_numbers_map, "151", "agx", &subfields);
     }
     LOG_INFO("Processed " + std::to_string(record_count) + " MARC record(s).");
 }
 
 
 void FindEquivalentKeywords(std::unordered_map<std::string, std::string> const &keywords_to_gnd_numbers_map, std::unordered_set<std::string> const &keywords_to_compare, const std::string matches_output_file, const std::string no_matches_output_file) {
-    std::unordered_map<std::string, std::string> keyword_matches;
+    std::unordered_map<std::string, std::string> keyword_to_ppn_map;
```
ruschein, line 102: The pattern we use for maps is keys_to_values_map, that is keys and values are plurals. So please rename this to keywords_to_ppns_map.

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -59,69 +59,65 @@ void ExtractSubfieldsForTag(const MARC::Record &record, const std::string &subfi
...
         const auto gnd_number(keywords_to_gnd_numbers_map.find(keyword));
         if (gnd_number == keywords_to_gnd_numbers_map.end())
             keywords_without_match.insert(keyword);
         else
-            keyword_matches.insert(*gnd_number);
+            keyword_to_ppn_map.insert(*gnd_number);
     }
-    LOG_INFO("Found " + std::to_string(keyword_matches.size()) + " keyword matches.\n");
-    const double percentage((static_cast<double>(keyword_matches.size())/static_cast<double>(keywords_to_compare.size())) * 100);
+    LOG_INFO("Found " + std::to_string(keyword_to_ppn_map.size()) + " keyword matches.\n");
```
ruschein, line 111: Please put parentheses around "es" here as I assume it can happen that the map size is 1.

ruschein (changes requested):

cpp/marc_tools/single_marc_convert_keywords.cc
```diff
@@ -0,0 +1,149 @@
...
+
+#include <fstream>
+#include <iostream>
+#include <set>
+#include <stdexcept>
+#include <utility>
+#include <climits>
+#include <cstdio>
+#include <cstdlib>
+#include "FileUtil.h"
```
ruschein, line 28: You not using this anywhere.  Please remove it.  Also this PR is still set to Draft!

PR #3821, opened 2021-03-16
AppArmor profile for apache2

mtrojan-ub:


cpp/data/installer/apparmor/apache2
```diff
@@ -0,0 +1,31 @@
...
+  /usr/local/run/** rw,
+  /usr/local/var/lib/tuelib/** rw,
+  /usr/local/var/log/tuefind/** rw,
+  /usr/local/var/tmp/** rw,
+  /var/www/cgi-bin/** pux,
+
+  # tuefind-specific
+  /usr/local/vufind/** r,
+  /usr/local/vufind/local/tuefind/instances/** rwk,
+
```
ruschein, line 30: I'd remove this blank line.  Also, is the comma at the end of the previous line correct?

mtrojan-ub: The blank line has been removed. The comma is optional but allowed, so it's easier to copy/paste. (e.g. newer PHP versions also allow an optional comma at end for array values and even function calls).

cpp/data/installer/apparmor/apache2
```diff
@@ -0,0 +1,30 @@
+# vim:syntax=apparmor
+#include <tunables/global>
+
+/usr/sbin/apache2 flags=(attach_disconnected) {
+  #include <abstractions/base>
```
ruschein, line 5: From https://gitlab.com/apparmor/apparmor/-/wikis/QuickProfileLanguage "The # is optional and shouldn't be used in newer
profiles. "

ruschein (changes requested):

cpp/data/installer/apparmor/apache2
```diff
@@ -0,0 +1,30 @@
...
+
+  #include <abstractions/apache2-common>
+  #include <abstractions/mysql>
+  #include <abstractions/php>
+  #include <abstractions/postfix-common>
+
+  # apache2 basics
+  /etc/** r,                    # apache.conf, also /etc/ssl and others
```
ruschein, line 13: Maybe we should be more specific here.  For example it can be dangerous to be able to read the contents of /etc/passwd

cpp/data/installer/apparmor/apache2
mtrojan-ub: Has been removed now, however i could not find any example in my /etc/apparmor.d that uses `include` without `#`.

cpp/data/installer/apparmor/apache2
mtrojan-ub: done

PR #3822, opened 2021-03-16
Add support for notfications about newly harvested selective evaluati…

jriedl:
…on journals


cpp/zotaut/journal_timeliness_checker.cc
```diff
@@ -47,23 +49,69 @@ bool MaxDeliveredAtSmallerThanUpdateWindow(const time_t &max_delivered_at, const
 }
 
 
+
+std::unique_ptr<KeyValueDB> CreateOrOpenKeyValueDB(const std::string &name) {
```
mtrojan-ub, line 53: This function looks very generic, it could be moved to KeyValueDb.h and .cc as a static factory method.

cpp/zotaut/journal_timeliness_checker.cc
```diff
@@ -47,23 +49,69 @@ bool MaxDeliveredAtSmallerThanUpdateWindow(const time_t &max_delivered_at, const
...
-    // Make sure articles stored as online first are retried after an update_window period
-    upload_tracker->deleteOnlineFirstEntriesOlderThan(StringUtil::ToUnsigned(zeder_id), zeder_instance, update_window);
+     // Make sure articles stored as online first are retried after half an update_window period or at most 30 days
+    upload_tracker->deleteOnlineFirstEntriesOlderThan(StringUtil::ToUnsigned(zeder_id), zeder_instance, std::min(update_window / 2 , static_cast<unsigned>(30)));
 
     const time_t max_delivered_at(upload_tracker->getLastUploadTime(StringUtil::ToUnsigned(zeder_id),
                                   Zeder::ParseFlavour(zeder_instance)));
 
+
+    // For "selective evalution"-journals insert a  notification
```
mtrojan-ub, line 91: double space (a... notification)

cpp/zotaut/journal_timeliness_checker.cc
```diff
@@ -47,23 +49,69 @@ bool MaxDeliveredAtSmallerThanUpdateWindow(const time_t &max_delivered_at, const
...
+std::unique_ptr<KeyValueDB> CreateOrOpenKeyValueDB(const std::string &name) {
+    const std::string DB_FILENAME(UBTools::GetTuelibPath() + name + ".db");
+    if (not FileUtil::Exists(DB_FILENAME))
+        KeyValueDB::Create(DB_FILENAME);
+
+    return std::unique_ptr<KeyValueDB>(new KeyValueDB(DB_FILENAME));
+}
+
+
+time_t GetLastNotificationTime(std::unique_ptr<KeyValueDB> &notified_db,  const std::string &zeder_id) {
```
mtrojan-ub, line 62: zeder_id is not unique, this should always be a combination of zeder_id+zeder_instance, or use zeder_journals.id from the DB instead

mtrojan-ub:
Instead of using another KeyValueDB, i think it would be better to add this information to the MySQL database for several reasons (keep similar data together / we already have journal-related and timestamp-related data there, and also easy possibility to show it in the cgi program if needed).

cpp/zotaut/journal_timeliness_checker.cc
```diff
@@ -91,9 +94,10 @@ int Main(int argc, char *argv[]) {
         if (section.find("user_agent") != section.end())
             continue; // Not a journal section.
 
-        const auto delivery_mode(section.getEnum("zotero_delivery_mode", ZoteroHarvester::Config::STRING_TO_UPLOAD_OPERATION_MAP,
-                                                 ZoteroHarvester::Config::UploadOperation::NONE));
-        if (delivery_mode != ZoteroHarvester::Config::UploadOperation::LIVE or section.getBool("zeder_newly_synced_entry", false))
+        const auto delivery_mode(static_cast<ZoteroHarvester::Config::UploadOperation>(
+             section.getEnum("zotero_delivery_mode", ZoteroHarvester::Config::STRING_TO_UPLOAD_OPERATION_MAP,
```
mtrojan-ub, line 98: Normally, zotaut ini keys should not be referenced directly with a string; instead ZoteroHarvester::Config::GetIniKeyString should be used. However a) this is very complicated and i've never been happy with that solution either and b) there are already other parts of this file that use keys directly, so i guess we just leave it as-is for the moment before we open up another big construction site for a small change like this...

cpp/zotaut/journal_timeliness_checker.cc
```diff
@@ -91,9 +94,10 @@ int Main(int argc, char *argv[]) {
...
+                             ZoteroHarvester::Config::UploadOperation::NONE)));
+        if (section.getBool("zeder_newly_synced_entry", false))
```
mtrojan-ub, line 100: Is this really correct? Shouldn't we skip entries that are newly synced, instead of processing only them?
(and if the delivery_mode has no effect on this condition, we might as well perform this check before getting the delivery mode)

cpp/zotaut/journal_timeliness_checker.cc
```diff
@@ -48,17 +48,20 @@ bool MaxDeliveredAtSmallerThanUpdateWindow(const time_t &max_delivered_at, const
...
 
 void ProcessJournal(ZoteroHarvester::Util::UploadTracker * const upload_tracker, const std::string &journal_name,
-                    const std::string &zeder_id, const std::string &zeder_instance, const unsigned update_window,
+                    const std::string &zeder_id, const std::string &zeder_instance,
+                    const ZoteroHarvester::Config::UploadOperation delivery_mode, const unsigned update_window,
                     std::string * tardy_list)
 {
-    // Make sure articles stored as online first are retried after an update_window period
-    upload_tracker->deleteOnlineFirstEntriesOlderThan(StringUtil::ToUnsigned(zeder_id), zeder_instance, update_window);
+     // Make sure articles stored as online first are retried after half an update_window period or at most 14 days
```
mtrojan-ub, line 55: unneccesary space before `//`

cpp/zotaut/journal_timeliness_checker.cc
jriedl: The approach of determining the keys is identical to the one already present beforehand. Only the type conversion with the `enum` had to be newly introduced to enable passing the result to a function.

cpp/zotaut/journal_timeliness_checker.cc
```diff
@@ -48,17 +48,20 @@ bool MaxDeliveredAtSmallerThanUpdateWindow(const time_t &max_delivered_at, const
...
+    upload_tracker->deleteOnlineFirstEntriesOlderThan(StringUtil::ToUnsigned(zeder_id), zeder_instance, std::min(update_window / 2 , static_cast<unsigned>(14)));
 
-    const time_t max_delivered_at(upload_tracker->getLastUploadTime(StringUtil::ToUnsigned(zeder_id),
-                                  Zeder::ParseFlavour(zeder_instance)));
+    if (delivery_mode == ZoteroHarvester::Config::UploadOperation::LIVE) {
```
mtrojan-ub, line 58: As far as i can remember, one of the problems with the journal_timeliness_checker has been that both nu and ub28 process all journals. So e.g. nu might complain about missing articles for live journals, even if it is not his job to process live journals at all. Can we solve that somehow? And how do we test it? (maybe handle the delivery_mode depending on server environment or CLI parameter?)

=> might also be handled in a follow-up PR

cpp/zotaut/journal_timeliness_checker.cc
jriedl: Newly synced are skipped - the `false` is the default value to return. I left the determination of the delivery at its previous place to reduce merge conflicts with the existing selective evaluation code if it has to be reintegrated

cpp/zotaut/journal_timeliness_checker.cc
jriedl: I thought about integrating some kind of host name determination and handling but for the time being removing overdue OF-articles seemed the decisive concern right now. Behaviour of the tardy journals should be identical previous versions

PR #3823, opened 2021-03-17
AppArmorUtil: SetProfileMode needs full path => Local profiles

mtrojan-ub:


cpp/lib/include/AppArmorUtil.h
```diff
@@ -40,11 +40,11 @@ std::vector<std::pair<std::string, ProfileMode>> GetProfiles();
 /** \brief  Return the mode for the given profile. */
 ProfileMode GetProfileMode(const std::string &profile_id);
 
-/** \brief  Install the profile from the given path. Filename will later be used as profile_id. */
-void InstallProfile(const std::string &profile_path);
+/** \brief  Install the profile from the given path as local profile. Filename will later be used as profile_id. */
+void InstallLocalProfile(const std::string &profile_path);
```
ruschein, line 44: "as a local profile".  Should "Filename" be "profile_path"?

cpp/lib/src/AppArmorUtil.cc
```diff
@@ -75,9 +75,11 @@ ProfileMode GetProfileMode(const std::string &profile_id) {
...
 
-void InstallProfile(const std::string &profile_path) {
-    static const std::string PROFILES_DIR("/etc/apparmor.d/local/");
-    const std::string target_path(PROFILES_DIR + FileUtil::GetBasename(profile_path));
+static const std::string PROFILES_DIR("/etc/apparmor.d/local");
+
+
+void InstallLocalProfile(const std::string &profile_path) {
+    const std::string target_path(PROFILES_DIR + "/" + FileUtil::GetBasename(profile_path));
     FileUtil::CopyOrDie(profile_path, target_path);
```
ruschein, line 83: Is it safe to always overwrite the installed profile, if there is one?  Should we add an additional boolean function parameter called "overwrite" set to true by default?

ruschein (changes requested):

cpp/lib/src/AppArmorUtil.cc
mtrojan-ub: As far as i know, this is safe and the new profile will just replace the old one. CopyOrDie will just replace the file without error, and AppArmor will use the new settings. Since CopyOrDie does not support and overwrite parameter, i guess it's not necessary to introduce an overwrite parameter here either.

cpp/lib/include/AppArmorUtil.h
```diff
@@ -40,11 +40,13 @@ std::vector<std::pair<std::string, ProfileMode>> GetProfiles();
 /** \brief  Return the mode for the given profile. */
 ProfileMode GetProfileMode(const std::string &profile_id);
 
-/** \brief  Install the profile from the given path. Filename will later be used as profile_id. */
-void InstallProfile(const std::string &profile_path);
+/** \brief  Install the profile from the given path as a local profile.
+ *          The filename of the profile_path will later be used as the profile_id.
```
ruschein, line 44: Shouldn't that read "as the profile ID"?

ruschein: Also after looking at the implementation, I would shy away from using "path" as that implies a path.  How about profile_config_filename or something similar?

ruschein (changes requested):

cpp/lib/include/AppArmorUtil.h
mtrojan-ub: Done... its just not marked as "outdated" because line+1 was marked with the comment.

cpp/lib/include/AppArmorUtil.h
mtrojan-ub: Everytime i read a variable called "filename", i assume this means only the name of the file without the full path, which is mandatory in this case.

PR #3829, opened 2021-03-24
Adjust Lobid-Handling

jriedl:


cpp/lib/include/LobidUtil.h
```diff
@@ -33,6 +33,8 @@ namespace LobidUtil {
  *         - http://lobid.org/resources/api
  */
 std::string GetAuthorGNDNumber(const std::string &author, const std::string &additional_query_params="");
+std::string GetAuthorLastFirstGNDNumber(const std::string &author_surname, const std::string &author_firstname,
```
ruschein, line 36: I think a better name would be GetAuthorLastFirstAndGNDNumber

cpp/lib/include/LobidUtil.h
```diff
@@ -33,6 +33,8 @@ namespace LobidUtil {
...
+                               const std::string &additional_query_params="");
```
ruschein, line 37: Missing spaces.

ruschein (changes requested):

cpp/lib/include/LobidUtil.h
jriedl: The Function only returns a GNDNumber but unlike the original one uses separate first und last name in the argument list.

ruschein: Ok, then how about GetGNDNumberFromAuthorName?

jriedl: This would break the analogy to `GetAuthorGNDNumber` above or this would have to be renamed as well and the code base adjusted.

ruschein: I am for keeping that name because I don't think it is confusing at all.  The new name strikes me as confusing though.

cpp/lib/include/LobidUtil.h
ruschein: Missing spaces about the equals operator.

cpp/lib/include/LobidUtil.h
jriedl: How about returning to my original approach of overloading `GetAuthorGNDNumber` and removing the default parameter in the third argument to make the call unique?

ruschein: How about just overloading GetAuthorGNDNumber?

jriedl: That's the suggestion and was the original approach. However, one would have to do away with the default parameter `additional_query_params=""` because otherwise the compiler cannot properly disambiguate the first two `std::string` arguments.

cpp/lib/include/LobidUtil.h
```diff
@@ -33,6 +33,8 @@ namespace LobidUtil {
  *         - http://lobid.org/resources/api
  */
 std::string GetAuthorGNDNumber(const std::string &author, const std::string &additional_query_params="");
+std::string GetAuthorGNDNumber(const std::string &author_surname, const std::string &author_firstname,
+                                        const std::string &additional_query_params);
```
ruschein, line 37: Indentation looks wrong.

cpp/lib/include/LobidUtil.h
```diff
@@ -33,6 +33,8 @@ namespace LobidUtil {
  *         - http://lobid.org/resources/api
  */
 std::string GetAuthorGNDNumber(const std::string &author, const std::string &additional_query_params="");
```
ruschein, line 35: Please add spaces here around the equal sign.

cpp/lib/src/MiscUtil.cc
```diff
@@ -293,7 +293,7 @@ bool NormaliseDOI(const std::string &doi_candidate, std::string * const normalis
 }
 
 
-const ThreadSafeRegexMatcher CONTAINS_DOI_MATCHER("(10[.][0-9]{4,}(?:[.][0-9]+)*/(?:(?![\"&\\'<>])\\S)+)");
+const ThreadSafeRegexMatcher CONTAINS_DOI_MATCHER("(10.\\d{4,9}/[-._;()/:A-Z0-9]+)");
```
ruschein, line 296: We only use all caps for scalar constants.

cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -993,7 +995,7 @@ void GenerateMarcRecordFromMetadataRecord(const MetadataRecord &metadata_record,
             subfields.appendSubfield('c', creator.title_);
```
ruschein, line 995: Indentation is not consistent with the following line.

cronjobs/zotero_harvester_util.py
```diff
@@ -0,0 +1,17 @@
+#!/bin/python3
+# -*- coding: utf-8 -*-
+import configparser
+
```
ruschein, line 4: Missing blank line.

cronjobs/zotero_harvester_util.py
```diff
@@ -0,0 +1,17 @@
...
+def GetZoteroConfiguration():
+    zotero_harvester_conf_path = "/usr/local/var/lib/tuelib/zotero-enhancement-maps/zotero_harvester.conf"
+    zotero_harvester_conf_tmp_path = "/tmp/zotero_harvester.conf"
+    # Section at the beginning needed for configparser
+    with open(zotero_harvester_conf_path, "r") as zotero_harvester_conf, \
+         open(zotero_harvester_conf_tmp_path, "w") as zotero_harvester_tmp_conf:
+            zotero_harvester_tmp_conf.write("[DEFAULT]\n")
+            zotero_harvester_tmp_conf.write(zotero_harvester_conf.read())
+            zotero_harvester_tmp_conf.close()
```
ruschein, line 13: The close is done implicitly as part of the "with open" construct.

ruschein (changes requested):

cpp/lib/src/MiscUtil.cc
jriedl: c.f. definition of DOI matcher in the same file.

cpp/lib/src/ZoteroHarvesterConversion.cc
jriedl: Looks correct in editor

PR #3830, opened 2021-03-24
Ruschein

ruschein:


cpp/lib/include/StringUtil.h
```diff
@@ -854,6 +854,13 @@ inline std::string BackslashEscape(const std::string &chars_to_escape, const std
 std::string ExtractHead(std::string * const target, const std::string &delimiter_string = " ", std::string::size_type start = 0);
 
 
+// \note If the delimiter is the first character in "s", an empty entry will be the first entry in the
+//       returned vector.  If the delimiter is the last entry in "s", no entry will be created at the end
```
mtrojan-ub, line 858: Why on earth would we want a common library function with such an inconsistent non-configurable standard behaviour?

mtrojan-ub: Please pardon my french... i just suggest that it would be better to either trim both sides or trim none. If more flexibility is needed, there could be either some kind of enum like e.g. TrimBehaviour or the caller just strips leading/trailing delimiters as needed manually before calling the function.

mtrojan-ub:
I'm a bit worried about the following scenario:

The email_watcher is live, and from one day to another, new messages will no longer be encoded in any yet supported transfer encoding. 

Can that happen? What are we going to do then?

ruschein:
> I'm a bit worried about the following scenario:
> 
> The email_watcher is live, and from one day to another, new messages will no longer be encoded in any yet supported transfer encoding.
> 
> Can that happen? What are we going to do then?

I suspect that we don't get matches then and will get email notifications which will alert us to the need to adjust the code.

cpp/lib/include/StringUtil.h
ruschein: I agree that this is tricky and it is not obvious to me what is best here. BUT, the typical delimiter is a newline so that means that we're splitting text.  If newlines are used to terminate lines, which is the typical use case, then the behaviour as implemented would be correct.  Please consider the following example:

\n
fred\n
bob\n

This would return 3 entries, the first of which would be empty.

ruschein: Moved into email_watcher.cc and renamed to SplitIntoLines (also dropped the delimiter parameter).

cpp/lib/include/MBox.h
```diff
@@ -29,7 +31,26 @@ class MBox {
...
+    class BodyPart {
+        std::vector<std::pair<std::string, std::string>> mime_headers_;
+        std::string body_;
+    public:
+        BodyPart() = default;
+        BodyPart(const BodyPart &other) = default;
+        BodyPart(const std::vector<std::pair<std::string, std::string>> &mime_headers, const std::string &body)
+            : mime_headers_(mime_headers), body_(body) { }
+
+        /** \note Ther keys are all lowercase. */
```
mtrojan-ub, line 44: Typo `Ther`

PR #3835, opened 2021-03-29
Riedl

jriedl:


mtrojan-ub:
This PR looks tiny, but since the existing functions are that generic, thats exactly how it should work! Great!

PR #3839, opened 2021-03-31
boot_notification.sh: Remove check if system has been started up (this is part of the unit file now)

mtrojan-ub:


mtrojan-ub:
e.g. if the system is started up but /usr/bin/systemctl is-system-running detects a failed task, it will return a status code != 0, aborting the script and preventing the mail from being sent

PR #3841, opened 2021-03-31
Update not only the SQL database but also text files under FID-Entwic…

ruschein:
…klung as requested by Martin.

### DIFF
--- a/cpp/pipelines/collect_journal_stats.cc
+++ b/cpp/pipelines/collect_journal_stats.cc
@@ -1,7 +1,7 @@
 /** \brief Updates Zeder (via Ingo's SQL database) w/ the last N issues of harvested articles for each journal.
  *  \author Dr. Johannes Ruscheinski (johannes.ruscheinski@uni-tuebingen.de)
  *
- *  \copyright 2018-2020 Universitätsbibliothek Tübingen.  All rights reserved.
+ *  \copyright 2018-2021 Universitätsbibliothek Tübingen.  All rights reserved.
  *
  *  This program is free software: you can redistribute it and/or modify
  *  it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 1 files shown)


cpp/pipelines/collect_journal_stats.cc
```diff
@@ -263,6 +264,31 @@ void UpdateDatabase(const IniFile &ini_file, const std::string &system_type, con
 }
 
 
+const std::string TEXT_FILE_DIRECTORY("/mnt/ZE020150/FID-Entwicklung/Zeder_Supervision");
+
+
+void UpdateTextFiles(const std::unordered_map<std::string, ZederIdAndPPNType> &ppns_to_zeder_ids_and_types_map,
+                     const std::unordered_map<std::string, DbEntry> &ppns_to_most_recent_entries_map)
+{
+    const auto DIRECTORY_PREFIX(TEXT_FILE_DIRECTORY + "/" + DnsUtil::GetHostname() + "/");
```
jriedl, line 273: I will merge it, but seems a bit dangerous, since `::gethostname()` used in `DnsUtil::GetHostname()` might return a wide variety of hostname versions depending on the configuration of the host system, so it might make sense to avoid the automatic creation of new directories and use a predefined manual setup instead.

cpp/pipelines/collect_journal_stats.cc
```diff
@@ -263,6 +264,31 @@ void UpdateDatabase(const IniFile &ini_file, const std::string &system_type, con
...
+        const auto ppns_and_zeder_id_and_type(ppns_to_zeder_ids_and_types_map.find(ppn));
+        if (unlikely(ppns_and_zeder_id_and_type == ppns_to_zeder_ids_and_types_map.cend()))
+            LOG_ERROR("Map lookup failed for \"" + ppn + "\"!");
+        const auto output(FileUtil::OpenForAppendingOrDie(DIRECTORY_PREFIX
+                                                          + std::to_string(ppns_and_zeder_id_and_type->second.zeder_id_)
+                                                          + ".txt"));
+        (*output) << db_entry.jahr_ << ',' << db_entry.band_ << ',' << db_entry.heft_ << db_entry.seitenbereich_ << '\n';
+    }
+
+    LOG_INFO("Updated " + std::to_string(updated_files.size()) + " file(s) under " + DIRECTORY_PREFIX + ".");
```
jriedl, line 288: `updated_files` seems unused

ruschein: Fixed

PR #3842, opened 2021-04-01
Ruschein2

ruschein:


cpp/throwaways/find_reference_subfields.cc
```diff
@@ -0,0 +1,72 @@
...
+
+void ListCrossLinkFields(MARC::Reader * const reader) {
+    std::set<std::string> cross_link_fields;
+    while (const auto record = reader->read()) {
+        for (const auto &field : record) {
+            if (field.isControlField() or not field.hasSubfield('w'))
+                continue;
+
+            const auto subfields(field.getSubfields());
+            for (const auto &subfield : subfields) {
```
jriedl, line 47: Apparently you are using all `$w`. `CROSS_LINK_FIELD_TAGS` defined above should be used at some place?

jriedl: Or, if used to detect _new_ fields why not just the `(DE-627)`-filter?

PR #3845, opened 2021-04-06
Don't abort if the Zeder database can't be contacted.

ruschein:


cpp/pipelines/collect_journal_stats.cc
```diff
@@ -53,12 +55,18 @@ struct ZederIdAndPPNType {
 
 
 std::unordered_map<std::string, ZederIdAndPPNType> GetPPNsToZederIdsAndTypesMap(const std::string &system_type) {
+    std::unordered_map<std::string, ZederIdAndPPNType> ppns_to_zeder_ids_and_types_map;
+
     const Zeder::SimpleZeder zeder(system_type == "ixtheo" ? Zeder::IXTHEO : Zeder::KRIMDOK, { "eppn", "pppn" });
-    if (unlikely(zeder.empty()))
-        LOG_ERROR("found no Zeder entries matching any of our requested columns!"
-                  " (This *should* not happen as we included the column ID!)");
+    if (unlikely(zeder.empty())) {
```
mtrojan-ub, line 61: Returning false in FullDumpDownloader::downloadData is the correct thing to do! However, this will lead to Zeder::SimpleZeder causing a LOG_ERROR, so you will not be able to check for zeder.empty() at this point anymore.

PR #3846, opened 2021-04-06
Triggers FTW! :-)

ruschein:


cpp/mysql_list_tables.cc
```diff
@@ -75,6 +84,8 @@ int Main(int argc, char *argv[]) {
 
     ProcessTablesOrViews(db_connection.get(), /* process_tables = */true);
     ProcessTablesOrViews(db_connection.get(), /* process_tables = */false);
+    ProcessTriggers(db_connection.get(), "vufind");
```
mtrojan-ub, line 87: database names should not be hardcoded

cpp/mysql_list_tables.cc
```diff
@@ -50,6 +50,15 @@ void ProcessTablesOrViews(DbConnection * const db_connection, const bool process
 }
 
 
+void ProcessTriggers(DbConnection * const db_connection, const std::string &database_name) {
+    db_connection->queryOrDie("SHOW TRIGGERS FROM `" + database_name + "`");
+    DbResultSet result_set(db_connection->getLastResultSet());
+    while (const auto row = result_set.getNextRow())
+        std::cout << "CREATE TRIGGER `" << row["Trigger"] << "` ON `" << database_name << '.' << row["Table"]
```
mtrojan-ub, line 57: This does not look like the official CREATE TRIGGER syntax:
https://dev.mysql.com/doc/refman/8.0/en/create-trigger.html

- How about using the original namings for the parts? e.g. `body` instead of `statement`
- `trigger_time` and `trigger_event` must occur before the `ON` keyword
- what about `trigger_order`?
- `` `database_name.table_name` `` could be invalid, it should be `` `database_name`.`table_name` `` instead

cpp/mysql_diff_schemas.cc
```diff
@@ -72,16 +115,27 @@ void LoadSchema(const std::string &filename,
...
             current_schema.emplace_back(line);
         }
     }
+
     if (not current_schema.empty()) {
         std::sort(current_schema.begin(), current_schema.end(), SchemaLineIsLessThan);
         (*table_or_view_name_to_schema_map)[current_table_or_view] = current_schema;
     }
+
+    std::sort(triggers->begin(), triggers->end(), TriggerLineIsLessThan);
```
mtrojan-ub, line 138: the std::string that contains a trigger might contain multiple lines, so maybe we should rename this to `TriggerStatementIsLessThan`

cpp/mysql_diff_schemas.cc
```diff
@@ -53,16 +53,59 @@ inline bool SchemaLineIsLessThan(const std::string &line1, const std::string &li
...
+        return "";
+    ++cp; // Skip over the closing backtick.
+
+    return quoted_text;
+}
+
+
+void ExtractTriggerNameAndTable(const std::string &trigger, std::string * const name, std::string * const table) {
+    auto cp(trigger.cbegin());
+    *name = ExtractNextBacktickQuotedText(trigger.cend(), cp);
```
mtrojan-ub, line 77: I don't think it is as simple as that => optional DEFINER statement could use backticks as well

cpp/mysql_list_tables.cc
ruschein: I think it is dangerous to include an uncontrolled list of databases here because we may want to install different software on different servers and some of that software may have its own database tables with triggers etc.  Also, we currently only do table versioning and comparisons on ub_tools.

cpp/mysql_list_tables.cc
ruschein: I am not sure, what you mean here.  The "names" are what mysql is using for the columns.  I can't use anything else.

cpp/mysql_diff_schemas.cc
ruschein: No because we generate it ourselves with mysql_list_tables.

cpp/mysql_diff_schemas.cc
```diff
@@ -53,16 +53,59 @@ inline bool SchemaLineIsLessThan(const std::string &line1, const std::string &li
...
+}
+
+
+inline bool TriggerLineIsLessThan(const std::string &line1, const std::string &line2) {
+    std::string name1, table1;
+    ExtractTriggerNameAndTable(line1, &name1, &table1);
+    std::string name2, table2;
+    ExtractTriggerNameAndTable(line2, &name2, &table2);
+
+    return name1 + table1 < name2 + table2;
```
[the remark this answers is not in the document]
ruschein: It doesn't matter.  All that matters is that we don't miss any matches when we compare our two schemas.

cpp/mysql_list_tables.cc
ruschein: Not hardcoded any longer.

cpp/mysql_diff_schemas.cc
ruschein: In our input the DEFINER part of the statement is now always present and will be skipped over,

cpp/pipelines/collect_journal_stats.cc
```diff
@@ -59,7 +59,7 @@ std::unordered_map<std::string, ZederIdAndPPNType> GetPPNsToZederIdsAndTypesMap(
 
     const Zeder::SimpleZeder zeder(system_type == "ixtheo" ? Zeder::IXTHEO : Zeder::KRIMDOK, { "eppn", "pppn" });
     if (not zeder) {
-        EmailSender::SimplerSendEmail("no-reply", { system_type + "team@ub.uni-tuebingen.de" },
+        EmailSender::SimplerSendEmail("no-reply@ub.uni-tuebingen.de", { system_type + "team@ub.uni-tuebingen.de" },
```
mtrojan-ub, line 62: missing `-` between `system_type` and `team@...`

PR #3851, opened 2021-04-07
sql_updates/vufind.4: Change type of vufind.ixtheo_id_result_sets.id

mtrojan-ub:


cpp/data/sql_updates/vufind.4
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE vufind.ixtheo_id_result_sets DROP FOREIGN KEY ixtheo_id_result_sets_ibfk_1;
+ALTER TABLE vufind.ixtheo_id_result_sets CHANGE COLUMN id id BIGINT UNSIGNED NOT NULL;
```
andreas-ub, line 2: maybe the use of modify column is easier.

cpp/data/sql_updates/vufind.4
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE vufind.ixtheo_id_result_sets DROP FOREIGN KEY ixtheo_id_result_sets_ibfk_1;
```
mtrojan-ub, line 1: ixtheo tables will only exist on ixtheo machines, so we need some kind of IF EXISTS logic here

PR #3852, opened 2021-04-08
Riedl

jriedl:


cpp/lib/include/ZoteroHarvesterConfig.h
```diff
@@ -67,6 +67,8 @@ struct ZoteroMetadataParams {
     std::map<std::string, std::unique_ptr<ThreadSafeRegexMatcher>> fields_to_suppress_;
     std::map<std::string, std::string> fields_to_override_;
     std::map<std::string, std::unique_ptr<ThreadSafeRegexMatcher>> exclusion_filters_;
+//    std::map<std::string, std::unique_ptr<ThreadSafeRegexMatcher>> rewrite_filters_;
```
ruschein, line 70: Please do not commit commented-out code.

cpp/lib/include/RegexMatcher.h
```diff
@@ -83,6 +83,7 @@ class ThreadSafeRegexMatcher {
     MatchResult match(const std::string &subject, const size_t subject_start_offset = 0,
                       size_t * const start_pos = nullptr, size_t * const end_pos = nullptr) const;
     std::string replaceAll(const std::string &subject, const std::string &replacement) const;
+    std::string replaceWithBackreferences(const std::string &subject, const std::string &replacement, const bool global = false);
```
ruschein, line 86: I think it might be a good idea to add an interface documentation here, including a simple example usage.

cpp/lib/src/RegexMatcher.cc
```diff
@@ -172,6 +172,65 @@ std::string ThreadSafeRegexMatcher::replaceAll(const std::string &subject, const
 }
 
 
+static std::string InsertReplacement(const ThreadSafeRegexMatcher::MatchResult &result, const std::string &replacement_pattern) {
```
ruschein, line 175: Why don't you reuse the already existing implementation?

cpp/lib/src/RegexMatcher.cc
```diff
@@ -172,6 +172,65 @@ std::string ThreadSafeRegexMatcher::replaceAll(const std::string &subject, const
...
+        if (not global)
+            break;
+    }
+
+    while (subject_start_offset < subject.length())
+        replaced_string += subject[subject_start_offset++];
+
+    return replaced_string;
+}
+
```
ruschein, line 229: Why so many blank lines?

ruschein (changes requested):

cpp/lib/src/RegexMatcher.cc
jriedl: As discussed `ThreadSafeRegexMatcher` returns a  `MatchResult` which forces a different handling.

cpp/lib/src/RegexMatcher.cc
```diff
@@ -172,6 +172,62 @@ std::string ThreadSafeRegexMatcher::replaceAll(const std::string &subject, const
 }
 
 
+static std::string InsertReplacement(const ThreadSafeRegexMatcher::MatchResult &result, const std::string &replacement_pattern) {
```
ruschein, line 175: It appears that this function, apart from the name on line 187 and the type of the first parameter is identical to the other InsertReplacement function in this file.  This calls for replacing both functions with a function template.  Just call the type of the first argument something like MatchedGroups and the name of the argument matched_groups and you should be able to get by with a single template function.

PR #3855, opened 2021-04-12
Use erase-remove idiom to remove fields

jriedl:


cpp/lib/src/MARC.cc
```diff
@@ -863,6 +863,26 @@ Record::iterator Record::erase(const Tag &tag, const bool first_occurrence_only)
 }
 
 
+bool Record::deleteFieldWithSubfieldCodeMatching(const Tag &tag, const char subfield_code, const ThreadSafeRegexMatcher &matcher) {
+    bool matched(false);
+    fields_.erase(std::remove_if(fields_.begin(), fields_.end(),
+                  [&](const Field &field) -> bool
+                  { if ((field.getTag() != tag) or not field.hasSubfield(subfield_code))
```
mtrojan-ub, line 870: indentation inside this block looks a bit strange, maybe it would be better to put a newline between `{ if` and indent the rest as block with the regular 4 whitespaces (until the `return false;` statenent)

PR #3866, opened 2021-04-21
Ruschein

ruschein:


cpp/marc_tools/marc_filter.cc
```diff
@@ -627,16 +629,28 @@ bool ReplaceSubfields(const std::vector<std::string> &subfield_specs, RegexMatch
...
+                tags_of_deleted_fields.insert(field->getTag().toString());
+            }
         }
     }
 
+    // Did we generate completely empty fields?
+    if (unlikely(not indices_of_deleted_fields.empty())) {
+        LOG_WARNING("regex \"" + matcher.getPattern() + "\" led to empty fields in the record w/ control number "
+                    + record->getControlNumber() + " and field(s) " + StringUtil::Join(tags_of_deleted_fields, ',') + "!");
+        record->deleteFields(indices_of_deleted_fields);
```
jriedl, line 645: Just a a note: An operation on subfields might lead to invalidating an iterator on the fields of the record

PR #3871, opened 2021-04-22
Fixed another bug which led to old data persisting.

ruschein:


cpp/pipelines/collect_journal_stats.cc
```diff
@@ -298,9 +299,17 @@ void UpdateTextFiles(const bool debug, const std::unordered_map<std::string, Zed
...
+        File *output;
+        if (updated_files.find(filename) != updated_files.end())
+            output = FileUtil::OpenForAppendingOrDie(filename).get();
+        else { // It's the first time we're opeing this file so we want to overwrite old contents should they exist.
+            output = FileUtil::OpenOutputFileOrDie(filename).get();
+            updated_files.emplace(filename);
+        }
+
         (*output) << db_entry.jahr_ << ',' << db_entry.band_ << ',' << db_entry.heft_ << ',' << db_entry.seitenbereich_ << '\n';
+        delete output;
```
jriedl, line 312: `delete` seems strange without explicit `new` - `FileUtil` returns `unique_ptr` in both cases. Why not  `std::unique_ptr<File> output`?

PR #3874, opened 2021-04-23
New pattern replacement implementation.

ruschein:


cpp/rss_aggregator.cc
```diff
@@ -163,20 +163,49 @@ bool ProcessRSSItem(const std::string &feed_id, const SyndicationFormat::Item &i
...
+                replacement += ch;
+        } else if (ch == ':')
+            in_pattern = false;
+        else if (ch == ';') {
+            const auto regex_matcher(RegexMatcher::RegexMatcherFactoryOrDie(pattern));
+            item->setDescription(regex_matcher->replaceAll(item->getDescription(), replacement));
+            delete regex_matcher;
+            pattern.clear(), replacement.clear();
+            in_pattern = true;
+        } if (ch == '\\')
```
andreas-ub, line 187: if should be in a new line?

PR #3875, opened 2021-04-23
Added a new ReplaceAll utility function.

ruschein:


cpp/lib/src/RegexMatcher.cc
```diff
@@ -413,6 +413,19 @@ bool RegexMatcher::Matched(const std::string &regex, const std::string &subject,
 }
 
 
+std::string RegexMatcher::ReplaceAll(const std::string &regex, const std::string &subject, const std::string &replacement,
+                                     const unsigned options)
+{
+    std::string err_msg;
+    auto matcher(RegexMatcherFactory(regex, &err_msg, options));
+    if (matcher == nullptr)
+        LOG_ERROR("failed to compile \"" + regex + "\": " + err_msg);
```
andreas-ub, line 422: maybe hard reaction, alternatively one could return the original string

PR #3882, opened 2021-04-29
added ELC content for subscription package records

andreas-ub:


cpp/generate_subscription_packets.cc
```diff
@@ -108,6 +108,8 @@ void GenerateBundleDefinition(const Zeder::SimpleZeder &zeder, const std::string
 {
     unsigned included_journal_count(0);
     std::set<std::string> bundle_ppns; // We use a std::set because it is automatically being sorted for us.
+    bool contains_print = false;
```
ruschein, line 111: Please use parentheses for variable initialisations.

cpp/pipelines/add_subscription_bundle_entries.cc
```diff
@@ -65,6 +65,10 @@ MARC::Record GenerateBundleRecord(const std::string &record_id, const std::strin
         record.insertField("BIB", { { 'a', "1" } });
     if (include_churchlaw)
         record.insertField("CAN", { { 'a', "1" } });
+    if (media_type.compare("online_and_print") == 0 or media_type.compare("online") == 0)
```
ruschein, line 68: Please just use the == operator and not the std::string compare member function.  It is more readable and we don't use compare anywhere else.

cpp/pipelines/add_subscription_bundle_entries.cc
```diff
@@ -65,6 +65,10 @@ MARC::Record GenerateBundleRecord(const std::string &record_id, const std::strin
...
+        record.insertField("ELC", 'a', "1");
+    if (media_type.compare("online_and_print") == 0 or media_type.compare("print") == 0)
+        record.insertField("ELC", 'b', "1");
```
ruschein, line 71: Please generate just a single ELC field.

ruschein (changes requested):

PR #3916, opened 2021-05-19
ControlNumberGuesser: Restored path to sql template file

mtrojan-ub:


cpp/lib/src/ControlNumberGuesser.cc
```diff
@@ -33,7 +33,7 @@
 
 const std::set<std::string> ControlNumberGuesser::EMPTY_SET;
 const std::string ControlNumberGuesser::DATABASE_PATH(UBTools::GetTuelibPath() + "control_number_guesser.sq3");
-const std::string ControlNumberGuesser::INSTALLER_SCRIPT_PATH("/usr/local/ub_tools/cpp/data/installer/control_number_guesser.sq3");
+const std::string ControlNumberGuesser::INSTALLER_SCRIPT_PATH("/usr/local/ub_tools/cpp/data/installer/control_number_guesser.sql");
```
ruschein, line 36: This look wrong to me.  We used the extension sq3 to distinguish sqlite3 from MySQL/MariaDB.

PR #3919, opened 2021-05-20
remove temp. files in failed branch

andreas-ub:


cpp/marc_tools/marc_grep_multiple.sh
```diff
@@ -22,25 +22,30 @@ for filename in "$@"; do
             found=0
 	fi
     else
-	tar_filename=${filename%.gz}
+    filebasename="/tmp/$(basename ${filename})"
+	tar_foldername=${filebasename%.gz}
```
mtrojan-ub, line 26: strange indentation in this file, tabs and spaces seem to be mixed. please normalize!

cpp/marc_tools/marc_grep_multiple.sh
```diff
@@ -22,25 +22,30 @@ for filename in "$@"; do
             found=0
 	fi
     else
-	tar_filename=${filename%.gz}
+    filebasename="/tmp/$(basename ${filename})"
```
mtrojan-ub, line 25: Variable names should be improved. This variable is called filebasename, but contains a full path. Also, there should be some kind of uuid so the script does not try to re-use an existing directory in case something goes wrong (you can e.g. use the uuidgen command).

cpp/marc_tools/marc_grep_multiple.sh
```diff
@@ -11,25 +11,25 @@ shift
 found=1
 for filename in "$@"; do
     if [[ ! $filename =~ \.tar\.gz$ ]]; then
-	marc_grep_output=$(marc_grep "$filename" "$marc_grep_conditional_expression" 3>&2 2>&1 1>&3)
+    marc_grep_output=$(marc_grep "$filename" "$marc_grep_conditional_expression" 3>&2 2>&1 1>&3)
```
mtrojan-ub, line 14: this line should be indented 1 level to the right

cpp/marc_tools/marc_grep_multiple.sh
```diff
@@ -11,25 +11,25 @@ shift
...
         if [[ $? -ne 0 ]]; then
             echo "marc_grep failed ($marc_grep_output)!"
             exit 1
         fi
         last_line=$(echo "$marc_grep_output" | tail -1)
         if [[ ! $last_line =~ ^Matched\ 0 && $last_line =~ ^Matched ]]; then
-	    echo "was found in $filename"
+        echo "was found in $filename"
```
mtrojan-ub, line 21: this one too

cpp/marc_tools/marc_grep_multiple.sh
```diff
@@ -11,25 +11,25 @@ shift
...
             found=0
-	fi
+    fi
     else
     tmppath=`mktemp --directory`
```
mtrojan-ub, line 25: there is still something wrong with the indentation of this whole block

PR #3936, opened 2021-06-07
first version of pipeline order algorithm

andreas-ub:


cronjobs/create_pipeline_execution.py
```diff
@@ -0,0 +1,123 @@
+#!/bin/python3
+# -*- coding: utf-8 -*-
+
+# Algorithm to bring pipeline phases in an ideal order
+
+def split_to_equal_parts(lst, n):
+        for x in range(0, len(lst), n):
```
mtrojan-ub, line 7: indentation

cronjobs/create_pipeline_execution.py
```diff
@@ -0,0 +1,123 @@
...
+def Main():
+
+    # Number of parallel FIFO groups
+    number_of_groups = 2 # = 1 without parallelization
+    # Number of FIFO phases in a FIFO group
+    items_in_group = 3
+    phases = ["3","4","5","6","14","7","8","9","13"]
+    # Dependency list has to be built dynamically later
+    # e.g. ["5","3"] means that "5" can only be run after "3"
+    deps = [["5","3"],["7","4"],["8","5"],["9","7"],["6","13"],["3","13"],["b","a"],["c","a"],["m","k"],["z","x"],["z","k"],["c","b"]]
```
mtrojan-ub, line 35: Sometimes phases have multiple dependencies. For example, have a look at ixtheo_marc_pipeline.sh:

```
StartPhase "Add Keyword Synonyms from Authority Data"
(add_synonyms \
    GesamtTiteldaten-post-phase"$((PHASE-1))"-"${date}".mrc \
    Normdaten-partially-augmented2-"${date}".mrc \
    GesamtTiteldaten-post-phase"$PHASE"-"${date}".mrc >> "${log}" 2>&1 && \
EndPhase || Abort) &
wait
```
Maybe it would be better to use a more complex datatype, here's an example of a hybrid list (there might be better options!):
`deps = [["add_synonyms", ["map_ddc_to_ixtheo_notations", "add_authority_beacon_information"]]]`

cronjobs/create_pipeline_execution.py
```diff
@@ -0,0 +1,123 @@
...
+            each_chunk = lst[x: n+x]
+            if len(each_chunk) < n:
+                each_chunk = each_chunk + [None for y in range(n-len(each_chunk))]
+            yield each_chunk
+
+def contains_elem(lst_2d, elem):
+    result = False
+    for lst in lst_2d:
+        if elem in lst:
+            result = True
```
mtrojan-ub, line 17: Can be made shorter, just return True directly here and return False in the last line (no need for a result variable here, also => shorter execution)

cronjobs/create_pipeline_execution.py
```diff
@@ -0,0 +1,230 @@
...
+        self.title_norm = _title_norm
+        self.name = _name
+
+# Dependency list has to be built dynamically later
+# Preprocessing
+# Check Record Integrity at the Beginning of the Pipeline
+# marc_check --do-not-abort-on-empty-subfields --do-not-abort-on-invalid-repeated-fields \
+#            --write-data=GesamtTiteldaten-post-phase"$PHASE"-"${date}".mrc GesamtTiteldaten-"${date}".mrc >> "${log}" 2>&1
+
+# convention in dict-keys:
```
mtrojan-ub, line 27: Please move this documentation to the Phase property class, also add some documentation to the other properties (e.g. the meaning of "t", "x" and "n" for title_norm (maybe there might be a better name for title_norm as well?)

cronjobs/create_pipeline_execution.py
```diff
@@ -0,0 +1,230 @@
...
+                process_chunk(open_fifo_title)
+                open_fifo_title = []
+            elif phase_title_norm.startswith("n"):
+                process_chunk(open_fifo_norm)
+                open_fifo_title = []
+
+
+def Main():
+    parser = argparse.ArgumentParser(description='bring pipeline phases in an ideal order')
+    parser.add_argument("--run_steps", default=3, type=int, help="(max.) number of elements inside a running group")
```
mtrojan-ub, line 189: max_group_size might be better (or max-group-size to be more unix-like)

cronjobs/create_pipeline_execution.py
```diff
@@ -0,0 +1,230 @@
...
+#            --write-data=GesamtTiteldaten-post-phase"$PHASE"-"${date}".mrc GesamtTiteldaten-"${date}".mrc >> "${log}" 2>&1
+
+# convention in dict-keys: 
+#   leading "_" can only run after predecessor has finished
+#   trailing "_" successor must not run before job has finished
+#   leading and trailing "_" has to run independently
+
+phase_counter = 0
+
+phases = {
```
mtrojan-ub, line 34: If you have no keys, you should use a list [] rather than a dictionary {}

mtrojan-ub: or, you add the key outside the phase object, then you do not need the get_phase_by_key function anymore

cronjobs/create_pipeline_execution.py
```diff
@@ -0,0 +1,230 @@
...
+    global phase_counter
+    if len(lst) > 0:
+        for i, elem in enumerate(lst):
+            phase_counter += 1
+            if len(lst) > 1 and i != len(lst) - 1:
+                print("making fifo _ ", elem.title_norm, " phase counter: ", phase_counter)
+            print("processing ", elem.key, " -> ", elem.title_norm, " phase counter: ", phase_counter)
+        print("---")
+
+def split_to_equal_parts(lst, n):
```
mtrojan-ub, line 142: This function combines splitting of the function with executing the phases, so you should chose a different function name.

mtrojan-ub: And also some more comments inside the function would be helpful.

cronjobs/create_pipeline_execution.py
```diff
@@ -0,0 +1,231 @@
...
+                open_fifo_title = []
+
+
+def Main():
+    parser = argparse.ArgumentParser(description='bring pipeline phases in an ideal order')
+    parser.add_argument("--max-group-size", default=3, type=int, help="(max.) number of elements inside a running group")
+    args = parser.parse_args()
+
+    is_debug = False
+    max_items_in_group = args.max_group_size
```
mtrojan-ub, line 195: Variable name should stay the same, thats easier to read & more consistant

cronjobs/create_pipeline_execution.py
```diff
@@ -0,0 +1,231 @@
...
+def split_and_write_pipeline_steps(lst, n):
+    open_fifo_title = []
+    open_fifo_norm =[] 
+
+    for elemkey in lst:
+        elem = get_phase_by_key(elemkey)
+        phase_mode = elem.mode
+        phase_title_norm = elem.title_norm
+
+        if phase_title_norm.startswith("x"):
```
mtrojan-ub, line 153: When comparing phase_title_norm, you should use == instead of startswith. startswith suggests that the string may have multiple characters, which is allowed for phase_mode, but not for phase_title_norm.

cpp/mysql_database_patcher.cc
```diff
@@ -156,8 +156,16 @@ int Main(int argc, char *argv[]) {
...
     }
 
-    for (const auto &update_filename : update_filenames)
+    std::string last_schema;
+    for (const auto &update_filename : update_filenames) {
+        std::string database;
+        unsigned version;
+        SplitIntoDatabaseAndVersion(update_filename, &database, &version);
+        if (last_schema.empty() or database != last_schema)
+            db_connection.queryOrDie("use " + database);
```
mtrojan-ub, line 165: Please write sql statements like USE as uppercase

PR #3938, opened 2021-06-07
Riedl

jriedl:


cpp/lib/src/ZoteroHarvesterDownload.cc
```diff
@@ -434,18 +434,68 @@ Tasklet::Tasklet(ThreadUtil::ThreadSafeCounter<unsigned> * const instance_counte
 
 namespace ApiQuery {
 
+void SelectNonExistingEntriesFromZTSMultiple(const std::string &all, std::string * const filtered, const Util::UploadTracker &upload_tracker,
```
mtrojan-ub, line 437: there should be a better variable name than "all".

PR #3946, opened 2021-06-10
Finally

jriedl:


cpp/throwaways/generate_gnd_association_and_spelling_replacements.sh
```diff
@@ -0,0 +1,93 @@
...
+    echo "usage: $0 A-Z_output A_Z_bereinigt gnd_full.mrc result"
+    exit 1
+fi
+
+readonly AZ_OUTPUT=$1
+readonly AZ_CLEANED=$2
+readonly GND_FULL=$3
+readonly RESULT_FILE=$4
+readonly tmp_prefix=$(basename $0)
+#readonly gnd_matches="./matched3.txt"
```
mtrojan-ub, line 15: please remove these commented lines

cpp/throwaways/generate_gnd_association_and_spelling_replacements.sh
```diff
@@ -0,0 +1,93 @@
...
+#readonly gnd_matches="./tmp/matched_210609_1.txt"
+
+function MkTempFile {
+  echo $(mktemp /tmp/${tmp_prefix}.XXXXX)
```
mtrojan-ub, line 19: indentation

cpp/throwaways/generate_gnd_association_and_spelling_replacements.sh
```diff
@@ -0,0 +1,93 @@
...
+          for (i = 1; i < NF; i++) { 
+              printf(";%s", $i); 
+          } 
+          print ""; 
+     }' > ${az_output_reordered} 
+}
+
+
+function JoinAssociationAndSpelling {
+    readonly association_and_spelling=$(MkTempFile)
```
mtrojan-ub, line 48: This should be define outside the function, because it is also used in at least 1 other function later

cpp/throwaways/generate_gnd_association_and_spelling_replacements.sh
```diff
@@ -0,0 +1,93 @@
...
+#readonly gnd_matches="./matched3.txt"
+#readonly gnd_matches="./tmp/matched_210609_1.txt"
+
+function MkTempFile {
+  echo $(mktemp /tmp/${tmp_prefix}.XXXXX)
+}
+
+
+function GenerateKeywordToGNDMap {
+   readonly gnd_matched=$(MkTempFile)
```
mtrojan-ub, line 24: These TempFiles should also be defined outside this function, because they are re-used in other functions later

cpp/throwaways/generate_gnd_association_and_spelling_replacements.sh
```diff
@@ -0,0 +1,93 @@
+# Create krimdok local keyword file from the original output and make misspellings and potential GND references explicit
+#!/bin/bash
+
+
+if [ $# != 4 ]; then
+    echo "usage: $0 A-Z_output A_Z_bereinigt gnd_full.mrc result"
```
mtrojan-ub, line 6: Please do not mix german and english parameter names, and also use consistent prefixes.
If the first parameter is called `A-Z_output`, the second one should be renamed from `A_Z_bereinigt` to something like `A-Z_cleaned`.

jriedl: This is intentional because the original file names provided by K.D. are named like this and the script only works with these.

PR #3954, opened 2021-06-17
Extract frequencies

jriedl:
Not yet finished, so please don't merge

### DIFF
--- a/cpp/data/sql_updates/vufind.7
+++ b/cpp/data/sql_updates/vufind.7
@@ -1,0 +1,1 @@
+ALTER TABLE vufind.user ADD journal_subscription_format ENUM ('meistertask');
--- a/cpp/data/sql_updates/vufind.7
+++ b/cpp/data/sql_updates/vufind.7
@@ -1,1 +1,1 @@
-ALTER TABLE vufind.user ADD journal_subscription_format ENUM ('meistertask');
+ALTER TABLE vufind.user ADD COLUMN ixtheo_journal_subscription_format ENUM ('meistertask') DEFAULT NULL
--- a/cpp/new_journal_alert.cc
+++ b/cpp/new_journal_alert.cc
@@ -61,6 +61,19 @@ namespace {
 }
 
 
+typedef enum journal_subscription_format { MEISTERTASK } journal_subscription_format;
+
+
+const std::map<journal_subscription_format, std::string> JOURNAL_SUBSCRIPTION_FORMAT_TO_STRING_MAP {
+    { MEISTERTASK, "meistertask"}
+};
+
+
+const std::map<std::string, journal_subscription_format> JOURNAL_SUBSCRIPTION_FORMAT_ENUM_MAP {
+    { "meistertask", MEISTERTASK }
+};
+
+
 struct SerialControlNumberAndMaxLastModificationTime {
     std::string serial_control_number_;
     std::string last_modification_time_;
@@ -165,6 +178,8 @@ public:
     ~GenerateEmailContents() {};
     virtual std::string generateContent(const std::string &user_type, const std::string &name_of_user, const std::string &vufind_host,
                            const std::vector<NewIssueInfo> &new_issue_infos) const = 0;
+    virtual std::string generateContent(const std::string &vufind_host, const NewIssueInfo &new_issue_infos) const = 0;
+
 };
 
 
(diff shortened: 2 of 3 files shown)


cpp/data/sql_updates/vufind.7
```diff
@@ -0,0 +1 @@
+ALTER TABLE vufind.user ADD journal_subscription_format ENUM ('meistertask');
```
mtrojan-ub, line 1: - Since the whole subscription logic is in the IxTheo module, we should move this field to the ixtheo_user table.
- If we use an ENUM it would be better to have the values 'default' and 'meistertask', and add DEFAULT 'default' for the migration of existing rows.
- You should also provide a 2nd PR in the tuefind repo to add DB changes to the installer.
- To be compatible with all MySQL & MariaDB versions, you should use "ADD COLUMN" instead of "ADD".

cpp/new_journal_alert.cc
```diff
@@ -172,6 +206,220 @@ namespace std {
 namespace {
 
 
+class GenerateEmailContents {
+public:
+    ~GenerateEmailContents() {};
+    virtual std::string generateContent(const std::string &user_type, const std::string &name_of_user, const std::string &vufind_host,
```
mtrojan-ub, line 212: Seems like each of the two derived classes implements only one of these methods and aborts in the other one with a "not implemented" message. But i still think it's a good idea to implement a class concept.

One possibility might be to merge the two functions by changing the signature of generateContent to take a superset of all provided information (e.g. as a struct), and then each function only uses the subset of information that it actually needs.

What do you think?

cpp/new_journal_alert.cc
```diff
@@ -172,6 +206,220 @@ namespace std {
...
+         if (not new_issue_info.issue_.empty()) {
+             if (not volume_year_and_issue.empty())
+                 volume_year_and_issue += ", ";
+             volume_year_and_issue += new_issue_info.issue_;
+         }
+
+         email_contents += HtmlUtil::HtmlEscape(volume_year_and_issue) + "<br/>\n";
+
+         const std::string URL("https://" + vufind_host + "/Record/" + new_issue_info.series_control_number_);
+         email_contents += "<a href=" + URL + "#details>" +  URL + "</a>\n";
```
mtrojan-ub, line 318: Dont forget to export double quotes around the value of the href attribute.

cpp/new_journal_alert.cc
```diff
@@ -172,6 +206,220 @@ namespace std {
...
+                 email_contents += "    <li>" + HtmlUtil::HtmlEscape(volume_year_and_issue) + "</li>\n";
+                 last_volume_year_and_issue = volume_year_and_issue;
+                 email_contents += "    <ul>\n"; // start items
+             }
+
+             const std::string URL("https://" + vufind_host + "/Record/" + new_issue_info.control_number_);
+             std::string authors;
+             for (const auto &author : new_issue_info.authors_)
+                 authors += "&nbsp;&nbsp;&nbsp;" + HtmlUtil::HtmlEscape(author);
+             email_contents += "      <li><a href=" + URL + ">" + HtmlUtil::HtmlEscape(new_issue_info.issue_title_) + "</a>" + authors
```
mtrojan-ub, line 270: Dont forget to export double quotes around the value of the href attribute.

cpp/new_journal_alert.cc
```diff
@@ -172,6 +206,220 @@ namespace std {
...
+
+             if (volume_year_and_issue != last_volume_year_and_issue) {
+                 if (not new_serial)
+                     email_contents += "    </ul>\n"; // end items
+                 email_contents += "    <li>" + HtmlUtil::HtmlEscape(volume_year_and_issue) + "</li>\n";
+                 last_volume_year_and_issue = volume_year_and_issue;
+                 email_contents += "    <ul>\n"; // start items
+             }
+
+             const std::string URL("https://" + vufind_host + "/Record/" + new_issue_info.control_number_);
```
mtrojan-ub, line 266: Don't forget to UrlEncode the dynamic parts of the URL.

cpp/new_journal_alert.cc
```diff
@@ -172,6 +206,220 @@ namespace std {
...
+         }
+         if (not new_issue_info.issue_.empty()) {
+             if (not volume_year_and_issue.empty())
+                 volume_year_and_issue += ", ";
+             volume_year_and_issue += new_issue_info.issue_;
+         }
+
+         email_contents += HtmlUtil::HtmlEscape(volume_year_and_issue) + "<br/>\n";
+
+         const std::string URL("https://" + vufind_host + "/Record/" + new_issue_info.series_control_number_);
```
mtrojan-ub, line 317: Don't forget to UrlEncode the dynamic parts of the URL.

cpp/data/sql_updates/vufind.7
jriedl: As discussed.

cpp/new_journal_alert.cc
jriedl: As discussed.

cpp/new_journal_alert.cc
jriedl: As discussed

cpp/new_journal_alert.cc
jriedl: Done.

cpp/new_journal_alert.cc
jriedl: As discussed.

PR #3979, opened 2021-07-08
added gnd to wikidata lookup functionality

andreas-ub:


cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -0,0 +1,195 @@
+/** \file WikidataTest.cc
```
mtrojan-ub, line 1: wrong file name in comment

cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -0,0 +1,195 @@
...
+//Scenario 2:
+// use converted file from scenario 1 from cpp/data to create a map during 
+//  pipeline processing. The norm_data_input is extended by wikidata ids where possible
+//  and saved to 024 field indicator1:7 where wikidata id is not yet present
+[[noreturn]] void Usage() {
+    ::Usage("norm_data_marc_input norm_data_marc_output mapping_txt_file OR_USE_THIS_TO_CREATE_MAPPING_FILE: --create_mapping_file dnb_input_unzipped_file mapping_txt_file");
+}
+
+
+std::string ltrim(const std::string &s) {
```
mtrojan-ub, line 43: You should be able to use existing functions from StringUtil rather than writing your own ones.
StringUtil::Trim
StringUtil::LeftTrim
StringUtil::RightTrim

cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -0,0 +1,195 @@
...
+    if (marc_input_filename_or_create_flag == "--create_mapping_file") {
+        //e.g. "/..../authorities-person_lds_20210613.jsonld" and /usr/local/ub_tools/cpp/data/gnd_to_wiki.txt
+        ParseDataDnbFile(marc_output_filename_or_dnb_input, mapping_txt_filename); 
+        return EXIT_SUCCESS;
+    }
+
+    std::unordered_map<std::string, std::string> gnd_to_wikidataid;
+    ParseGndWikidataMappingFile(mapping_txt_filename, &gnd_to_wikidataid);
+    
+    std::unique_ptr<MARC::Reader> marc_reader(MARC::Reader::Factory(marc_input_filename_or_create_flag, MARC::FileType::BINARY));
```
mtrojan-ub, line 145: Is it really necessary to explicitly set BINARY here?

cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -0,0 +1,195 @@
...
+
+    if (unlikely(marc_input_filename_or_create_flag == marc_output_filename_or_dnb_input))
+        LOG_ERROR("Norm data input file name equals output file name!");
+
+    while (MARC::Record record = marc_reader.get()->read()) {
+        // 035|a (DE-588)118562215
+        std::string record_gnd;
+        std::string wikidata_id;
+
+        auto field_035(record.getFirstField("035"));
```
mtrojan-ub, line 156: You can replace the whole block by using MARC::GetGNDCode

cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -0,0 +1,195 @@
...
+        }
+        //record lookup
+        if (not record_gnd.empty()) {
+            auto gnd_to_wikidataid_iter = gnd_to_wikidataid.find(record_gnd);
+            if (gnd_to_wikidataid_iter != gnd_to_wikidataid.end()) {
+                wikidata_id = gnd_to_wikidataid_iter->second;
+                //std::cout << "Match: " << record_gnd << " --- " << wikidata_id << "\n";
+            }
+        }
+        //record write wikidata id (if not already there)
```
mtrojan-ub, line 176: It might be useful to move this whole block into MARC.h as well, similar to existing functions like Record::getDOIs()

mtrojan-ub: Also, performance might be better if you check wikidata_existing and skipping the record before executing the lookup against the gnd_to_wikidataid mapping file.

cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -0,0 +1,195 @@
...
+
+#include "util.h"
+#include "StringUtil.h"
+#include "MARC.h"
+#include <iostream>
+#include <fstream>
+#include <regex>
+
+//Scenario 1:
+// use --create_mapping_file parameter <fileparth> to generate the mapping file
```
mtrojan-ub, line 29: typo => "fileparth"

cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -0,0 +1,195 @@
...
+//  of the downloaded dnb authoriy dump (must bei unzipped first)
+//  Download from:  https://data.dnb.de/opendata/authorities-person_lds.jsonld.gz 
+//  and unzip to e.g. authorities-person_lds_20210613.jsonld
+//  output is stdout
+//Scenario 2:
+// use converted file from scenario 1 from cpp/data to create a map during 
+//  pipeline processing. The norm_data_input is extended by wikidata ids where possible
+//  and saved to 024 field indicator1:7 where wikidata id is not yet present
+[[noreturn]] void Usage() {
+    ::Usage("norm_data_marc_input norm_data_marc_output mapping_txt_file OR_USE_THIS_TO_CREATE_MAPPING_FILE: --create_mapping_file dnb_input_unzipped_file mapping_txt_file");
```
mtrojan-ub, line 39: Instead of OR_USE_THIS_TO_CREATE_MAPPING_FILE, please move the two invocation modes to different lines (maybe similar to installer.cc)

Also, if a program can be called in multiple modes, it might be more consistent to always require a first parameter that explicitly sets the mode (also like installer.cc) but i'm not absolutely sure about this one.

mtrojan-ub: In addition, it would be nice to move the comments into the usage output.

cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -0,0 +1,195 @@
...
+
+void ParseGndWikidataMappingFile(std::string filename, std::unordered_map<std::string, std::string> * const gnd_to_wikidataid) {
+    std::ifstream file(filename);
+    if (file.is_open()) {
+        std::string line;
+        std::string act_gnd;
+        std::string act_wiki;
+        while (std::getline(file, line)) {
+            if (StringUtil::StartsWith(line, "Name:") and StringUtil::Contains(line, "GND:") and StringUtil::Contains(line, "Wikidata:")) {
+                act_gnd = line.substr(line.find("GND:") + 4);
```
mtrojan-ub, line 114: Please do not hardcode lengths like 4 or 9, determine the length of "GND:" or "Wikidata:" instead

cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -0,0 +1,195 @@
...
+        LOG_ERROR("input or output files could not be opened");
+}
+
+void ParseGndWikidataMappingFile(std::string filename, std::unordered_map<std::string, std::string> * const gnd_to_wikidataid) {
+    std::ifstream file(filename);
+    if (file.is_open()) {
+        std::string line;
+        std::string act_gnd;
+        std::string act_wiki;
+        while (std::getline(file, line)) {
```
mtrojan-ub, line 112: If you would write a CSV file instead of this custom text file you could simply use TextUtil::ParseCSVFileOrDie to parse this and iterate over the lines.

cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -0,0 +1,195 @@
...
+    std::ofstream output_file(output_filename);
+    if (input_file.is_open() and output_file.is_open()) {
+        std::string line;
+        std::string act_gnd;
+        std::string act_name;
+        std::string act_wiki;
+        bool sameAs_reached(false);
+        bool read_preferred_name(false);
+        bool read_gnd_id(false);
+        while (std::getline(input_file, line)) {
```
mtrojan-ub, line 69: I would be more happy if we would use some SAX-like library to parse big JSON files like this; however since it's already the 4th kind of different data source we process and the first one that works more or less, we can leave it like this for now...

https://github.com/nlohmann/json

PR #3985, opened 2021-07-13
Ruschein

ruschein:


cpp/oai_pmh_harvester.cc
```diff
@@ -166,20 +153,20 @@ bool ListRecords(const std::string &url, const unsigned time_limit_in_seconds_pe
     }
 
     *resumption_token = ExtractResumptionToken(message_body, cursor, complete_list_size);
-    return not resumption_token->empty();
+    return false;//not resumption_token->empty();
```
mtrojan-ub, line 156: Is this for debugging?

cpp/data/installer/ixtheo.cronjobs
```diff
@@ -37,3 +37,4 @@ LOG_DIR=/usr/local/var/log/tuefind
 45 23 * * * "$BIN/rss_subset_aggregator" "--mode=email" "$EMAIL" "relbib" > "$LOG_DIR/rss_subset_aggregator_relbib.log" 2>&1
 {IF DEFINED(production)}0 17 1 Jul,Dec * "$BIN/ch.sh"{ENDIF}
 {IF DEFINED(production)}0 0 1 * * "$BIN/cologne.sh"{ENDIF}
+{IF DEFINED(production)}30 0 * * * "$BIN/generate_new_journal_alert_stats.sh"{ENDIF}
```
mtrojan-ub, line 40: Similar to other stats programs like collect_solr_stats_data and others, it might be good to run this in non-production mode as well (in case we would like to change something)

cpp/generate_new_journal_alert_stats.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -o errexit -o nounset
+
+readonly CONFIG_PATH=/usr/local/var/lib/tuelib/generate_new_journal_alert_stats.conf
+generate_new_journal_alert_stats relbib $(inifile_lookup "$CONFIG_PATH" "" relbib_recipients)
```
mtrojan-ub, line 5: Instead of sending these via E-Mail, i think it would be better to keep it similar to `export_vufind_redirects`. In that program, the generated statistics are written into a CSV file which can be accessed in the `/mnt/ZE020110/FID-Projekte/Statistik` folder.

cpp/harvest_ukdataservice.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
mtrojan-ub, line 1: Can we move all of these harvesting-related programs into a new subfolder? like e.g. `cpp/harvest` 
This would also affect programs like ch.sh, cologne.sh, the whole bnb_yaz stuff and maybe more.
(Of course, this would imply modifying cronjobs + Makefiles + maybe system_updates as well)

mtrojan-ub: ... as i thought about this a bit more, ch.sh and cologne.sh is not data we harvest, but data we provide to others. Should we group those scripts into a seperate subfolder as well?

cpp/oai_pmh_harvester.cc
```diff
@@ -241,11 +311,27 @@ int Main(int argc, char **argv) {
     }
 
     if (argc != 6 and argc != 7)
-        Usage();
+        ::Usage("[--skip-dups] [--ignore-ssl-certificates] base_url metadata_prefix [harvest_set_or_identifier] "
```
mtrojan-ub, line 314: Wouldn't it be a good idea to use a separate dups_db for each data source? (even if the hashes do not overlap, maybe for performance reasons? Or to be able to reset the hashes for a single source?)

In that case it would be necessary to pass the path to the dups_db via argv

cpp/oai_pmh_harvester.cc
ruschein: Yes.  Also it is fixed!

cpp/data/installer/ixtheo.cronjobs
ruschein: Done.

cpp/generate_new_journal_alert_stats.sh
ruschein: Done.

cpp/harvest_ukdataservice.sh
ruschein: We now have ../cpp/data_harvests.

cpp/oai_pmh_harvester.cc
ruschein: Done.

cpp/data/installer/control_number_guesser.sql
```diff
@@ -1,47 +0,0 @@
--- Control numbers are saved as a single text blob consisting of individual numbers concatenated together with the '|' character.
```
mtrojan-ub, line 1: Why has control_number_guesser.sql been removed?
(It is needed by lib/src/ControlNumberGuesser.cc)

cpp/generate_new_journal_alert_stats.sh
mtrojan-ub: Now the call of generate_new_journal_alert_stats does not match the usage anymore.
=> relbib_recipient is no longer needed
=> this means we might get rid of the whole config file
=> this also means that we could simply register the ixtheo+relbib conjobs directly in the crontab without thte need of an additional shell script

cpp/data/installer/ixtheo.cronjobs
```diff
@@ -12,6 +12,7 @@ LOG_DIR=/usr/local/var/log/tuefind
 0 0 * * * "$BIN/log_rotate" --max-rotations=4 "$LOG_DIR" "(?<!(java_mem_stats))\\.log$"
 0 0 * * * "$BIN/log_rotate" --no-of-lines-to-keep=200 "$LOG_DIR" "^java_mem_stats\\.log$"
 30 0 * * * cd "$BSZ_DATEN" && "$BIN/purge_old_data.py" "$EMAIL" > "$LOG_DIR/purge_old_data.log" 2>&1
+30 0 * * * "$BIN/generate_new_journal_alert_stats.sh"
```
mtrojan-ub, line 15: - We need a system_update script as well to add this cronjob to all existing servers
(- Also we could register ixtheo / relbib as different cases, see other comment)

cpp/generate_new_journal_alert_stats.cc
```diff
@@ -201,7 +200,10 @@ int Main(int argc, char **argv) {
     std::shared_ptr<DbConnection> db_connection(VuFind::GetDbConnection());
     CollectConfigStats(db_connection.get(), user_type, &stats);
     CollectUsageStats(user_type, &stats);
-    GenerateAndMailReport(email_recipient, stats);
+
+    const auto report_file(FileUtil::OpenOutputFileOrDie(REPORT_DIRECTORY + "new_journal_alert_stats." + user_type
```
mtrojan-ub, line 204: Since this task will run on multiple machines, we need to add the host name to the file name so they will not overwrite each other.

cpp/generate_new_journal_alert_stats.cc
```diff
@@ -152,38 +154,35 @@ void CollectUsageStats(const std::string &user_type, Stats * const stats) {
...
-                                      + "Total number of currently subscribed journals: "
-                                      + std::to_string(stats.no_of_subscribed_journals_) + "\n"
-                                      + "Number of subscribed journals w/ notifications: "
-                                      + std::to_string(stats.no_of_subscribed_journals_with_notifications_) + "\n"
-                                      + "Average number of notified articles per notified journal: "
-                                      + StringUtil::ToString(stats.average_number_of_notified_articles_per_notified_journal_))
-        > 299)
-        LOG_ERROR("failed to send an email report to \"" + email_address + "\"!");
+void GenerateReport(File * const report_file, const Stats &stats) {
+    (*report_file) << "\"Host\"," << TextUtil::CSVEscape(DnsUtil::GetHostname()) << '\n'
```
mtrojan-ub, line 158: The hostname can be removed here, since it should be part of the file name.

cpp/generate_new_journal_alert_stats.cc
```diff
@@ -152,38 +154,35 @@ void CollectUsageStats(const std::string &user_type, Stats * const stats) {
...
+                   << "\"Report interval in days\"," << stats.report_interval_in_days_ << '\n'
```
mtrojan-ub, line 159: Missing ending enclosure `"` before each newline

cpp/krim_ssoar.sh
```diff
@@ -3,7 +3,8 @@
 set -e
 
 echo "Download the data from SSOAR"
-oai_pmh_harvester https://www.ssoar.info/OAIHandler/request marcxml col_collection_10214 KRIM_SSOAR krim_ssoar.xml 20
+oai_pmh_harvester https://www.ssoar.info/OAIHandler/request marcxml col_collection_10214 KRIM_SSOAR \
```
mtrojan-ub, line 6: Do we need to migrate any existing dups db as soon as we activate this?

cpp/oai_pmh_harvester.cc
```diff
@@ -170,85 +157,179 @@ bool ListRecords(const std::string &url, const unsigned time_limit_in_seconds_pe
...
        ignore_ssl_certificates = true;
        --argc, ++argv;
     }
 
-    if (argc != 6 and argc != 7)
-        Usage();
+    if (argc != 7 and argc != 8)
+        ::Usage("[--skip-dups] [--ignore-ssl-certificates] base_url metadata_prefix [harvest_set_or_identifier] "
+                "control_number_prefix output_filename time_limit_per_request path_to_dups_database\n"
+                "If \"--skip-dups\" has been specified, records that we already encountered in the past won't\n"
```
mtrojan-ub, line 310: Do we really need separate args `[--skip-dups]` and `path_to_dups_database`?
It might be easier if we merge these together as [--path-to-dups-database=<path>], which implies that dups should be checked if the path is given.

cpp/oai_pmh_harvester.cc
```diff
@@ -170,85 +157,179 @@ bool ListRecords(const std::string &url, const unsigned time_limit_in_seconds_pe
...
+    const std::string harvest_set_or_identifier(argc == 7 ? argv[3] : "");
+    if (not harvest_set_or_identifier.empty()
+        and (not StringUtil::StartsWith(harvest_set_or_identifier, "set=")
+             and not StringUtil::StartsWith(harvest_set_or_identifier, "identifier=")))
+        LOG_ERROR("harvest_set_or_identifier must start with set= or identifier=!");
+
     const std::string control_number_prefix(argc == 7 ? argv[3] : argv[2]);
     const std::string output_filename(argc == 7 ? argv[5] : argv[4]);
     const std::string time_limit_per_request_as_string(argc == 7 ? argv[6] : argv[5]);
+    const std::string path_to_dups_database(argc == 7 ? argv[7] : argv[6]);
```
mtrojan-ub, line 332: Is this really correct? `argv[7] ` will not exist if `argc == 7`

cpp/generate_new_journal_alert_stats.cc
ruschein: No, the numbers should not intended to be quoted.

cpp/generate_new_journal_alert_stats.sh
ruschein: I completely removed this script and updated the crontab instead.

cpp/data/installer/control_number_guesser.sql
ruschein: I'm sorry, I can't remember why I did that.  In any case it is back now.

cpp/data/installer/ixtheo.cronjobs
ruschein: Done.

cpp/generate_new_journal_alert_stats.cc
ruschein: Done.

cpp/generate_new_journal_alert_stats.cc
ruschein: Done.

cpp/generate_new_journal_alert_stats.cc
ruschein: Wrong.

cpp/krim_ssoar.sh
ruschein: Done.

cpp/generate_new_journal_alert_stats.cc
mtrojan-ub: Right now, the field has an opening quote without an ending quote. So you should either add the ending quote or remove the opening quote, but it can not stay as it is right now.

cpp/oai_pmh_harvester.cc
ruschein: That sounds like a great idea! => Done!

cpp/oai_pmh_harvester.cc
ruschein: No, that was wrong, but it is fixed now.

cpp/krim_ssoar.sh
mtrojan-ub: This does not seem to match the new usage --path-to-dups-database.

cpp/data/system_updates/12.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -o errexit
+
+
+readonly HOSTNAME=$(hostname)
+if [[ $HOSTNAME == "ub16.uni-tuebingen.de" || $HOSTNAME == "ptah.ub.uni-tuebingen.de" ]]; then
```
mtrojan-ub, line 6: There are additional hostnames for which this should be changed, especially mb049.ub.uni-tuebingen.de and ub55.v264.uni-tuebingen.de. I think it would be better to check for TUEFIND_FLAVOUR=ixtheo instead.

cpp/data/system_updates/12.sh
```diff
@@ -0,0 +1,14 @@
...
+    if [ -e /var/spool/cron/root ]; then # CentOS
+        root_crontab=/var/spool/cron/root
+    else # Ubuntu
+        root_crontab=/var/spool/cron/crontabs/root
```
mtrojan-ub, line 10: This is correct for the servers, but e.g. on my local development machine this file will not exist because my root does not have a crontab. So i guess this would lead to an error when sed is called. 

I think it would be better to determine the OS depending on whether /etc/lsb-release (Ubuntu) or /etc/redhat-release (CentOS) exists. Then the path could be set according to the OS. If the crontab file does not exist yet, the replacing should be skipped.

cpp/generate_new_journal_alert_stats.sh
ruschein: The config file is also gone, BTW!

cpp/krim_ssoar.sh
ruschein: Fixed.

cpp/data/system_updates/12.sh
ruschein: Done.

cpp/data/system_updates/12.sh
ruschein: I think the simplest solution is to check the two locations and to not update if neither exists which is what I have done now.

cpp/data/system_updates/12.sh
```diff
@@ -0,0 +1,15 @@
...
+if [[ $TUEFIND_FLAVOUR == "ixtheo" ]]; then
+    if [ -e /var/spool/cron/root ]; then # CentOS
+        root_crontab=/var/spool/cron/root
+    elif [ -e /var/spool/cron/crontabs/root ]; then # Ubuntu
+        root_crontab=/var/spool/cron/crontabs/root
+    else
+        exit 0 # There is no crontab for root.
+    fi
+
+    sed --in-place '/^# END VUFIND AUTOGENERATED/i 30 0 * * * "$BIN/generate_new_journal_alert_stats ixtheo days_in_last_month"\n30 0 * * * "$BIN/generate_new_journal_alert_stats relbib days_in_last_month"' $root_crontab
```
mtrojan-ub, line 14: This does not insert the new cronjob at the same position as in ixtheo.cronjobs, but it should do so.

ruschein: Done!

PR #3993, opened 2021-07-16
added wikipedia id extraction and norm data tagging

andreas-ub:


cpp/pipelines/add_subsystem_tags.cc
```diff
@@ -425,27 +424,100 @@ void AddSubsystemTags(MARC::Reader * const marc_reader, MARC::Writer * const mar
...
+
+
 } //unnamed namespace
 
 
 int Main(int argc, char **argv) {
-    if (argc != 4)
-        ::Usage("marc_input authority_records marc_output");
+    if (argc != 5)
+        ::Usage("marc_input authority_records marc_output authority_output");
```
mtrojan-ub, line 493: maybe we should rename "authority_records" to "authority_input" to keep things consistent

cpp/pipelines/add_subsystem_tags.cc
```diff
@@ -425,27 +424,100 @@ void AddSubsystemTags(MARC::Reader * const marc_reader, MARC::Writer * const mar
 }
 
 
+void ExtractAuthors(MARC::Reader * const marc_reader, std::map<std::string, std::set<std::string>> * authors,
```
mtrojan-ub, line 427: `* authors` should be `* const authors`

cpp/pipelines/add_subsystem_tags.cc
```diff
@@ -425,27 +424,100 @@ void AddSubsystemTags(MARC::Reader * const marc_reader, MARC::Writer * const mar
...
+                        const std::unordered_set<std::string> &bible_studies_gnd_numbers,
+                        const std::unordered_set<std::string> &canon_law_gnd_numbers) {
+    static std::vector<std::string> tags_to_check{ "100", "110", "111", "700", "710", "711" };
+    unsigned record_count(0);
+    while (const MARC::Record record = marc_reader->read()) {
+        ++record_count;
+        for (auto tag_to_check : tags_to_check) {
+            for (auto &field : record.getTagRange(tag_to_check)) {
+                const auto subfields(field.getSubfields());
```
mtrojan-ub, line 436: I wonder why there is no getSubfields(subfield_code) function yet... but it should be a good idea to add one to MARC.h and shorten this code a bit.

mtrojan-ub: Maybe including a second parameter to ignore empty subfields to shorten this even more

mtrojan-ub: Or you could implement a function like getFirstSubfieldWithCodeAndPrefix

cpp/pipelines/add_subsystem_tags.cc
```diff
@@ -425,27 +424,100 @@ void AddSubsystemTags(MARC::Reader * const marc_reader, MARC::Writer * const mar
...
+                        if (likely(not author.empty())) {
+                            if (not StringUtil::StartsWith(author, "(DE-627)"))
+                                continue;
+                            std::string author_id(author.substr(std::string("(DE-627)").length()));
+                            auto it_author = authors->find(author_id);
+                            std::set<std::string> instances;
+                            if (it_author != authors->end())
+                                instances = it_author->second;
+                            
+                            if (IsRelBibRecord(record))
```
mtrojan-ub, line 449: Methods like IsRelBibRecord, IsBibleStudiesRecord... etc. have lots of logic. Maybe we should execute these only once per record instead of once per author and store their result in boolean variables?

cpp/pipelines/add_subsystem_tags.cc
mtrojan-ub: The authors map can get very large, so it might be better to use std::unordered_map for performance reasons if we do not need any guaranteed order.

cpp/pipelines/add_subsystem_tags.cc
```diff
@@ -425,27 +424,100 @@ void AddSubsystemTags(MARC::Reader * const marc_reader, MARC::Writer * const mar
...
+                                instances.emplace("r");
+                            if (IsCanonLawRecord(record, canon_law_gnd_numbers))
+                                instances.emplace("c");
+                            if (IsBibleStudiesRecord(record, bible_studies_gnd_numbers))
+                                instances.emplace("b");
+                            
+                            if (authors->find(author_id) == authors->end())
```
mtrojan-ub, line 456: This will replace the value even if it already existed and even if we added the values using the iterator (which shouldn't make an emplace necessary).

How about we turn this around => Create an empty std::set there if the key doesn't exist, and then add all values later using the iterator?

cpp/pipelines/add_subsystem_tags.cc
mtrojan-ub: And we always should add an "i" too for ixtheo, so we know there is an existing title for this author here

cpp/pipelines/add_subsystem_tags.cc
```diff
@@ -425,27 +424,100 @@ void AddSubsystemTags(MARC::Reader * const marc_reader, MARC::Writer * const mar
...
+        }
+    }
+}
+
+
+void TagAuthors(MARC::Reader * const authority_reader, MARC::Writer * const authority_writer, std::map<std::string, std::set<std::string>> &authors) {
+    while (MARC::Record record = authority_reader->read()) {
+         auto it_authors = authors.find(record.getControlNumber());
+         if (it_authors != authors.end()) {
+             std::vector<MARC::Subfield> tits{ { 'a', "ixtheo" } };
```
mtrojan-ub, line 473: "tits" is a bad abbreviation :smirk:

cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -51,19 +51,21 @@ void ParseDataDnbFile(std::string input_filename, std::string output_filename) {
...
+        std::string act_wikipedia;
         bool sameAs_reached(false);
         bool read_preferred_name(false);
         bool read_gnd_id(false);
         while (std::getline(input_file, line)) {
             if (line == "}, {") {
-                if (not act_wiki.empty() and not act_name.empty() and not act_wiki.empty()) {
-                    output_file << "Name: " << act_name << " GND: " << act_gnd << " Wikidata: " << act_wiki << "\n";
-                    std::cout << "Name: " << act_name << " GND: " << act_gnd << " Wikidata: " << act_wiki << "\n";
+                if (not act_wikidata.empty() and not act_name.empty() and not act_wikidata.empty()) {
```
mtrojan-ub, line 61: duplicate condition `not act_wikidata.empty()`. Is the third condition meant to be `not act_wikipedia.empty()`?

cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -51,19 +51,21 @@ void ParseDataDnbFile(std::string input_filename, std::string output_filename) {
...
+                    output_file << "Name: " << act_name << " GND: " << act_gnd << " Wikidata: " << act_wikidata << " Wikipedia: " << act_wikipedia << "\n";
+                    //std::cout << "Name: " << act_name << " GND: " << act_gnd << " Wikidata: " << act_wikidata << " Wikipedia: " << act_wikipedia << "\n";
```
mtrojan-ub, line 63: This line should either be re-enabled or removed entirely

cpp/pipelines/add_authority_wikidata_ids.cc
```diff
@@ -92,21 +98,27 @@ void ParseDataDnbFile(std::string input_filename, std::string output_filename) {
...
-        std::string act_wiki;
+        std::string act_wikidata;
+        std::string act_wikipedia;
         while (std::getline(file, line)) {
             const std::string NAME = "Name:";
             const std::string GND = "GND:";
             const std::string WIKIDATA = "Wikidata:";
-            if (StringUtil::StartsWith(line, NAME) and StringUtil::Contains(line, GND) and StringUtil::Contains(line, WIKIDATA)) {
+            const std::string WIKIPEDIA = "Wikipedia:";
+            if (StringUtil::StartsWith(line, NAME) and StringUtil::Contains(line, GND) and StringUtil::Contains(line, WIKIDATA) and StringUtil::Contains(line, WIKIPEDIA)) {
```
mtrojan-ub, line 114: What happens if a dataset has a Wikidata entry but no Wikipedia entry?

mtrojan-ub (changes requested):
see comments

cpp/pipelines/add_authority_wikidata_ids.cc
andreas-ub: Thanx, yes this should be act_wikipedia.empty()

cpp/lib/src/MARC.cc
```diff
@@ -279,6 +279,18 @@ std::string Record::Field::getFirstSubfieldWithCode(const char subfield_code) co
 }
 
 
+std::string Record::Field::getFirstSubfieldWithCodeAndPrefix(const char subfield_code, const std::string prefix) const {
```
mtrojan-ub, line 282: We can use a reference for &prefix

PR #3995, opened 2021-07-19
added functionality to write sekkor entries only if present in total aut data

andreas-ub:


cpp/merge_tools/apply_differential_update.cc
```diff
@@ -92,23 +113,26 @@ void PatchArchiveMembersAndCreateOutputArchive(LocalDataDB * const local_data_db
...
 
     const auto authority_writer(MARC::Writer::Factory(output_directory + "/aut.mrc", MARC::FileType::BINARY));
-    std::unordered_set<std::string> previously_seen_authority_ppns;
+    PreFetchPPNSOfInput(input_archive_members, { BSZUtil::AUTHORITY_RECORDS }, &authority_ppns_in_input);
     CopySelectedTypes(local_data_db, difference_archive_members, authority_writer.get(), { BSZUtil::AUTHORITY_RECORDS },
-                      &previously_seen_authority_ppns);
+                      &authority_ppns_in_delta, authority_ppns_in_input, true);
     CopySelectedTypes(local_data_db, input_archive_members, authority_writer.get(), { BSZUtil::AUTHORITY_RECORDS },
-                      &previously_seen_authority_ppns);
+                      &authority_ppns_in_delta, authority_ppns_in_input, false);
```
mtrojan-ub, line 135: As soon as we pass authority_ppns_in_delta to the function a second time, it will no longer contain only delta ppn's, so we should find  a better name for this variable.

mtrojan-ub: ... or => overload the function and only pass it in the first call

cpp/merge_tools/apply_differential_update.cc
```diff
@@ -63,22 +64,42 @@ void CopyAndCollectPPNs(LocalDataDB * const local_data_db, MARC::Reader * const
...
-                       const std::set<BSZUtil::ArchiveType> &selected_types, std::unordered_set<std::string> * const previously_seen_ppns)
+                       const std::set<BSZUtil::ArchiveType> &selected_types, std::unordered_set<std::string> * const authority_ppns_in_delta,
+                       const std::unordered_set<std::string>  &authority_ppns_in_input, bool check_input_ppns = false)
 {
     for (const auto &archive_member : archive_members) {
         if (selected_types.find(BSZUtil::GetArchiveType(archive_member)) != selected_types.cend()) {
             const auto reader(MARC::Reader::Factory(archive_member, MARC::FileType::BINARY));
-            CopyAndCollectPPNs(local_data_db, reader.get(), writer, previously_seen_ppns);
+            //only check if delta records are present in source record if it is a sekkor file, not if TA file etc.
+            check_input_ppns |= StringUtil::Contains(archive_member, "sekkor");
```
mtrojan-ub, line 85: The |= can cause side effects. Imagine you have an archive with multiple members, check_input_ppns is passed as false and then check_input_ppns would stay to true if the first member is sekkor but the others aren't which should not happen, but anyway it would be more consistant to use a copy here.

cpp/merge_tools/apply_differential_update.cc
```diff
@@ -63,22 +64,42 @@ void CopyAndCollectPPNs(LocalDataDB * const local_data_db, MARC::Reader * const
...
+                         std::unordered_set<std::string> * const authority_ppns_in_input)
+{
+    for (const auto &archive_member : archive_members) {
+        if (selected_types.find(BSZUtil::GetArchiveType(archive_member)) != selected_types.cend()) {
+            const auto reader(MARC::Reader::Factory(archive_member, MARC::FileType::BINARY));
+            while (auto record = reader->read()) {
+                const auto PPN(record.getControlNumber());
+                authority_ppns_in_input->emplace(PPN);
+            }
+            reader->rewind();
```
mtrojan-ub, line 102: If we initialize a new reader on every iteration of the for loop and don't re-use that later, there is no need to rewind.

cpp/merge_tools/apply_differential_update.cc
```diff
@@ -63,22 +64,42 @@ void CopyAndCollectPPNs(LocalDataDB * const local_data_db, MARC::Reader * const
...
+
+void PreFetchPPNSOfInput(const std::vector<std::string> &archive_members, const std::set<BSZUtil::ArchiveType> &selected_types, 
+                         std::unordered_set<std::string> * const authority_ppns_in_input)
+{
+    for (const auto &archive_member : archive_members) {
+        if (selected_types.find(BSZUtil::GetArchiveType(archive_member)) != selected_types.cend()) {
+            const auto reader(MARC::Reader::Factory(archive_member, MARC::FileType::BINARY));
+            while (auto record = reader->read()) {
+                const auto PPN(record.getControlNumber());
+                authority_ppns_in_input->emplace(PPN);
```
mtrojan-ub, line 100: this could be a one-liner

cpp/merge_tools/apply_differential_update.cc
```diff
@@ -63,22 +64,42 @@ void CopyAndCollectPPNs(LocalDataDB * const local_data_db, MARC::Reader * const
                 record.truncate(first_local_field);
             }
 
-            previously_seen_ppns->emplace(PPN);
+            authority_ppns_in_delta->emplace(PPN);
             if (record.getFirstField("ORI") == record.end())
                 record.insertField("ORI", 'a', FileUtil::GetLastPathComponent(reader->getPath()));
-            writer->write(record);
+            if (check_input_ppns == false || authority_ppns_in_input.find(PPN) != authority_ppns_in_input.end())
```
mtrojan-ub, line 70: please use "not" and "or" here for better readability

cpp/merge_tools/apply_differential_update.cc
mtrojan-ub: Or it might even be better to move this to the calling site and decide it once depending on whether the archive file name contains sekkor

mtrojan-ub:
We should wait with the merge until after our liveserver has been updated next time, so we have some days to test it.

cpp/lib/include/FileUtil.h
```diff
@@ -191,6 +191,12 @@ class Directory {
 off_t GetFileSize(const std::string &path);
 
 
+/** \return The last modification timestamp of the file named by "path".
+ *  \note   Returns false if "path" does not exist or we don't have the rights to stat it.
+ */
+bool GetLastModificationDate(const std::string &path, timespec * mtim);
```
mtrojan-ub, line 197: - timespec contains a timestamp and not a date, so the function should rather be renamed to `GetLastModificationTimestamp`
- mtim should be changed to `timespec * const mtim`
- I'm not very familiar with the timespec data type, but it looks like the existing function GetMostRecentlyModifiedFile uses this data type as well. Maybe you could change that function to internally use the new function as well?

cpp/lib/include/Zeder.h
```diff
@@ -405,7 +405,7 @@ class EndpointDownloader {
 public:
     virtual ~EndpointDownloader() = default;
 public:
-    virtual bool download(EntryCollection * const collection) = 0;
+    virtual bool download(EntryCollection * const collection, bool disable_cache_mechanism) = 0;
```
mtrojan-ub, line 408: This should be a const bool. Also it should have false as a default value, so we only need to override it if we explicitly want to turn the Cache off.

cpp/zotaut/zeder_to_zotero_importer.cc
```diff
@@ -170,7 +170,7 @@ void DownloadZederEntries(const Zeder::Flavour flavour, const std::unordered_set
         new Zeder::FullDumpDownloader::Params(endpoint_url, entries_to_download, columns_to_download, filter_regexps));
 
     auto downloader(Zeder::FullDumpDownloader::Factory(Zeder::FullDumpDownloader::Type::FULL_DUMP, std::move(downloader_params)));
-    if (not downloader->download(downloaded_entries))
+    if (not downloader->download(downloaded_entries, true))
```
mtrojan-ub, line 173: If we pass a bool to a function, our convention is to add a short comment to see its name, e.g. `/*disable_cache_mechanism=*/ true`

cpp/zotaut/zotero_enhancement_maps_generator.cc
```diff
@@ -63,7 +63,7 @@ void DownloadZederInstanceEntries(const Zeder::Flavour flavour,
         new Zeder::FullDumpDownloader::Params(endpoint_url, entries_to_download, columns_to_download, filter_regexps));
 
     auto downloader(Zeder::FullDumpDownloader::Factory(Zeder::FullDumpDownloader::Type::FULL_DUMP, std::move(downloader_params)));
-    if (not downloader->download(&downloaded_flavour_entries))
+    if (not downloader->download(&downloaded_flavour_entries, true))
```
mtrojan-ub, line 66: Here too

cpp/lib/src/Zeder.cc
```diff
@@ -620,17 +620,70 @@ void FullDumpDownloader::parseRows(const Params &params, const std::shared_ptr<J
...
 
     std::unordered_map<std::string, ColumnMetadata> column_to_metadata_map;
     std::shared_ptr<JSON::JSONNode> json_data;
+    std::shared_ptr<JSON::JSONNode> json_cached_data;
+
+    bool use_cache = not disable_cache_mechanism;
+    bool cache_present = false;
+    bool downloaded = false;
+    std::string zeder_cache_path = UBTools::GetTuelibPath();
+    if (StringUtil::EndsWith(params->endpoint_url_, "ixtheo", true))
```
mtrojan-ub, line 636: Another option would be to replace the endpoint_path by the Zeder Flavour in the params class. Then the downloader could determine the url internally and still have the information about the instance in a more structured way. What do you think?

cpp/lib/src/Zeder.cc
```diff
@@ -620,17 +620,70 @@ void FullDumpDownloader::parseRows(const Params &params, const std::shared_ptr<J
...
+                if (not FileUtil::GetLastModificationDate(zeder_cache_path, &mtim))
+                    use_cache = false;
+                else {
+                    double dif = std::difftime(now.tv_sec, mtim.tv_sec);
+                    if (dif > 60.0 /*min*/ * 60.0 /*sec*/)
+                        use_cache = false;
+                }
+            }
+        }
+        else
```
mtrojan-ub, line 662: Please check the error condition in the if branch and put the rest into the else branch, just like when using json_parser.parse()

cpp/lib/src/Zeder.cc
```diff
@@ -620,17 +620,70 @@ void FullDumpDownloader::parseRows(const Params &params, const std::shared_ptr<J
...
+            downloaded = true;
+    }
+    else {
+        json_data = std::move(json_cached_data);
+        LOG_INFO("Used zeder cache file");
+    }
 
+    if (downloaded)
```
mtrojan-ub, line 683: The logic in this function seems to be mixed-up very much. Can we simplify / restructure it somehow?

cpp/zotaut/zeder_to_zotero_importer.cc
andreas-ub: so one the one hand we make it more verbose by adding the comment one the other hand it is far less verbose by using default parameters.

cpp/pipelines/add_superior_and_alertable_flags.cc
```diff
@@ -101,11 +102,24 @@ void ProcessRecord(MARC::Writer * const marc_writer, const std::unordered_set<st
 
 
 void AddSuperiorFlag(MARC::Reader * const marc_reader, MARC::Writer * const marc_writer,
-                     const std::unordered_set<std::string> &superior_ppns)
+                     const std::unordered_set<std::string> &superior_ppns, const std::string &flavour)
 {
     unsigned modified_count(0);
+    
+    Zeder::SimpleZeder zeder(Zeder::IXTHEO, { "eppn", "pppn", "kat" });
```
mtrojan-ub, line 109: As discussed => please extend SimpleZeder to allow an optional arg as well that will be used as filter_regexps, and use it to compare kat=IxTheo.

PR #3998, opened 2021-07-20
Don't abort when encountering canon law references which we don't kno…

ruschein:
…w how to handle!


mtrojan-ub:
... but this would violate our strategy to abort rather than writing warnings to the log which nobody might check for years

PR #3999, opened 2021-07-20
Fixes for tuelib issue 1587.  Fixes over 90% of the problem cases!

ruschein:


cpp/lib/include/RangeUtil.h
```diff
@@ -101,7 +101,7 @@ bool ParseCanonLawRanges(const std::string &ranges, unsigned * const range_start
 
 
 /** \return True if "text" contained a valid time range, o/w false. */
-bool ConvertTextToTimeRange(const std::string &text, std::string * const range, const bool special_case_centuries = false);
+bool ConvertTextToTimeRange(std::string text, std::string * const range, const bool special_case_centuries = false);
```
mtrojan-ub, line 104: Please add documentation regarding the meaning of special_case_centuries.

mtrojan-ub:
Seems to be correct, but it also looks like lots of copy & paste code. I wonder if we could use a different approach here:
- Split the range into range_start + range_end
- If range_start is empty => set it to MINUS_INFINITY, else normalize it
- If range_end is empty => set it to 31.12.CURRENT_YEAR (or PLUS_INFINITY?), else normalize it
- Normalize would be a function that handles things like "ca.", "XX", Before Christ patterns and so on. This could have a generic part for range_start+range_end, and a special part which does stuff depending on whether it is range_start or range_end, e.g. if XX.XX.2020 is range_start, we need 01.01, but on range_end we need 31.12.
- As last step, convert the std::string representation of the range_start and range_end to the output range.

What do you think?

ruschein:
I am not sure if all of this would work.  For example we don't actually have a year representation of minus infinity, we only have a representation as a part of a range.  Also this would require extensive re-testing and I spent a fair amount of time on that.  Overall it probably would make things more readable though.

PR #4004, opened 2021-07-26
changed wikidata id and wikipedia link extraction

andreas-ub:
changed wikidata id extraction and wikipedia link extraction to prevent dupplicates


cpp/lib/src/MARC.cc
```diff
@@ -2571,6 +2571,21 @@ bool GetWikidataId(const Record &record, std::string * const wikidata_id) {
 }
 
 
+bool GetWikipediaLink(const Record &record, std::string * const wikipedia_link) {
+    wikipedia_link->clear();
+    for (const auto &_670_field : record.getTagRange("670")) {
+        const Subfields _670_subfields(_670_field.getSubfields());
+        const std::string _670a_field(_670_subfields.getFirstSubfieldWithCode('a'));
+        const std::string _670u_field(_670_subfields.getFirstSubfieldWithCode('u'));
+        if (StringUtil::Contains(_670a_field, "wikipedia")) {
```
mtrojan-ub, line 2580: In 670a it is "Wikipedia" instead of "wikipedia".

PR #4008, opened 2021-07-28
modified post-merge git-hook to delete intermediate files

andreas-ub:


cpp/mysql_database_patcher.cc
```diff
@@ -161,8 +161,10 @@ int Main(int argc, char *argv[]) {
         std::string database;
         unsigned version;
         SplitIntoDatabaseAndVersion(update_filename, &database, &version);
-        if (last_schema.empty() or database != last_schema)
-            db_connection.queryOrDie("USE " + database);
+        if (last_schema.empty() or database != last_schema) {
+            if (db_connection.mySQLDatabaseExists(database))
```
mtrojan-ub, line 165: There is already existing mySQLDatabaseExists logic inside the ApplyUpdate function, so please move the USE statement there.

cpp/mysql_database_patcher.cc
```diff
@@ -89,17 +89,25 @@ void LoadAndSortUpdateFilenames(const bool test, const std::string &directory_pa
...
     SplitIntoDatabaseAndVersion(update_filename, &database, &update_version);
+    *current_schema = database;
 
     if (not db_connection->mySQLDatabaseExists(database)) {
         LOG_INFO("database \"" + database + "\" does not exist, skipping file " + update_filename);
         return;
     }
 
+    if (last_schema.empty() or database != last_schema) {
+        if (db_connection->mySQLDatabaseExists(database)) {
```
mtrojan-ub, line 105: Please avoid mySQLDatabaseExists being called twice in this function.

PR #4020, opened 2021-08-10
Riedl

jriedl:


cpp/create_full_text_db.cc
```diff
@@ -338,6 +343,13 @@ int Main(int argc, char **argv) {
         ++argv, --argc;
     }
 
+    bool include_list_of_references(false);
+        if (argc > 1 and StringUtil::StartsWith(argv[1], "--include-list-of-references")) {
```
andreas-ub, line 347: "if" could be placed more to the left.

PR #4021, opened 2021-08-10
Initial support for Postgres.

ruschein:


cpp/data/installer/scripts/install_centos_packages.sh
```diff
@@ -44,9 +44,10 @@ dnf --assumeyes update
 InstallIfMissing "ca-certificates"
 dnf --assumeyes install \
     ant bc cifs-utils clang crontabs ftp gawk gcc-c++ git glibc-all-langpacks ImageMagick incron java-1.8.0-openjdk-devel \
-    jq libcurl-devel libdb-devel libsq3-devel libstemmer-devel libuuid-devel libwebp libxml2-devel libxml2 libxslt lsof lz4 make mariadb \
-    mariadb-devel mariadb-server mariadb-server-utils mod_ssl mutt openssl-devel pcre-devel policycoreutils-python-utils \
-    poppler-utils python3 python3-pexpect python3-pyflakes rpmdevtools rsync sqlite sudo tidy unzip xerces-c-devel yaz
+    jq libcurl-devel libdb-devel libsq3-devel libpq-devel libstemmer-devel libuuid-devel libwebp libxml2-devel libxml2 \
+    libxslt lsof lz4 make mariadb mariadb-devel mariadb-server mariadb-server-utils mod_ssl mutt openssl-devel pcre-devel \
+    policycoreutils-python-utils poppler-utils postgresql postgresql-server python3 python3-pexpect python3-pyflakes \
```
mtrojan-ub, line 49: Since we only want the postgresql support to connect to external servers, we do not want to install a full-fledged postgresql-server on all machines right now (including production). Or is it mandatory if we want to use the API?

cpp/data/system_updates/15.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -o errexit -o nounset
+
+if [ -f /etc/lsb-release ]; then
+    apt-get --quiet --yes install libpq-dev postgresql postgresql-client
+else
+    dnf --assumeyes install libpq-devel postgresql postgresql-server
```
mtrojan-ub, line 7: ... here as well

cpp/data/system_updates/15.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -o errexit -o nounset
+
+if [ -f /etc/lsb-release ]; then
+    apt-get --quiet --yes install libpq-dev postgresql postgresql-client
+else
+    dnf --assumeyes install libpq-devel postgresql
+fi
```
jriedl, line 8: `15.sh` has been created in a different context.

cpp/lib/src/DbConnection.cc
```diff
@@ -105,15 +106,18 @@ DbConnection::~DbConnection() {
...
                 if (result_code != SQLITE_OK)
                     LOG_ERROR("failed to finalise an Sqlite3 statement! (" + getLastErrorMessage() + ", code was "
-                          + std::to_string(result_code) + ")");
+                              + std::to_string(result_code) + ")");
             }
             if (::sqlite3_close(sqlite3_) != SQLITE_OK)
                 LOG_ERROR("failed to cleanly close an Sqlite3 database!");
+        } else { // Postgres
+            ::PQfinish(pg_conn_);
+            pg_conn_ = nullptr;
```
jriedl, line 120: maybe another `else if`-branch and an error in a general catch-all `else`?

PR #4039, opened 2021-08-31
Ruschein

ruschein:
as discussed today:
- please move the files into a subdirectory and adjust the Makefile
- we will leave this open until next tuesday


mtrojan-ub:
As discussed later today, the PR can be merged by now + pulled on ptah and the other changes can be done later.

PR #4040, opened 2021-08-31
implemented cgi-restarting and automatic reconnecting on client side

andreas-ub:


cpp/cgi_progs/Makefile
```diff
@@ -23,7 +25,17 @@ $(PROGS): % : %.o ../lib/libubtue.a Makefile
 install: $(PROGS)
 	@echo "Installing CGI programs..."
 	mkdir --parents $(CGI_DIR)
-	cp $(PROGS) $(CGI_DIR)
+	if test -f /etc/centos-release; then \
+	    APACHE_START=systemctl start httpd; \
+	    APACHE_STOP=systemctl stop httpd; \
+	fi
+	if [ $(shell pgrep -c -f message_dispatcher) -ne 0 ]; then \
```
jriedl, line 32: Please use the long versions of parameters

cpp/cgi_progs/message_dispatcher.cc
```diff
@@ -66,6 +90,7 @@ int Main(int argc, char *argv[]) {
     // Wait for incoming message 
     while(true) {
        sd_bus_wait(bus, UINT64_MAX);
+       InitializeTimer();
```
jriedl, line 93: Must probably be called once before the while loop is entered. Otherwise it might never exit on itself.

cpp/cgi_progs/message_dispatcher.cc
```diff
@@ -33,6 +33,30 @@
...
 #include "util.h"
 #include "WebUtil.h"
+#include "TimerUtil.h"
+
+const int TIMEOUT = 3 * 1000 * 60;
+
+
+extern "C" void InterruptCgi(int /*signal_no*/) {
+    std::cerr << "Translator timeout reached - stopping cgi (will be reinitialized by sse  client)" << std::endl;
+    std::exit(0);
```
jriedl, line 43: Probably the "finish" operations should be executed here as well.

cpp/cgi_progs/Makefile
```diff
@@ -23,7 +25,17 @@ $(PROGS): % : %.o ../lib/libubtue.a Makefile
 install: $(PROGS)
 	@echo "Installing CGI programs..."
 	mkdir --parents $(CGI_DIR)
-	cp $(PROGS) $(CGI_DIR)
+	if test -f /etc/centos-release; then \
```
jriedl, line 28: Probably better to use `[...]` instead of `test` for consistency

cpp/cgi_progs/Makefile
```diff
@@ -3,6 +3,8 @@ INC = $(LIB)/include
 include ../Makefile.inc
 PROGS = display_translate_stats full_text_cache_monitor full_text_lookup ini_validator translate_chainer translator translator_ajax zotero_cgi restart_zts message_dispatcher
 CGI_DIR = /var/www/cgi-bin
+APACHE_START=systemctl start apache2
+APACHE_STOP=systemctl stop apache2
```
jriedl, line 7: How about something like
`APACHE_SYSTEMCTL_ARGUMENT := $(shell if [ -f /etc/lsb_release ]; then echo "apache2"; elif [ -f /etc/centos-release ]; echo "httpd"; else echo "")`

This can be given as an argument to systemctl without further branching.

and abort if unset with something like:
```bash
ifndef MY_FLAG
.ABORT
endif
```

PR #4054, opened 2021-09-10
added update part for existing vufind translations

andreas-ub:


cpp/pipelines/extract_vufind_translations_for_translation.cc
```diff
@@ -77,6 +77,10 @@ void InsertTranslations(
                 if (not next_version_id.empty())
                     continue;
             }
+
+            const std::string UPDATE_STMT(
+                "UPDATE vufind_translations SET translation=\"" + translation + "\" WHERE id=" + row["id"]);
```
mtrojan-ub, line 82: `translation` should be escaped when building the SQL statement

PR #4070, opened 2021-09-21
added first phase of subsystem (relbib etc.) refactoring

andreas-ub:


cpp/pipelines/add_subscription_bundle_entries.cc
```diff
@@ -53,18 +53,27 @@ MARC::Record GenerateBundleRecord(const std::string &record_id, const std::strin
     record.insertField("SPR", { { 'a', "1" /* is superior work */ },
                                 { 'b', "1" /* series has not been completed */ } });
     record.insertField("935", 'c', "subskriptionspaket" );
+    std::vector<MARC::Subfield> subsystems{};
```
mtrojan-ub, line 56: You can remove the brackets `{}` here.

cpp/lib/src/MARC.cc
```diff
@@ -1592,6 +1592,32 @@ bool Record::addSubfield(const Tag &field_tag, const char subfield_code, const s
...
+    return success;
+}
+
+
+bool Record::hasSubfieldWithValue(const Tag &field_tag, const char subfield_code, const std::string &subfield_value) const {
+    const auto &field(findTag(field_tag));
+    if (field == fields_.end()) {
+        return false;
+    }
+    return field->hasSubfieldWithValue(subfield_code, subfield_value);
```
mtrojan-ub, line 1618: What if we have a repeatable field and the first field doesnt have the subfield but the second one has?

cpp/lib/src/MARC.cc
```diff
@@ -1592,6 +1592,32 @@ bool Record::addSubfield(const Tag &field_tag, const char subfield_code, const s
...
+    const auto &field(findTag(field_tag));
+    bool success = true;
+
+    if (field == fields_.end()) {
+        success = insertField(field_tag, std::string(1, subfield_code) + subfield_value);
+    }
+
+    if (success) {
+        const auto field_lookup = findTag(field_tag);
+        if (not field_lookup->hasSubfieldWithValue(subfield_code, subfield_value)) {
```
mtrojan-ub, line 1605: Also: What happens if we have a repeatable field and the second or third one already contains the value?

cpp/pipelines/add_subscription_bundle_entries.cc
```diff
@@ -53,18 +53,27 @@ MARC::Record GenerateBundleRecord(const std::string &record_id, const std::strin
...
     if (exclude_ixtheo)
         record.addSubfield("935", 'x', "1");
-    if (include_relbib)
-        record.insertField("REL", { { 'a', "1" } });
-    if (include_bibstudies)
-        record.insertField("BIB", { { 'a', "1" } });
-    if (include_churchlaw)
-        record.insertField("CAN", { { 'a', "1" } });
+    else
+        record.addSubfieldCreateFieldUnique("SUB", 'a', "ixtheo");
```
mtrojan-ub, line 64: Should be `IXT` rather than `ixtheo`

cpp/lib/src/MARC.cc
```diff
@@ -1592,6 +1592,31 @@ bool Record::addSubfield(const Tag &field_tag, const char subfield_code, const s
...
 }
 
+bool Record::addSubfieldCreateFieldUnique(const Tag &field_tag, const char subfield_code, const std::string &subfield_value) {
+    const auto &field(findTag(field_tag));
+    bool success = true;
+    if (field == fields_.end()) {
+        success = insertField(field_tag, std::string(1, subfield_code) + subfield_value);
+    }
+    if (success) {
+        if (not this->hasSubfieldWithValue(field_tag, subfield_code, subfield_value)) {
```
mtrojan-ub, line 1602: 'this->' really needed? we normally dont use it

cpp/lib/include/MARC.h
```diff
@@ -752,6 +752,15 @@ class Record {
...
     bool addSubfield(const Tag &field_tag, const char subfield_code, const std::string &subfield_value);
 
+    /** \brief  Adds a subfield to the first existing field with tag "field_tag". Field is created if it does not exist.
+     *  \return True if a subfield and optionally field created or if subfield with value is already present.
+     */
+    bool addSubfieldCreateFieldUnique(const Tag &field_tag, const char subfield_code, const std::string &subfield_value);
+
+    /** \brief  Checks if a specified field exists which additionally contains a subfield with a given value.
+     */
+    bool hasSubfieldWithValue(const Tag &field_tag, const char subfield_code, const std::string &subfield_value) const;
```
mtrojan-ub, line 762: a better name would be `hasFieldWithSubfieldValue`

cpp/lib/src/MARC.cc
```diff
@@ -1592,6 +1592,31 @@ bool Record::addSubfield(const Tag &field_tag, const char subfield_code, const s
...
+            success = addSubfield(field_tag, subfield_code, subfield_value);
+        }
+    }
+    return success;
+}
+
+
+bool Record::hasSubfieldWithValue(const Tag &field_tag, const char subfield_code, const std::string &subfield_value) const {
+    auto field(findTag(field_tag));
+    while (field != fields_.cend()) {
```
mtrojan-ub, line 1612: please also check if the tag is still the same (fields_ contains all fields, not only the one with the tag you are looking for).

cpp/lib/src/MARC.cc
```diff
@@ -1592,6 +1592,31 @@ bool Record::addSubfield(const Tag &field_tag, const char subfield_code, const s
     return true;
 }
 
+bool Record::addSubfieldCreateFieldUnique(const Tag &field_tag, const char subfield_code, const std::string &subfield_value) {
+    const auto &field(findTag(field_tag));
+    bool success = true;
+    if (field == fields_.end()) {
+        success = insertField(field_tag, std::string(1, subfield_code) + subfield_value);
```
mtrojan-ub, line 1599: please avoid passing a string directly (e.g. due to missing indicators)

`insertField(field_tag, {{subfield_code, subfield_value}})`

PR #4089, opened 2021-10-12
Kimdok CORE Updates

ruschein:


cpp/core/convert_core_json_to_marc.cc
```diff
@@ -0,0 +1,340 @@
...
+    else if (item_type == "Article")
+        return MARC::Record::SERIAL_COMPONENT_PART;
+    else {
+        LOG_WARNING("unkown item type: " + item_type);
+        return MARC::Record::UNDEFINED;
+    }
+}
+
+
+std::string NormalizeAuthorName(const std::string &author_name) {
```
mtrojan-ub, line 123: Can we re-use the Zotaut logic for normalizing author names here?

ruschein: We also have NormaliseAuthorName in ControlNumberGuesser.cc which does some of the same work and some additional stuff.  Part of that could also be extracted into a common function.

cpp/data/system_updates/20.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
mtrojan-ub, line 1: With the introduction of the new system_updates schema, scripts with ids <= 99 are no longer allowed. Please rename your script to `100.sh`

mtrojan-ub:
You should also add a .gitignore file in cpp/core and register all the binaries there.

ruschein:
Done!


--
Dr. Johannes Ruscheinski
Universitätsbibliothek Tübingen - IT-Abteilung -
Wilhelmstr. 32, 72074 Tübingen

Tel: +49 7071 29-72820
FAX: +49 7071 29-5069
Email: ***@***.***

 The Sophisticate:  "The world isn't black and white.  No one does pure good or pure bad. It's all gray.  Therefore, no one is better than anyone else."
    The Zetet:  "Knowing only gray, you conclude that all grays are the same shade.  You mock the simplicity of the two-color view, yet you replace it with a one-color view..."
      —Marc Stiegler, David's Sling


________________________________

mtrojan-ub:
Sorry, but due to the fact that we merged another PR in the meantime, you should rename it to 102.sh instead.

ruschein:
> Sorry, but due to the fact that we merged another PR in the meantime, you should rename it to 102.sh instead.

Done

mtrojan-ub:
If you would like to be merged soon, could you please find a better title for the PR that "work in progress"? (Please use a title that has actually something to do with the content at all...)

cpp/data/system_updates/102.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -o errexit
+
+
+if [[ $(hostname) != "ub16.uni-tuebingen.de" ]]; then
```
mtrojan-ub, line 5: If you want this to happen on ub16, it must also happen on its successor ub55.v264.uni-tuebingen.de.

And => is there any specific reason why we do not want the cronjob on the test servers? Why? (its just generating a list from Zeder anyway)

cpp/data/installer/ixtheo.cronjobs
```diff
@@ -45,3 +45,4 @@ LOG_DIR=/usr/local/var/log/tuefind
 45 23 * * * "$BIN/rss_subset_aggregator" "--mode=email" "$EMAIL" "relbib" > "$LOG_DIR/rss_subset_aggregator_relbib.log" 2>&1
 30 22 * * 0 "$BIN/generate_kalliope_originators.py" > "${LOG_DIR}/generate_kalliope_originators.py" 2>&1'
 {IF DEFINED(production)}0 0 1 * * "$BIN/cologne.sh"{ENDIF}
+{IF DEFINED(production)}0 1 1 * * "$BIN/update_print_issns_titles_online_ppns_and_online_issns_list.sh"{ENDIF}
```
mtrojan-ub, line 48: Here also => why do we want this only in production?

cpp/core/convert_core_json_to_marc.cc
```diff
@@ -0,0 +1,386 @@
...
+        : journal_title_(journal_title), ppn_(ppn), online_issn_(online_issn) { }
+};
+
+
+// Parses an input file that has three (the last component may be empty) parts per line that are colon-separated.
+// Embedded colons may be backslash escaped.
+void LoadISSNsToJournalTitlesPPNsAndISSNsMap(
+    std::unordered_map<std::string, JournalTitlePPNAndOnlineISSN> * const issns_to_journal_titles_ppns_and_issns_map)
+{
+    const std::string MAP_FILE_PATH(UBTools::GetTuelibPath() + "print_issns_titles_online_ppns_and_online_issns.csv");
```
mtrojan-ub, line 70: If we extend the list or change the row order later we must rename it, maybe also write a migration script and so on... And while reading this file name, when i reach the last part i already forgot what the first part was. Why not simply call the file journals.csv?

mtrojan-ub: And why is the column print issns not directly next to the column online issns?

cpp/core/convert_core_json_to_marc.cc
```diff
@@ -0,0 +1,386 @@
...
+                              const std::unordered_map<std::string, unsigned> &unmatched_issns_to_counts_map)
+{
+    for (const auto &issn_and_count : unmatched_issns_to_counts_map)
+        (*unmatched_issns_file) << issn_and_count.first << '\t' << issn_and_count.second << '\n';
+    LOG_INFO("Wrote a list of " + std::to_string(unmatched_issns_to_counts_map.size()) + " unmapped ISSN's to \""
+             + unmatched_issns_file->getPath() + "\".");
+}
+
+
+const std::string UNIQUE_ID_TO_DATE_MAP_PATH(UBTools::GetTuelibPath() + "convert_json_to_marc.db");
```
mtrojan-ub, line 323: This should also be renamed to convert_core_json_to_marc.db

cpp/core/convert_core_json_to_marc.cc
```diff
@@ -0,0 +1,386 @@
...
+    record->insertField("041", 'a', MARC::MapToMARCLanguageCode(language_object->getStringNode("code")->getValue()));
+}
+
+
+// \return True if an abstract was found, else false.
+bool ProcessAbstract(const JSON::ObjectNode &entry_object, MARC::Record * const record) {
+    if (not entry_object.hasNode("abstract") or entry_object.isNullNode("abstract"))
+        return false;
+    const auto abstract_node(entry_object.getStringNode("abstract"));
+    record->insertField("520", 'a', abstract_node->getValue());
```
mtrojan-ub, line 170: Should this be shortened if too long?

ruschein:
> If you would like to be merged soon, could you please find a better title for the PR that "work in progress"? (Please use a title that has actually something to do with the content at all...)

Done!

cpp/core/convert_core_json_to_marc.cc
ruschein: Done!

cpp/core/convert_core_json_to_marc.cc
ruschein: Yes, and done!

cpp/data/system_updates/102.sh
ruschein: I think we currently only need this on Krimdok systems and I have adjusted the script accordingly.

mtrojan-ub: OK but if we need it for krimdok, then why did you use ub16 before and not ub15? If you change this now, it is also necessary to move the cronjob installer definition from ixtheo.cronjobs to krimdok.cronjobs.

ruschein:
Moved.  Maybe it was because of lack of sleep?


--
Dr. Johannes Ruscheinski
Universitätsbibliothek Tübingen - IT-Abteilung -
Wilhelmstr. 32, 72074 Tübingen

Tel: +49 7071 29-72820
FAX: +49 7071 29-5069
Email: ***@***.***

 The Sophisticate:  "The world isn't black and white.  No one does pure good or pure bad. It's all gray.  Therefore, no one is better than anyone else."
    The Zetet:  "Knowing only gray, you conclude that all grays are the same shade.  You mock the simplicity of the two-color view, yet you replace it with a one-color view..."
      —Marc Stiegler, David's Sling


________________________________

mtrojan-ub:
Another PR has been merged in the meantime: #4132
Could you please `git pull origin master`? Please remember that the "core" directory has been renamed to "harvest" as discussed, so that change should be reflected within this PR as well.

mtrojan-ub:
By the way: since we only have very short time left before the holiday season, and our live servers will be updated directly in the first week after it when we're all back (on 12th january), and this PR will change some cronjobs as well, i think it might be better to leave this one open until the live servers have been updated.

git-config/hooks/pre-commit
```diff
@@ -2,8 +2,8 @@
 set -o errexit -o nounset
 
 
-forbidden_hosts=("ptah" "sobek" "nu" "ub16" "ub15" "ub27" "ub28")
-short_host=$(hostname --short)
+readonly forbidden_hosts=("ptah" "sobek" "nu" "ub16" "ub15" "ub27" "ub28")
```
mtrojan-ub, line 5: by the way, before we forget it: can you please remove ub27 (no ub_tools installation there) and add ub53 ub54 ub55? These will be our Ubuntu Server replacements for ub28, ub15+ub16 (please leave the old ones in there for now.)

git-config/hooks/pre-commit
```diff
@@ -18,4 +18,13 @@ done
 pyflakes3 /usr/local/ub_tools/cronjobs
 
 
+# Format C++ source files:
+readonly changed_files=$(git diff --cached --name-only --diff-filter=ACM)
+readonly GIT_ROOT=$(git rev-parse --show-toplevel)
+for changed_file in changed_files; do
+    # "-style=file" has the effect of searching up the directory tree until a file named ".clang-format" is found and then using it.
+    clang-format-12 -style=file -i "$GIT_ROOT/$changed_file"
```
mtrojan-ub, line 26: maybe we should leave this inactive for now so we have a chance to finally test your changes before enabling it in the hook.

git-config/hooks/pre-commit
```diff
@@ -18,4 +18,13 @@ done
 pyflakes3 /usr/local/ub_tools/cronjobs
 
 
+# Format C++ source files:
+readonly changed_files=$(git diff --cached --name-only --diff-filter=ACM)
```
mtrojan-ub, line 22: Can we limit this to .cc and .h files somehow? e.g. we do not want this to be applied to .py files (and others).

ruschein: Done!

git-config/hooks/pre-commit
mtrojan-ub: or maybe we just remove this section entirely... since mid 2021, you can only push with a ssh key which will not be installed on the machines anyway.

git-config/hooks/pre-commit
mtrojan-ub: One more thing: Can we also exclude external libraries like libstemmer?

git-config/hooks/pre-commit
ruschein: So, should I remove it?

git-config/hooks/pre-commit
ruschein: Done!

cpp/harvest/.gitignore
```diff
@@ -1 +1 @@
-nacjd
+convert_core_json_to_marc
```
mtrojan-ub, line 1: nacjd should stay in there too

cpp/data/installer/scripts/install_ubuntu_packages.sh
```diff
@@ -29,7 +29,7 @@ apt-add-repository --yes --update 'deb https://artifacts.elastic.co/packages/7.x
 
 # main installation
 apt-get --quiet --yes --allow-unauthenticated install \
-        ant apache2 apparmor-utils ca-certificates cifs-utils clang cron curl gcc git imagemagick incron jq libarchive-dev \
+        ant apache2 apparmor-utils ca-certificates cifs-utils clang clang-format-12 cron curl gcc git imagemagick incron jq libarchive-dev \
```
mtrojan-ub, line 32: If we add a package, we should also add a system_update that installs clang-format-12 on all existing ubuntu machines.

cpp/data/system_updates/102.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -o errexit
+
+
+if [[ $TUEFIND_FLAVOUR != "krimdok" ]]; then
```
mtrojan-ub, line 5: If it is not krimdok, then it can either be ixtheo (like ptah/ub16) or ub_tools_only (like nu/ub28). Do you really want it to run on both machine types?

cpp/marc_tools/convert_marc_to_utf8.cc
```diff
@@ -0,0 +1,97 @@
...
+#include "MARC.h"
+#include "TextUtil.h"
+#include "util.h"
+
+
+namespace {
+
+
+[[noreturn]] void Usage() {
+    ::Usage("(--all-fields|--non-utf8-fields-only) from_encoding marc_input marc_output\n"
```
mtrojan-ub, line 30: can we also add a parameter `to_encoding` instead of hardcoding `utf8` everywhere? As far as i can see, `EncodingConverter` already supports `to_encoding` anyway. Of course this means that `--non-utf8-fieds-only` needs to be renamed too, as well as the script name being changed to something more generic.

ruschein: No because the MARC-21 standard only supports two encodings, UTF8 and the proprietary MARC-8.

cpp/harvest/convert_core_json_to_marc.cc
```diff
@@ -0,0 +1,386 @@
...
+// Embedded colons may be backslash escaped.
+void LoadISSNsToJournalTitlesPPNsAndISSNsMap(
+    std::unordered_map<std::string, JournalTitlePPNAndOnlineISSN> * const issns_to_journal_titles_ppns_and_issns_map)
+{
+    const std::string MAP_FILE_PATH(UBTools::GetTuelibPath() + "journals.csv");
+    std::vector<std::vector<std::string>> lines;
+    TextUtil::ParseCSVFileOrDie(MAP_FILE_PATH, &lines);
+
+    for (const auto &line : lines) {
+        if (lines[0].empty() or line[1].empty())
```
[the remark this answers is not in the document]
ruschein: Fixed!

cpp/data/installer/scripts/install_ubuntu_packages.sh
ruschein: Done!

git-config/hooks/pre-commit
ruschein: It's not external any more as we adopted it.

cpp/data/system_updates/103.sh
```diff
@@ -0,0 +1,16 @@
...
+
+
+if [[ $TUEFIND_FLAVOUR != "krimdok" ]]; then
+    exit 0
+fi
+
+if [ -e /var/spool/cron/root ]; then # CentOS
+    root_crontab=/var/spool/cron/root
+else # Ubuntu
+    root_crontab=/var/spool/cron/crontabs/root
```
mtrojan-ub, line 12: Our development systems might not have a root crontab, therefore the file will not exist and then sed might fail. See 101.sh for a good example.

cpp/data/system_updates/103.sh
```diff
@@ -0,0 +1,16 @@
...
+fi
+
+sed --in-place '/^# END VUFIND AUTOGENERATED/i 0 1 1 * * "$BIN/update_print_issns_titles_online_ppns_and_online_issns_list.sh"' \
```
mtrojan-ub, line 15: 1) This is not the same command as in the krimdok.cronjobs installer file (log file missing).
2) It should be inserted at the same position as in krimdok.cronjobs (between expire_searches and purge_old_data), and not before the END VUFIND AUTOGENERATED section

cpp/data/system_updates/103.sh
ruschein: Done

git-config/hooks/pre-commit
mtrojan-ub: If we change the code style that also means it will be harder to diff against the original files in the future in case we want to apply updates without losing our changes

mtrojan-ub: But i guess since we will never do that it should be ok anyway

PR #4091, opened 2021-10-13
count_author_gnd_refs: Only count distinct references (1 per record)

mtrojan-ub:


cpp/count_author_gnd_refs.cc
```diff
@@ -78,16 +80,20 @@ void ProcessRecords(MARC::Reader * const marc_reader, const std::unordered_set<s
                     const std::string gnd_number(subfield0.substr(__builtin_strlen("(DE-588)")));
                     const auto gnd_number_and_count(gnd_numbers_and_counts->find(gnd_number));
                     if (gnd_number_and_count != gnd_numbers_and_counts->end()) {
-                        ++gnd_number_and_count->second;
-                        ++matched_count;
+                        if (processed_gnd_numbers.find(gnd_number) == processed_gnd_numbers.end()) {
+                            ++gnd_number_and_count->second;
+                            ++distinct_matched_count;
```
andreas-ub, line 85: variable prob. not needed if it equals size of processed_gnd_numbers

mtrojan-ub: These variables persist over all records, but processed_gnd_numbers is reset for each record.

PR #4100, opened 2021-10-26
refactoring of system updater and sql updater

andreas-ub:


cpp/system_updater.cc
```diff
@@ -84,16 +123,31 @@ int Main(int argc, char *argv[]) {
...
+        }
+        else if (script_name.ends_with(".sql")) {
+            if (dry_run)
+                std::cout << SYSTEM_UPDATES_DIR << " " << script_name << std::endl;
+            else
+                ApplyDatabaseUpdate(&db_connection, SYSTEM_UPDATES_DIR, script_name, &current_schema, last_schema);
+            last_schema = current_schema;
+        }
+        else
+            continue;
```
mtrojan-ub, line 144: I think this should cause an error including a message, instead of continuing and just skipping it silently.

mtrojan-ub:
As discussed, we made sure manually that all machines are on the latest sql_updates + system_updates level.

The following changes should be made before this PR can be merged:
- All files sql_updates/* and system_updates/* should be moved to a new folder system_updates.old/*, which is only for manual use + later examples. All files in the .old folder will no longer be maintained via the system_updater. (sql_updates directory removed, but sql_updates should be kept / which is empty for now until the first )
- installer + system_updater should set the version in /usr/local/var/lib/tuelib/system_version to 99, if they detect a lower number (or no updates present yet)
- system_updater should skip updates with version numbers < 99 with a warning that these might belong to the old update schema and are no longer supported

PR #4101, opened 2021-10-26
Riedl email crawl

jriedl:


cpp/lib/src/StringUtil.cc
```diff
@@ -2851,4 +2851,24 @@ std::string GenerateRandom(const size_t length, const std::string &alphabet) {
 }
 
 
+std::vector<std::string> SplitIntoLines(const std::string &s) {
```
mtrojan-ub, line 2854: This looks the same as `StringUtil::Split(s, '\n', &lines);`

PR #4112, opened 2021-11-11
Riedl

jriedl:


cpp/pipelines/augment_mohr_titles_with_dois.cc
```diff
@@ -0,0 +1,160 @@
...
+                     // Insert doi if not present
+                     if (std::find(record_dois.begin(), record_dois.end(), doi_for_isbn) == record_dois.end()) {
+                         LOG_INFO("Inserting previously not existing DOI \"" + doi_for_isbn + "\" for electronic record \
+                                   \"" + record.getControlNumber());
+                         record.insertField("024", { { 'a', doi_for_isbn }, { '2', "doi" } });
+                     }
+                     break;
+                 }
+             }
+             marc_writer->write(record);
```
andreas-ub, line 100: maybe only one write at the end of the while loop

PR #4120, opened 2021-11-22
added check for identical script numbers

andreas-ub:


cpp/system_updater.cc
```diff
@@ -131,9 +131,17 @@ int Main(int argc, char *argv[]) {
 
     std::sort(script_names.begin(), script_names.end(), ScriptLessThan);
     std::string last_schema, current_schema;
+    int last_version_number = -1;
```
mtrojan-ub, line 134: `previous_version_number` should be a better variable name than `last_version_number`

cpp/system_updater.cc
```diff
@@ -131,9 +131,17 @@ int Main(int argc, char *argv[]) {
 
     std::sort(script_names.begin(), script_names.end(), ScriptLessThan);
     std::string last_schema, current_schema;
+    int previous_version_number = -1;
 
     for (const auto &script_name : script_names) {
         LOG_INFO("Running " + script_name);
+        const unsigned version_number(GetVersionFromScriptName(script_name));
+
+        if (previous_version_number == (int) version_number)
```
mtrojan-ub, line 140: Please use StringUtil::ToInt instead

ruschein: I don't think using StringUtil::ToInt will work as that is for converting std::string to int!  What would be better is to use static_cast but that will have the same effect.

PR #4124, opened 2021-11-24
docker: Shared Makefile.inc for ixtheo-ubuntu & krimdok-ubuntu (+reduced amount of data copied from /mnt)

mtrojan-ub:
... this way we save about 10 GB of data which no longer need to be copied each time we build a docker container.

### DIFF
--- a/docker/Makefile.inc
+++ b/docker/Makefile.inc
@@ -7,7 +7,6 @@ ubuntu_prepare_deps:
 	if [ -z "$(docker images -q ub_tools-deps-ubuntu)" ]; then \
 		cd /usr/local/ub_tools/docker/ub_tools-deps-ubuntu; \
 		make ; \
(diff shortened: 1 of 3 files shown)


docker/Makefile.inc
```diff
@@ -0,0 +1,28 @@
+ubuntu_clean_deps:
+	docker image rm -f ub_tools-deps-ubuntu
+	rm -rf mnt/ZE020150
+	rm -f .smbcredentials
+
+ubuntu_prepare_deps:
+	if [ -z "$(docker images -q ub_tools-deps-ubuntu)" ]; then \
+		cd /usr/local/ub_tools/docker/ub_tools-deps-ubuntu; \
+		make ; \
+		cd /usr/local/ub_tools/docker/ixtheo-ubuntu; \
```
andreas-ub, line 10: not sure about ixtheo here, because inc file is included in krimdok-ubuntu

PR #4135, opened 2021-12-16
Zeder: UploadArticleList

mtrojan-ub:
This is just the library function. Calling it inside the cronjob can be done later.


cpp/lib/include/Zeder.h
```diff
@@ -473,5 +473,10 @@ class SimpleZeder {
     inline const_iterator end() const { return entries_.end(); }
 };
```
andreas-ub, line 475: Please add one more blank line :-D

mtrojan-ub: Well, if it's our standard, that means that's also stringent for me, right ;)
Done, see #4137

PR #4139, opened 2021-12-16
Riedl

jriedl:


cpp/lib/src/ZoteroHarvesterConversion.cc
```diff
@@ -1439,7 +1453,7 @@ void ConversionTasklet::run(const ConversionParams &parameters, ConversionResult
             AugmentMetadataRecord(&new_metadata_record, parameters);
 
             LOG_DEBUG("Augmented metadata record: " + new_metadata_record.toString());
-            if (new_metadata_record.url_.empty())
+            if (new_metadata_record.url_.empty() and new_metadata_record.doi_.empty())
```
mtrojan-ub, line 1456: If you change the condition, please also change the error message in the exception.

cpp/lib/src/ZoteroHarvesterConfig.cc
```diff
@@ -542,11 +542,13 @@ bool ParseExpectedLanguages(const std::string &expected_languages_string, Langua
     std::string expected_languages(expected_languages_string);
     if (expected_languages[0] == '*')
         return false;
-    else if (expected_languages[0] == '!' or expected_languages[0] == '?') {
+    else if (expected_languages[0] == '!' or expected_languages[0] == '?' or expected_languages[0] == '=') {
```
mtrojan-ub, line 545: If you add a new mode here, please don't forget to add it to the documentation section in zotero_harvester.conf after the PR is merged.

PR #4152, opened 2022-01-27
added beacon subfolder and beacon ADB/NDB parsing

andreas-ub:


cpp/update_beacon_files.sh
```diff
@@ -38,6 +42,24 @@ else
     error_message+=$'Failed to download the Beacon file from Kalliope.\n'
 fi
 
+# additional beacon files for external references like wikipedia, e.g. ADB/NDB
+declare -a beacon_files=("https://www.historische-kommission-muenchen-editionen.de/beacon_db_register.txt")
```
mtrojan-ub, line 46: You could use an associative array instead, which contains the URL and the target file name. Then we could also add kalliope and archivportal-d to the same array and keep a single function. I also think that the file name of the 3rd one should be changed, because "beacon_db_register" doesn't tell you on first sight that it comes from adb-ndb or historische-kommission-muenchen-editionen.de or something like that.

cpp/pipelines/ixtheo_marc_pipeline_fifo.sh
```diff
@@ -153,8 +153,8 @@ wait
 
 StartPhase "Add BEACON Information to Authority Data"
 (add_authority_beacon_information Normdaten-partially-augmented1-"${date}".mrc \
-                                  Normdaten-partially-augmented2-"${date}".mrc kalliope.staatsbibliothek-berlin.beacon \
-                                  --type-file kalliope_originators.txt $(find . -name '*.beacon' ! -name "*kalliope.*") \
+                                  Normdaten-partially-augmented2-"${date}".mrc beacon_downloads/kalliope.staatsbibliothek-berlin.beacon \
```
mtrojan-ub, line 156: Shouldn't there also be an archivportal-d reference somewhere in the ixtheo pipeline file that needs to be changed as well?

PR #4157, opened 2022-02-01
changes to translation tool

andreas-ub:


cpp/cgi_progs/translator.cc
```diff
@@ -131,7 +132,7 @@ std::string CreateEditableRowEntry(const std::string &token, const std::string &
 
 
 void GetDisplayLanguages(std::vector<std::string> *const display_languages, const std::vector<std::string> &translation_languages,
-                         const std::vector<std::string> &additional_view_languages, enum Category category = KEYWORDS) {
+                         const std::vector<std::string> &additional_view_languages, enum Category category, const bool show_macs_col) {
```
jriedl, line 135: Default category should remain.
show_macs_col should also have a standard value.

cpp/cgi_progs/translator.cc
```diff
@@ -244,6 +247,14 @@ bool IsEmptyEntryWithoutTranslator(const std::string &entry) {
 }
 
 
+bool IsMacsColumnVisible(const IniFile &ini_file) {
+    const std::string ini_macs_visible(ini_file.getString(CONFIGURATION_SECTION, "show_macs_col"));
```
jriedl, line 251: Please use hasSection()  (cf. IniFile.h)

cpp/cgi_progs/translator.cc
```diff
@@ -244,6 +247,14 @@ bool IsEmptyEntryWithoutTranslator(const std::string &entry) {
...
+    if (ini_macs_visible.empty())
+        return false;
+    return StringUtil::FindCaseInsensitive(ini_macs_visible, "true") != std::string::npos;
```
jriedl, line 254: Please use getBool() (c.f. IniFile.h)

cpp/cgi_progs/translator.cc
```diff
@@ -534,8 +548,9 @@ void GetKeyWordTranslationsAsHTMLRowsFromDatabase(DbConnection &db_connection, c
                if (language_code == "ger")
                    row_values[index] = (german_updated == "1") ? CreateNonEditableUpdateHintEntry(translation, gnd_code)
                                                                : CreateNonEditableHintEntry(translation, gnd_code);
-               else if (status == "reliable")
-                   row_values[index] = CreateNonEditableRowEntry(translation);
+               // 20220131: for community version changes in "final" translations shall be possible
+               // else if (status == "reliable")
+               //     row_values[index] = CreateNonEditableRowEntry(translation);
```
jriedl, line 553: Think we should keep this case explicitly despite the comment, i.e.
```
else if (status == "reliable")
      row_values[index] = CreateEditableRowEntry(translation);
```

PR #4169, opened 2022-02-15
changes in harvester for core: remove zeder, processing in loop etc.

andreas-ub:


cpp/harvest/convert_core_json_to_marc.cc
```diff
@@ -286,12 +215,16 @@ void GenerateMARCFromJSON(const JSON::ArrayNode &root_array,
                 ++skipped_incomplete_count;
                 continue;
             }
-            ProcessContributors(*entry_object, &new_record, authors);
+
+            //Do not use contributors anymore (team decission in video conf. on 09.02.2022)
```
mtrojan-ub, line 219: decision

cpp/harvest/convert_core_json_to_marc.cc
```diff
@@ -286,12 +215,16 @@ void GenerateMARCFromJSON(const JSON::ArrayNode &root_array,
...
+            //ProcessContributors(*entry_object, &new_record, authors);
+
             if (not ProcessTitle(*entry_object, &new_record)) {
                 ++skipped_incomplete_count;
                 continue;
             }
             new_record.insertControlField("007", "cr||||");
+            new_record.insertField("084", { { 'a', "2,1" }, { '2', "ssgn" } });
```
mtrojan-ub, line 227: Please preserve an ascending field sequence (084 behind 035)

mtrojan-ub:
Still lots of hardcoded stuff in here, but that's also due to the fact that the exact definitions aren't final yet. I think we can leave this open a bit further until things get clearer.

cpp/rss_aggregator.cc
```diff
@@ -191,7 +193,7 @@ void PerformDescriptionSubstitutions(const std::string &patterns_and_replacement
 
 
 // \return the number of new items.
-unsigned ProcessFeed(const std::string &feed_id, const std::string &feed_name, const std::string &feed_url,
+int ProcessFeed(const std::string &feed_id, const std::string &feed_name, const std::string &feed_url,
```
mtrojan-ub, line 196: If the return type is changed, the comment above this line should be changed as well.

cpp/rss_aggregator.cc
```diff
@@ -274,15 +280,19 @@ constexpr unsigned SECONDS_TO_MILLISECONDS = 1000;
...
     auto result_set(db_connection->getLastResultSet());
     while (const auto row = result_set.getNextRow()) {
         LOG_INFO("Processing feed \"" + row["feed_name"] + "\".");
-        const unsigned new_item_count(ProcessFeed(row["id"], row["feed_name"], row["feed_url"], row.getValue("title_suppression_regex"),
+        const int new_item_count(ProcessFeed(row["id"], row["feed_name"], row["feed_url"], row.getValue("title_suppression_regex"),
                                                   row.getValue("descriptions_and_substitutions"), row.getValue("strptime_format"),
                                                   downloader, db_connection,
                                                   StringUtil::ToUnsigned(row["downloader_time_limit"]) * SECONDS_TO_MILLISECONDS));
-        LOG_INFO("Downloaded " + std::to_string(new_item_count) + " new items.");
+        if (new_item_count == ERROR_DOWNLOAD or new_item_count == ERROR_PARSING)
```
mtrojan-ub, line 292: If there is no difference in handling the specific errors, we could also compare it against < 0 (this would be more robust in case we add more error types later).

cpp/rss_aggregator.cc
```diff
@@ -321,15 +331,25 @@ int Main(int argc, char *argv[]) {
...
+        if (number_feeds_with_error > 0) {
+            const auto subject(program_basename + " on " + DnsUtil::GetHostname() + " (subsystem_type: " + subsystem_type + ")");
+            const auto message_body("number of feeds that could not be downloaded: " + std::to_string(number_feeds_with_error));
+            if (EmailSender::SimplerSendEmail("no_reply@ub.uni-tuebingen.de", { email_address }, subject, message_body, EmailSender::VERY_HIGH)
+                < 299)
+                return EXIT_FAILURE;
+            else
+                LOG_ERROR("failed to send an email error report!");
+        }
+        return number_feeds_with_error;
```
mtrojan-ub, line 351: Since we're in the Main() function, we should return EXIT_SUCCESS here instead.

PR #4179, opened 2022-03-01
core language extraction, extension in RangeUtil

andreas-ub:


cpp/lib/src/RangeUtil.cc
```diff
@@ -763,6 +747,8 @@ static void ReplaceXs(std::string * const time_range_candidate) {
             range_end = "30" + range_end.substr(2);
     }
 
+    range_start_plus_dash = StringUtil::ReplaceString("X", "0", range_start_plus_dash);
```
mtrojan-ub, line 750: This might work for years, but what would happen with a full date here?
XX.XX.199X

=> would either be changed to 00.00.1990 or 99.99.1999, which is of course invalid in both cases

andreas-ub: should already been replaced in a step before (at line 712), so it would be 01.01.1990
or to 31.12.1999 (at line 730), if it is after the '-'

PR #4184, opened 2022-03-10
restructured time range util

andreas-ub:


cpp/lib/src/RangeUtil.cc
```diff
@@ -756,8 +756,12 @@ static void ReplaceXs(std::string * const time_range_candidate) {
 bool ConvertTextToTimeRange(std::string text, std::string * const range, const bool special_case_centuries) {
     if (StringUtil::StartsWith(text, "ca. "))
         text = text.substr(4);
+    if (StringUtil::Contains(text, "-ca. "))
+        text = StringUtil::ReplaceString("-ca. ","", text);
```
mtrojan-ub, line 760: shouldn't the `-` be kept?

PR #4186, opened 2022-03-11
extended remove dangling refs to exclude records published in future

andreas-ub:


cpp/pipelines/remove_dangling_references.cc
```diff
@@ -52,7 +63,7 @@ void EliminateDanglingCrossReferences(MARC::Reader * const reader, MARC::Writer
                 for (const auto &subfield : field->getSubfields()) {
                     if (subfield.code_ == 'w' and StringUtil::StartsWith(subfield.value_, "(DE-627)")) {
                         const auto bsz_ppn(subfield.value_.substr(__builtin_strlen("(DE-627)")));
-                        if (unlikely(all_ppns.find(bsz_ppn) == all_ppns.cend())) {
+                        if (unlikely(all_ppns.find(bsz_ppn) == all_ppns.cend()) or all_ppns.find(bsz_ppn)->second == true) {
```
mtrojan-ub, line 66: If you would like to use the result of the find() operation more than once, please use a variable to avoid duplicate search operations.

cpp/pipelines/remove_dangling_references.cc
```diff
@@ -83,7 +98,7 @@ int Main(int argc, char *argv[]) {
     const auto marc_writer(MARC::Writer::Factory(argv[2]));
     const auto log_file(FileUtil::OpenOutputFileOrDie(argv[3]));
 
-    std::unordered_set<std::string> all_ppns;
+    std::unordered_map<std::string, bool> all_ppns;
```
mtrojan-ub, line 101: Since it's not only a set or vector of ppns, the name of the variable should also contain something about the meaning of the bool flag.

cpp/pipelines/remove_dangling_references.cc
```diff
@@ -35,15 +36,25 @@ namespace {
 }
 
 
-void CollectAllPPNs(MARC::Reader * const reader, std::unordered_set<std::string> * const all_ppns) {
-    while (const auto record = reader->read())
-        all_ppns->emplace(record.getControlNumber());
+void CollectAllPPNs(MARC::Reader * const reader, std::unordered_map<std::string, bool /*mark to remove record*/> * const all_ppns) {
+    int next_year = std::stoi(TimeUtil::GetCurrentYear()) + 1;
```
mtrojan-ub, line 40: This could be const

cpp/pipelines/remove_dangling_references.cc
mtrojan-ub: Also: Didn't we want to have a different kind of logging if the record is skipped on purpose?

PR #4187, opened 2022-03-14
continued restructuring of sqlite db for control number guesser

andreas-ub:


cpp/lib/include/ControlNumberGuesser.h
```diff
@@ -62,12 +62,12 @@ class ControlNumberGuesser {
                                                    const std::string &year = "", const std::set<std::string> &dois = {},
                                                    const std::set<std::string> &issns = {}, const std::set<std::string> &isbns = {}) const;
 
-    bool getNextTitle(std::string * const title, std::set<std::string> * const control_numbers) const;
-    bool getNextAuthor(std::string * const author_name, std::set<std::string> * const control_numbers) const;
+    void getControlnumbersAndAuthors(std::unordered_map<std::string, std::set<std::string>> * const control_numbers_to_author_map);
```
mtrojan-ub, line 65: We should not create multiple upper/lowercase variants for the same string, so please always use `ControlNumber` instead of `Controlnumber` similar to the existing examples.

PR #4210, opened 2022-04-11
changes for core due to preparation of new api key

andreas-ub:


cpp/data/installer/ixtheo.sql
```diff
@@ -1,6 +1,4 @@
--- under CentOS we still have MariaDB 5, which has a limitation of 767 bytes for keys.
--- this means e.g. for VARCHAR with utf8mb4, we can use at most a VARCHAR(191)!
--- Also, the default collating sequence is a Swedish one.  This leads to aliasing problems for characters with
+-- The default collating sequence is a Swedish one.  This leads to aliasing problems for characters with
```
mtrojan-ub, line 1: If you change the comment here please also change ub_tools.sql

cpp/data/installer/scripts/install_ubtools.sh
```diff
@@ -15,39 +24,27 @@ if [[ ! $PATH =~ "/usr/local/bin" ]]; then
...
-    echo "OS type could not be detected! aborting"
-    exit 1
+if [ -n "$BRANCH" ]; then
+    BRANCH="master"
+fi
+echo "Branch is: ${BRANCH}"
+
+echo "Installing dependencies..."
+
+cd /tmp
```
mtrojan-ub, line 34: If we do a `cd` here, we should switch back to the previous directory after the actions in `/tmp` are finished.

mtrojan-ub: Either that or => write the full path /tmp/install_ubuntu_packages to a variable and re-use it in all of the calls

PR #4236, opened 2022-06-22
NACJD changes

mtrojan-ub:


cpp/harvest/nacjd.cc
```diff
@@ -153,15 +153,15 @@ void ExtractExistingIDsFromMarc(MARC::Reader * const marc_reader, std::set<std::
             StringUtil::ReplaceString("[ICPSR]", "", &id_035);
             StringUtil::TrimWhite(&id_035);
             parsed_marc_ids->emplace(id_035);
-        }
+        }
```
mtrojan-ub, line 156: This whitespace change looks strange... i would suppose that our pre-commit hook should automatically use clang-format to normalize this, but for some reason it failed. I will investigate.

cpp/harvest/nacjd.cc
```diff
@@ -343,9 +353,15 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
 
 
 int Main(int argc, char *argv[]) {
-    if (argc != 3)
+    if (argc < 3 or argc > 4)
         Usage();
 
+ 
+    int count_rows = 9;
```
mtrojan-ub, line 360: 1) before the change, count_rows has been hardcoded with the value 9000, so we might want to keep that value
2) Please always use the constructor notation when initializing variables: `int count_rows(9);`

cpp/harvest/nacjd.cc
```diff
@@ -313,14 +320,19 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
...
                 new_record.insertField("264", { { 'c', initial_release_date } });
                 new_record.insertField("520", { { 'a', description } });
                 new_record.insertField("540", { { 'a', license } });
+                new_record.insertField("655", { { 'a', "Forschungsdaten" } }, ' ', '4');
                 new_record.insertField("852", { { 'a', "DE-2619" } });
-                new_record.insertField("856", { { 'u', "https://www.icpsr.umich.edu/web/NACJD/studies/" + UrlUtil::UrlEncode(id) } }, '4' /*indicator1*/, '0' /*indicator 2*/);
+                new_record.insertField("856", { { 'u', "https://www.icpsr.umich.edu/web/NACJD/studies/" + UrlUtil::UrlEncode(id) } },
+                                       '4' /*indicator1*/, '0' /*indicator 2*/);
                 new_record.insertField("935", { { 'a', "mkri" } });
+                new_record.insertField("935", { { 'a', "nacj" } });
```
mtrojan-ub, line 334: "nacj" and "foda" must also have subfield "2" set to "LOK"

cpp/harvest/nacjd.cc
```diff
@@ -343,9 +355,14 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
 
 
 int Main(int argc, char *argv[]) {
-    if (argc != 3)
+    if (argc < 3 or argc > 4)
```
mtrojan-ub, line 358: Whenever you change the processing of argc/argv, you should always add documentation in the Usage() output at the top of this file (beginning at line 44).

cpp/harvest/nacjd.cc
```diff
@@ -343,9 +355,14 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
...
         Usage();
 
+    int count_rows(9000);
+    if (argv[3]) {
```
mtrojan-ub, line 362: Argc is allowed to be 3, and in this case accessing argv[3] will cause an error. Instead of `if (argv[3]) {` you should check for `if (argc == 4) {` instead.

cpp/harvest/nacjd.cc
```diff
@@ -343,9 +355,14 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
 
 
 int Main(int argc, char *argv[]) {
-    if (argc != 3)
+    if (argc < 3 or argc > 4)
         Usage();
 
+    int count_rows(9000);
```
mtrojan-ub, line 361: count_rows should always have a positive value, so instead of `int` you should use `unsigned` (Also in the function headers).

cpp/harvest/nacjd.cc
```diff
@@ -178,14 +180,20 @@ void ExtractIDsFromWebsite(const std::set<std::string> &parsed_marc_ids, unsigne
     std::ofstream json_new_titles(NACJD_NEW_TITLES_JSON);
     json_new_titles << "{ \"nacjd\" : [ " << '\n';
     bool first(true);
+    unsigned int limit_test(0);
```
mtrojan-ub, line 183: limit_test seems to be debug-related and should be removed

cpp/harvest/nacjd.cc
```diff
@@ -158,10 +159,11 @@ void ExtractExistingIDsFromMarc(MARC::Reader * const marc_reader, std::set<std::
 }
 
 
-void ExtractIDsFromWebsite(const std::set<std::string> &parsed_marc_ids, unsigned * const number_of_new_ids) {
+void ExtractIDsFromWebsite(const std::set<std::string> &parsed_marc_ids, unsigned * const number_of_new_ids, unsigned int count_rows) {
```
mtrojan-ub, line 162: count_rows should be const as well => `const unsigned int`

cpp/harvest/nacjd.cc
```diff
@@ -343,9 +356,14 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
 
 
 int Main(int argc, char *argv[]) {
-    if (argc != 3)
+    if (argc < 3 or argc > 4)
         Usage();
 
+    unsigned int count_rows(9000);
```
mtrojan-ub, line 362: just `unsigned`, not `unsigned int`

cpp/harvest/nacjd.cc
mtrojan-ub: if you wanna keep it, please rename it and change the type from `unsigned int` to `unsigned`

PR #4238, opened 2022-06-24
changes for NACJD

LysogorAndGmail:


cpp/harvest/nacjd.cc
```diff
@@ -43,11 +43,18 @@ namespace {
 
 [[noreturn]] void Usage() {
     ::Usage(
-        "marc_title_in_file new_marc_title_out_file [count_of_rows]\n"
+        "marc_title_in_file new_marc_title_out_file [mode] [mode_params]\n"
```
mtrojan-ub, line 46: This should only be `[mode] [mode_params]`. The other params are totally depending on the mode. E.g. if we download a single ID, we do not need marc_title_in_file and new_marc_title_out_file.

cpp/harvest/nacjd.cc
```diff
@@ -355,28 +385,41 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
...
 int Main(int argc, char *argv[]) {
-    if (argc < 3 or argc > 4)
+    if (argc < 3)
         Usage();
 
     unsigned count_rows(9000);
-    if (argc == 4) {
-        count_rows = std::stoi(argv[3]);
+    unsigned mode_param(0);
+    std::string mode("");
```
mtrojan-ub, line 393: `mode` should always be set to `argv[1]`.

cpp/harvest/nacjd.cc
```diff
@@ -355,28 +385,41 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
...
 
 int Main(int argc, char *argv[]) {
-    if (argc < 3 or argc > 4)
+    if (argc < 3)
         Usage();
 
     unsigned count_rows(9000);
-    if (argc == 4) {
-        count_rows = std::stoi(argv[3]);
+    unsigned mode_param(0);
```
mtrojan-ub, line 392: mode_param sounds very generic and i dont think it should be implemented like this (i will give more detailed examples in other comments).

cpp/harvest/nacjd.cc
```diff
@@ -355,28 +385,41 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
...
-    // Download possible new publications and store in json file
-    LOG_INFO("Extracting ICPSR ids from website...");
-    unsigned number_of_new_ids;
-    ExtractIDsFromWebsite(parsed_marc_ids, &number_of_new_ids, count_rows);
-    LOG_INFO(std::to_string(number_of_new_ids) + " new ids collected from website.");
+    if (mode == "get_by_ID") {
+        mode_param = std::stoi(argv[4]);
+        ExtractOneIDFromWebsite(mode_param);
+        LOG_INFO("ID " + std::to_string(mode_param) + " collected from website.");
+    } else {
```
mtrojan-ub, line 410: You should explicitly test for all possible "modes" in `else if` statements. The `else` statement should only containt a call to`Usage()`, so that the program aborts if we encounter any unknown mode string.

cpp/harvest/nacjd.cc
```diff
@@ -355,28 +385,41 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
...
-    ExtractExistingIDsFromMarc(marc_reader.get(), &parsed_marc_ids);
-    LOG_INFO("Found: " + std::to_string(parsed_marc_ids.size()) + " records with ICPSR ids.");
-
-    // Download possible new publications and store in json file
-    LOG_INFO("Extracting ICPSR ids from website...");
-    unsigned number_of_new_ids;
-    ExtractIDsFromWebsite(parsed_marc_ids, &number_of_new_ids, count_rows);
-    LOG_INFO(std::to_string(number_of_new_ids) + " new ids collected from website.");
+    if (mode == "get_by_ID") {
+        mode_param = std::stoi(argv[4]);
```
mtrojan-ub, line 407: For each mode we must test first whether we have a suitable number of arguments (`argc`) and exit by calling `Usage()` if it does not exist, before we try to access `argv`.

mtrojan-ub: Also, for this case we want to call this variable `id` instead of `mode_param`.

cpp/harvest/nacjd.cc
```diff
@@ -355,28 +385,41 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
...
-        count_rows = std::stoi(argv[3]);
+    unsigned mode_param(0);
+    std::string mode("");
+
+    if (argv[3]) {
+        mode = argv[3];
+    }
+
+    if (mode == "count_of_rows") {
+        count_rows = std::stoi(argv[4]);
```
mtrojan-ub, line 400: This whole block should be moved to an "else if" block where all modes are checked (`if (mode == "get_by_ID") {`... `} else if mode == "count_of_rows" {`). And also the declaration of tje count_rows variable should be moved into this block because we do not need the variable outside of the block.

cpp/harvest/nacjd.cc
mtrojan-ub: As far as i can see, what you implemented so far are 2 additional modes `get_by_ID` and `count_of_rows`. The old  behaviour you still have in the `else` branch right now should be moved to another named mode, e.g. `harvest`.

cpp/harvest/nacjd.cc
```diff
@@ -158,6 +165,29 @@ void ExtractExistingIDsFromMarc(MARC::Reader * const marc_reader, std::set<std::
     }
 }
 
+void ExtractOneIDFromWebsite(const unsigned marc_id) {
```
mtrojan-ub, line 168: The parameter `marc_id` should be renamed to `nacjd_id`.

cpp/harvest/nacjd.cc
```diff
@@ -158,6 +165,29 @@ void ExtractExistingIDsFromMarc(MARC::Reader * const marc_reader, std::set<std::
...
+    if (FileUtil::Exists(NACJD_TITLES))
```
mtrojan-ub, line 169: The paths NACJD_TITLES and NACJD_NEW_TITLES_JSON should not be hardcoded. There should be an `output_path` variable that should be parsed from argv and passed to this function (and also: for this function we only need a single output_path instead of 2).

cpp/harvest/nacjd.cc
```diff
@@ -43,11 +43,18 @@ namespace {
...
         "where marc_title_in_file contains also icpsr records (001 or 035a)\n"
         "these records are not processed any more\n"
         "new_marc_title_out_file contains all icpsr records not contained in input file.\n"
-        "count_of_rows is an optional parameter. If it is absent, the number of requests will be equal to 9000)");
+        "\n"
+        "marc_title_in_file new_marc_title_out_file count_of_rows number\n"
+        "\t- count_of_rows: mode name\n"
+        "\t- number: is an optional parameter. If it is absent, the number of requests will be equal to 9000\n"
+        "\n"
+        "marc_title_in_file new_marc_title_out_file get_by_ID ID\n"
```
mtrojan-ub, line 55: This block should be something like
"get_by_ID ID output_path\n"
"\t- ID: The NACJD of the dataset to download.\n"
"\t- output_path: The output path for JSON file that will contain the downloaded dataset.\n"

cpp/harvest/nacjd.cc
```diff
@@ -43,11 +43,18 @@ namespace {
...
 [[noreturn]] void Usage() {
     ::Usage(
-        "marc_title_in_file new_marc_title_out_file [count_of_rows]\n"
+        "marc_title_in_file new_marc_title_out_file [mode] [mode_params]\n"
         "where marc_title_in_file contains also icpsr records (001 or 035a)\n"
         "these records are not processed any more\n"
         "new_marc_title_out_file contains all icpsr records not contained in input file.\n"
-        "count_of_rows is an optional parameter. If it is absent, the number of requests will be equal to 9000)");
+        "\n"
+        "marc_title_in_file new_marc_title_out_file count_of_rows number\n"
```
mtrojan-ub, line 51: this should only be
`"count_of_rows number\n"`

with a fitting description. But since i don't really understand what this mode should do and it is also not implemented yet, we should talk about it before you continue working on this mode.

cpp/harvest/nacjd.cc
```diff
@@ -43,25 +43,31 @@ namespace {
...
+        "\t- input_file: contains also icpsr records (001 or 035a)\n"
+        "these records are not processed any more\n"
+        "\t- output_file contains all icpsr records not contained in input file.\n"
         "\t- number: is an optional parameter. If it is absent, the number of requests will be equal to 9000\n"
         "\n"
-        "marc_title_in_file new_marc_title_out_file get_by_ID ID\n"
-        "\t- get_by_ID: mode name\n"
-        "\t- ID: article identifier\n");
+        "get_by_ID output_file ID\n"
+        "\t- output_file contains all icpsr records not contained in input file.\n"
```
mtrojan-ub, line 59: The help text for output_file is incorrect. Since we have no input file here and we do not check for icpsr records in this mode, the output_file should contain the data for the single dataset.

cpp/harvest/nacjd.cc
```diff
@@ -345,26 +384,68 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
...
+void getByID(int argc, char **argv) {
+    // Parse args
+    if (argc != 4)
         Usage();
 
-    unsigned count_rows(9000);
-    if (argc == 4) {
-        count_rows = std::stoi(argv[3]);
-    }
+    // auto marc_writer(MARC::Writer::Factory(argv[2]));
```
mtrojan-ub, line 425: These comments look like debug comments and should be removed

cpp/harvest/nacjd.cc
```diff
@@ -345,26 +384,68 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
...
+    // parse json file and store relevant information in variables
+    // write marc_records via marc_writer to marc_file
+    LOG_INFO("Parsing intermediate json file and save to marc output...");
+    ParseJSONAndWriteMARC(marc_writer.get());
+    LOG_INFO("Finished.");
+}
+
+void getByCount(int argc, char **argv) {
+    // Parse args
+    if (argc != 5)
```
mtrojan-ub, line 444: The Usage() documentation says that "number" is an optional parameter. If this is correct, then argc can either be 4 or 5.

cpp/harvest/nacjd.cc
```diff
@@ -375,14 +456,34 @@ int Main(int argc, char *argv[]) {
     // Download possible new publications and store in json file
     LOG_INFO("Extracting ICPSR ids from website...");
     unsigned number_of_new_ids;
-    ExtractIDsFromWebsite(parsed_marc_ids, &number_of_new_ids, count_rows);
+    ExtractIDsFromWebsite(parsed_marc_ids, &number_of_new_ids, std::stoi(argv[4]));
```
mtrojan-ub, line 459: Since argv[4] is the optional `number`, we need to check for `argc == 5 ` and only pass the parameter in this case.

cpp/harvest/nacjd.cc
```diff
@@ -158,6 +171,32 @@ void ExtractExistingIDsFromMarc(MARC::Reader * const marc_reader, std::set<std::
...
+    if (file)
+        std::for_each(std::istream_iterator<char>(file), std::istream_iterator<char>(), HandleChar);
+    else
+        LOG_ERROR("couldn't open file: " + nacjd_title_path);
+    if (FileUtil::Exists(nacjd_JSON_path) and not FileUtil::DeleteFile(nacjd_JSON_path))
+        LOG_ERROR("Could not delete file: " + nacjd_JSON_path);
+
+    std::ofstream json_new_titles(nacjd_JSON_path);
+
+    json_new_titles << "{ \"nacjd\" : [ " << '\n';
```
mtrojan-ub, line 193: I wonder if we should use a different JSON wrapper here, but we might leave that open for now and discuss it for a later PR when we add the other modes.

cpp/harvest/nacjd.cc
```diff
@@ -43,18 +43,31 @@ namespace {
...
+        "get_by_ID output_file ID\n"
+        "\t- output_file contains all icpsr records not contained in input file.\n"
+        "\t- ID: The NACJD of the dataset to download\n");
 }
 
 
 std::vector<std::string> ids_website;
 const unsigned int TIMEOUT_IN_SECONDS(15);
 const std::string NACJD_TITLES("/tmp/nacjd_titles.html");
+const std::string NACJD_TITLES_FILE_NAME("/nacjd_titles.html");
```
mtrojan-ub, line 67: Please remove all of these NACJD_... constants that contain and kind of file or path.

cpp/harvest/nacjd.cc
```diff
@@ -158,6 +171,32 @@ void ExtractExistingIDsFromMarc(MARC::Reader * const marc_reader, std::set<std::
     }
 }
 
+void ExtractOneIDFromWebsite(const unsigned nacjd_id, const std::string nacjd_path) {
+    std::string nacjd_title_path(nacjd_path + NACJD_TITLES_FILE_NAME);
```
mtrojan-ub, line 175: I don't really understand why we need 2 files here. It should be enough to have a nacjd_json_path (which could be renamed to output_path or json_path, since the whole program is related to nacjd).

cpp/harvest/nacjd.cc
```diff
@@ -345,26 +384,68 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
...
-    unsigned count_rows(9000);
-    if (argc == 4) {
-        count_rows = std::stoi(argv[3]);
-    }
+    // auto marc_writer(MARC::Writer::Factory(argv[2]));
+
+    // std::string nacjd_file("/tmp");
+    auto marc_writer(MARC::Writer::Factory(argv[2] + NACJD_FILE_NAME));
+
+    unsigned nacjd_id = std::stoi(argv[3]);
```
mtrojan-ub, line 430: Please also use constructor notation here and make it const.

`const unsigned nacjd_id(std::stoi(argv[3]));`

(and you could even consider just renaming the variable to `id`).

cpp/harvest/nacjd.cc
```diff
@@ -158,26 +165,52 @@ void ExtractExistingIDsFromMarc(MARC::Reader * const marc_reader, std::set<std::
     }
 }
 
+void ExtractOneIDFromWebsite(const unsigned nacjd_id) {
+    std::string nacjd_title_path("/tmp/nacjd_titles.html");
```
mtrojan-ub, line 169: It seems like this function was partially copy/pasted from `ExtractIDsFromWebsite`. In that other function, the HTML file has the purpose to get the list of all IDs from it.

However, in this new function we only want to extract a single dataset whereof the ID is already known via the parameter nacjd_id. So i suppose we do not need the HTML file in this function at all and it can be entirely removed.

cpp/harvest/nacjd.cc
```diff
@@ -345,27 +378,66 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
...
 
+    const unsigned id = std::stoi(argv[3]);
+
+    ExtractOneIDFromWebsite(id);
+    LOG_INFO("ID " + std::to_string(id) + " collected from website.");
+
+    // parse json file and store relevant information in variables
+    // write marc_records via marc_writer to marc_file
+    LOG_INFO("Parsing intermediate json file and save to marc output...");
+    ParseJSONAndWriteMARC(marc_writer.get());
```
mtrojan-ub, line 429: The goal of the getByID operation is to download the dataset as JSON and store it in the output file. Conversion from JSON to MARC will be a different mode later. So you can remove the MARC conversion (as well as the initialization of the marc_writer) from this function, maybe this will also make things easier regarding the hardcoded file paths.

cpp/harvest/nacjd.cc
```diff
@@ -345,27 +378,66 @@ void ParseJSONAndWriteMARC(MARC::Writer * const title_writer) {
...
+    // Parse args
+    if (argc != 4)
+        Usage();
 
-    auto marc_reader(MARC::Reader::Factory(argv[1]));
     auto marc_writer(MARC::Writer::Factory(argv[2]));
 
+    const unsigned id = std::stoi(argv[3]);
+
+    ExtractOneIDFromWebsite(id);
```
mtrojan-ub, line 423: We should also pass argv[2] to this function. This can be used as destination path for the JSON file (instead of the hardcoded nacjd_JSON_path variable in that function.

cpp/harvest/nacjd.cc
```diff
@@ -158,26 +165,52 @@ void ExtractExistingIDsFromMarc(MARC::Reader * const marc_reader, std::set<std::
...
+}
 
 void ExtractIDsFromWebsite(const std::set<std::string> &parsed_marc_ids, unsigned * const number_of_new_ids, const unsigned count_rows) {
     const std::string DOWNLOAD_URL(
         "https://www.icpsr.umich.edu/web/NACJD/search/"
         "studies?start=0&ARCHIVE=NACJD&PUBLISH_STATUS=PUBLISHED&sort=DATEUPDATED%20desc&rows="
         + std::to_string(count_rows));
-    if (FileUtil::Exists(NACJD_TITLES))
-        FileUtil::DeleteFile(NACJD_TITLES);
+    if (FileUtil::Exists("/tmp/nacjd_titles.html"))
```
mtrojan-ub, line 200: This HTML file should only be used inside this function. Instead of hardcoding a path, please use FileUtil::AutoTempFile instead. It will automatically generate a temporary file in /tmp and automatically delete it as soon as the object goes out of scope (e.g. when the function terminates successfully).

cpp/harvest/nacjd.cc
```diff
@@ -158,26 +165,52 @@ void ExtractExistingIDsFromMarc(MARC::Reader * const marc_reader, std::set<std::
...
     if (file)
         std::for_each(std::istream_iterator<char>(file), std::istream_iterator<char>(), HandleChar);
     else
-        LOG_ERROR("couldn't open file: " + NACJD_TITLES);
+        LOG_ERROR("couldn't open file: /tmp/nacjd_titles.html");
 
-    if (FileUtil::Exists(NACJD_NEW_TITLES_JSON) and not FileUtil::DeleteFile(NACJD_NEW_TITLES_JSON))
-        LOG_ERROR("Could not delete file: " + NACJD_NEW_TITLES_JSON);
-    std::ofstream json_new_titles(NACJD_NEW_TITLES_JSON);
+    if (FileUtil::Exists("/tmp/nacjd_new_titles.json") and not FileUtil::DeleteFile("/tmp/nacjd_new_titles.json"))
```
mtrojan-ub, line 211: The path to the JSON output file should in this case be passed as a function parameter to `ExtractIDsFromWebsite`. When it is called in the `getAll` function, we can also generate an AutoTempFile there for now which will then be passed to `ParseJSONAndWriteMARC`. 

Of course, in the long term we want to remove this step entirely from this mode and add a `convert` mode instead, so the other option would be to just use argv[3] as JSON target path.

PR #4244, opened 2022-07-04
adding get_statistics functionality

LysogorAndGmail:


cpp/harvest/nacjd.cc
```diff
@@ -452,6 +444,170 @@ void convertJSONtoMARC(int argc, char **argv) {
...
+            if (license.empty())
+                ++no_license;
+            if (initial_release_date.empty())
+                ++no_initial_date;
+        }
+    }
+
+    LOG_INFO("\nStatistics:\n\tProcessed: " + std::to_string(total) + " entries.\n" + "\tProblems: \n" + "\t\tTitle not found or empty: "
+             + std::to_string(no_title) + "\n" + "\t\tDescription not found or empty: " + std::to_string(no_description) + "\n"
+             + "\t\tIdentifiier not found or empty: " + std::to_string(no_ID) + "\n" + "\t\tAlternate Identifiier not found or empty: "
```
mtrojan-ub, line 606: `Identifiier` => `Identifier`

cpp/harvest/nacjd.cc
```diff
@@ -61,7 +61,9 @@ namespace {
         "\n"
         "convert_JSON_to_MARC json_path marc_path\n"
         "\t- json_path: The input JSON file.\n"
-        "\t- marc_path: The output MARC file.\n");
+        "\t- marc_path: The output MARC file.\n"
+        "get_statistics json_path\n"
```
mtrojan-ub, line 65: Please don't forget to add an empty line between this mode and the previous one.

cpp/harvest/nacjd.cc
```diff
@@ -452,6 +444,170 @@ void convertJSONtoMARC(int argc, char **argv) {
     LOG_INFO("Finished.");
 }
 
+void getStatistics(int argc, char **argv) {
+    // Parse args
+    if (argc != 3)
+        Usage();
+
+    std::string json_path = argv[2];
```
mtrojan-ub, line 452: This value will never change. Please use the const keyword and change it to constructor notation:
`const std::string json_path(argv[2]);`

cpp/harvest/nacjd.cc
```diff
@@ -452,6 +444,170 @@ void convertJSONtoMARC(int argc, char **argv) {
...
+            }
+        }
+
+        if (keywords_node != nullptr) {
+            for (const auto &internal_keyword_node : *keywords_node) {
+                const auto keyword_node(JSON::JSONNode::CastToObjectNodeOrDie("keyword", internal_keyword_node));
+                const auto keyword_value_node(keyword_node->getStringNode("value"));
+                if (not keyword_value_node->getValue().empty())
+                    keywords.emplace(keyword_value_node->getValue());
+            }
```
mtrojan-ub, line 513: If the keywords set is empty after this loop we also want do perform `++no_keywords`.

cpp/harvest/nacjd.cc
```diff
@@ -452,6 +444,170 @@ void convertJSONtoMARC(int argc, char **argv) {
...
+                    continue;
+                if (ContainsValue(creators, "100"))
+                    creators.emplace(fullName, "700");
+                else
+                    creators.emplace(fullName, "100");
+            } else
+                LOG_ERROR("unknown creator type: " + type);
+        }
+
+        if (creators.empty() or license.empty() or initial_release_date.empty()) {
```
mtrojan-ub, line 594: The outer block is not really needed here.

PR #4254, opened 2022-07-20
RSS information test

LysogorAndGmail:


cpp/test/SyndicationFormat_test.cc
```diff
@@ -0,0 +1,57 @@
...
+    newDataFile << parseDocument;
+
+    return true;
+}
+
+int Main(int argc, char *argv[]) {
+    if (argc != 2)
+        Usage();
+
+    std::string fileName("/tmp/rssData");
```
mtrojan-ub, line 36: Most of these should be const

mtrojan-ub: And please try to use FileUtil::AutoTempFile instead of a hardcoded filename.

PR #4256, opened 2022-07-29
category statistics

LysogorAndGmail:


cpp/harvest/nacjd.cc
```diff
@@ -64,13 +64,29 @@ namespace {
...
+        "\n"
+        "get_categories_statistics json_path\n"
         "\t- json_path: The input JSON file.\n");
 }
 
 
 std::vector<std::string> ids_website;
 const unsigned int TIMEOUT_IN_SECONDS(15);
 
+const std::string statistics_categories_array[] = { "National Crime Victimization Survey",
```
mtrojan-ub, line 77: Please add a comment here that this list contains the series from the "National Statistics" section on the following page:
https://www.icpsr.umich.edu/web/pages/NACJD/discover-data.html

cpp/harvest/nacjd.cc
```diff
@@ -64,13 +64,29 @@ namespace {
...
+                                                    "FBI Uniform Crime Reporting",
+                                                    "Annual Parole Survey",
+                                                    "Annual Probation Survey",
+                                                    "Annual Survey of Jails",
+                                                    "Mortality in Correctional Institutions",
+                                                    "Federal Court Cases Integrated Database",
+                                                    "Federal Justice Statistics Program",
+                                                    "National Corrections Reporting Program" };
+
+int statistics_count[] = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
```
mtrojan-ub, line 88: 1) This should not be defined as a global variable, it should be declared locally intead
2) Instead of using a int[] with a hardcoded length, you should use a more dynamic datatype instead, e.g. a std::map<std::string, unsigned> (where the std::string is the name of the category).

cpp/harvest/nacjd.cc
mtrojan-ub: Please change the definition:
`const std::string statistics_categories_array[] = ...`
=> 
`const std::vector<std::string> statistics_categories = ...`

cpp/harvest/nacjd.cc
```diff
@@ -119,6 +135,17 @@ void HandleChar(const char c) {
     }
 }
 
+std::string CheckingStatisticsCategory(std::string data_title) {
```
mtrojan-ub, line 138: data_title should be const

cpp/harvest/nacjd.cc
```diff
@@ -119,6 +135,17 @@ void HandleChar(const char c) {
...
+    std::string category_name("");
+    for (std::string n : statistics_categories_array) {
```
mtrojan-ub, line 140: usually when iterating over a container and you do it like this, a copy of all variables is created. you can iterate by reference instead like this:

`for (const std::string &n : statistics_categories_array) {`

cpp/harvest/nacjd.cc
```diff
@@ -119,6 +135,17 @@ void HandleChar(const char c) {
...
+        size_t pos = data_title.find(n);
+        if (pos != std::string::npos) {
+            category_name = n;
```
mtrojan-ub, line 143: You can simply return the category here directly to skip the rest of the iteration when the first category is found, and then just return an empty string as a fallback if no categories are found at the end of the function.

cpp/harvest/nacjd.cc
mtrojan-ub: And a better name would be either CheckStatisticsCategory or GetStatisticsCategory

cpp/harvest/nacjd.cc
```diff
@@ -613,6 +643,71 @@ void getStatistics(int argc, char **argv) {
...
 
+void getCategoriesStatistics(int argc, char **argv) {
+    // Parse args
+    if (argc != 3)
+        Usage();
+
+    const std::string json_path = argv[2];
+
+    std::string json_document;
+    FileUtil::ReadStringOrDie(json_path, &json_document);
```
mtrojan-ub, line 654: You can save a few lines of code by using JSON::ParseFile(json_document);

cpp/harvest/nacjd.cc
```diff
@@ -613,6 +643,71 @@ void getStatistics(int argc, char **argv) {
              + std::to_string(no_initial_date) + "\n" + "\t\tKeywords not found or empty: " + std::to_string(no_keywords) + "\n");
 }
 
+void getCategoriesStatistics(int argc, char **argv) {
```
mtrojan-ub, line 646: There is lots of copy/paste code here (sometimes more than necessary), and there is already a get_statistics mode in the usage. Please add the necessary logic to the existing mode instead of adding a new one.

cpp/harvest/nacjd.cc
```diff
@@ -461,7 +486,9 @@ void getStatistics(int argc, char **argv) {
 
     const auto root_node(JSON::JSONNode::CastToObjectNodeOrDie("tree_root", internal_tree_root));
     unsigned total(0), no_title(0), no_description(0), no_ID(0), no_alternate_ID(0), no_creators(0), no_license(0), no_initial_date(0),
-        no_keywords(0);
+        no_keywords(0), no_statistics_category(0);
+
+    int statistics_count[] = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
```
mtrojan-ub, line 491: Instead of using a int[] with a hardcoded length, you should use a more dynamic datatype instead, e.g. a `std::map<std::string, unsigned>` (where the std::string is the name of the category)

PR #4294, opened 2022-11-02
Add feature split-data-provider by its id

steven-lolong:


cpp/harvest/core.cc
```diff
@@ -553,6 +556,95 @@ void DataProviders(int argc, char **argv) {
...
+    if (argc != 4)
+        Usage();
+
+    std::cout << "Preparing data..." << std::endl;
+    const auto works(CORE::GetWorksFromFile(argv[2]));
+
+    std::vector<unsigned> list_of_data_provider_id;
+    std::map<std::string, unsigned> data_provider_id_counter;
+    unsigned total_record_counter(0), progress(0), step(0), display_next(0), closed_counter(0);
+    std::string item, outputFile, outputDir(argv[3]);
```
mtrojan-ub, line 569: Please also use underscore here for all variable names, e.g. output_file, output_dir.

cpp/harvest/core.cc
```diff
@@ -553,6 +556,95 @@ void DataProviders(int argc, char **argv) {
...
+
+
+    std::cout << "Processing " << works.size() << " records ..." << std::endl;
+    for (auto work : works) {
+        if (!work.getDataProviderIds().empty()) {
+            for (auto data_provider_id : work.getDataProviderIds()) {
```
mtrojan-ub, line 575: Same here (or at least use `const auto` please)

cpp/harvest/core.cc
```diff
@@ -553,6 +556,95 @@ void DataProviders(int argc, char **argv) {
...
+                item = std::to_string(data_provider_id);
+                outputFile = outputDir + item + ".json";
```
mtrojan-ub, line 577: are you absolute sure that outputDir has a trailing slash?

cpp/harvest/core.cc
```diff
@@ -553,6 +556,95 @@ void DataProviders(int argc, char **argv) {
...
 
 
+void SplitDataProviderId(int argc, char **argv) {
+    if (argc != 4)
+        Usage();
+
+    std::cout << "Preparing data..." << std::endl;
+    const auto works(CORE::GetWorksFromFile(argv[2]));
+
+    std::vector<unsigned> list_of_data_provider_id;
```
mtrojan-ub, line 566: in CORE.h, we use `unsigned long` as type for a single data provider id. Please use that here / everywhere as well.

cpp/harvest/core.cc
mtrojan-ub, line 569: `item` is actually a `data_provider_id_string`. Please rename it or remove it entirely.

cpp/harvest/core.cc
```diff
@@ -553,6 +556,95 @@ void DataProviders(int argc, char **argv) {
...
+                }
+            }
+        }
+
+        ++total_record_counter;
+
+        // displaying progress
+        progress = (100 * (total_record_counter)) / works.size();
+        if (progress >= display_next) {
+            std::cout << "\r"
```
mtrojan-ub, line 598: is `\r` a carriage return? On Unix-based systems we should use a linefeed `\n` instead.

cpp/harvest/core.cc
```diff
@@ -553,6 +556,95 @@ void DataProviders(int argc, char **argv) {
...
+                      << "[" << std::string(progress / 5, (char)35u) << std::string(100 / 5 - progress / 5, ' ') << "]";
+            std::cout << progress << "%"
+                      << " [record " << total_record_counter << " of " << works.size() << "]";
+            std::cout.flush();
+            display_next += step;
```
mtrojan-ub, line 603: Does `step` ever change? It always seems to be 0.

cpp/harvest/core.cc
```diff
@@ -553,6 +556,95 @@ void DataProviders(int argc, char **argv) {
...
+
+    std::vector<unsigned> list_of_data_provider_id;
+    std::map<std::string, unsigned> data_provider_id_counter;
+    unsigned total_record_counter(0), progress(0), step(0), display_next(0), closed_counter(0);
+    std::string item, outputFile, outputDir(argv[3]);
+
+
+    std::cout << "Processing " << works.size() << " records ..." << std::endl;
+    for (auto work : works) {
+        if (!work.getDataProviderIds().empty()) {
```
mtrojan-ub, line 574: Maybe we should at least output a warning if we ever find a work without a single data provider id?

mtrojan-ub: (or should we write all of them to a separate file like e.g. empty.json?)

cpp/harvest/core.cc
```diff
@@ -553,6 +556,95 @@ void DataProviders(int argc, char **argv) {
...
+
+    // add bracket as a closing annotation in each file
+    display_next = 1;
+    std::cout << std::endl << "Found " << list_of_data_provider_id.size() << " unique Data Provider Id" << std::endl;
+    std::cout << "Updating Data Provider Id's file ..." << std::endl;
+    for (auto const &value : list_of_data_provider_id) {
+        outputFile = outputDir + std::to_string(value) + ".json";
+        CORE::OutputFileEnd(outputFile);
+
+        // displaying progress
```
mtrojan-ub, line 615: Do we really need a progress output for this step?

cpp/harvest/core.cc
steven-lolong: \r is a carriage return, but the next output is still in the same line.
\n is a carriage return, and the next output will show in the new line.

cpp/harvest/core.cc
```diff
@@ -553,6 +556,95 @@ void DataProviders(int argc, char **argv) {
...
+                      << "[" << std::string(progress / 5, (char)35u) << std::string(100 / 5 - progress / 5, ' ') << "]";
+            std::cout << progress << "%"
+                      << " [file " << closed_counter << " of " << list_of_data_provider_id.size() << "]";
+            std::cout.flush();
+            display_next += step;
+        }
+    }
+
+    // writing a summary
+    std::string rptFile(outputDir + "aSummaryRpt.json");
```
mtrojan-ub, line 629: Here as well: please declare it as `const std::string output_dir`

cpp/harvest/core.cc
steven-lolong: it is just a constant. 0 means will display the progress percentage every time it changes. Ex. the step is five means it will show the progress percentage every 5% of progress. 
Maybe I will change its type to const unsigned.

cpp/harvest/core.cc
steven-lolong: I think yes, because if the number of files is thousands then it takes some seconds to process it.

cpp/harvest/core.cc
steven-lolong: I added the information about data without the data provider id in the summary_report file and notification.

cpp/harvest/core.cc
steven-lolong: I added a checker on wether it has a backslash or not.

PR #4295, opened 2022-11-03
DOI link for publication

LysogorAndGmail:
This PR should be merged together with ubtue/tuefind#2495


cpp/dspace/doi.cc
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    return true;
+}
+
+int main() {
+    auto db_reader((DbConnection::VuFindMySQLFactory()));
+
+    if (not VuFind::GetTueFindFlavour().empty()) {
+        auto db_writer(DbConnection::VuFindMySQLFactory());
```
mtrojan-ub, line 88: I don't think we need 2 different DB connections to the same database (VuFind)

mtrojan-ub: (or maybe we do since we iterate & write at the same time... if that's the case, its ok)

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,98 @@
...
+            SendNotificationsForDOI("DOI link notification",
+                                    "DOI link: " + DOILink + " successfully generated for publication " + PublicationTitle);
+        }
+    }
+
+    return true;
+}
+
+int main() {
+    auto db_reader((DbConnection::VuFindMySQLFactory()));
```
mtrojan-ub, line 85: This should be initialized directly before the db_writer.

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,98 @@
...
+        > 299)
+        LOG_ERROR("Failed to send the DOI notification email!");
+}
+
+void UpdateItem(DbConnection * const db_writer, const std::string doi_link, const std::string publication_id) {
+    db_writer->queryOrDie("UPDATE tuefind_publications SET doi_link = " + db_writer->escapeAndQuoteString(doi_link)
+                          + ",doi_notification = CURDATE() WHERE id= " + publication_id + "");
+}
+
+bool DownloadAndUpdate(DbConnection * const db_writer, const std::string external_document_id, const std::string publicationID) {
```
mtrojan-ub, line 37: Should be const references:
`const std::string &external_document_id`
`const std::string &publication_id` (and this should not be camel case, we use lowecase/underscore in C++)

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,98 @@
...
+
+    const std::string &json_document(downloader.getMessageBody());
+    std::shared_ptr<JSON::JSONNode> full_tree;
+    JSON::Parser parser(json_document);
+    if (not parser.parse(&full_tree))
+        LOG_ERROR("failed to parse JSON (" + parser.getErrorMessage());
+
+    const std::shared_ptr<const JSON::ArrayNode> top_node_array(JSON::JSONNode::CastToArrayNodeOrDie("full_tree", full_tree));
+
+    std::string PublicationTitle;
```
mtrojan-ub, line 60: should be renamed to `publication_title`

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,98 @@
...
+
+    for (auto item_iter : *top_node_array) {
+        const std::string title_key(JSON::LookupString("/key", item_iter, ""));
+        const std::string title_value(JSON::LookupString("/value", item_iter, ""));
+        if (PublicationTitleKey == title_key) {
+            PublicationTitle = title_value;
+        }
+    }
+
+    for (auto item_iter : *top_node_array) {
```
mtrojan-ub, line 70: Why are you doing the same iteration twice?

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,98 @@
...
+    std::shared_ptr<JSON::JSONNode> full_tree;
+    JSON::Parser parser(json_document);
+    if (not parser.parse(&full_tree))
+        LOG_ERROR("failed to parse JSON (" + parser.getErrorMessage());
+
+    const std::shared_ptr<const JSON::ArrayNode> top_node_array(JSON::JSONNode::CastToArrayNodeOrDie("full_tree", full_tree));
+
+    std::string PublicationTitle;
+
+    for (auto item_iter : *top_node_array) {
```
mtrojan-ub, line 62: Shouldn't this be `const auto &item_iter`?

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,98 @@
...
+        if (PublicationTitleKey == title_key) {
+            PublicationTitle = title_value;
+        }
+    }
+
+    for (auto item_iter : *top_node_array) {
+        const std::string one_key(JSON::LookupString("/key", item_iter, ""));
+        const std::string DOILink(JSON::LookupString("/value", item_iter, ""));
+        if (DOIKey == one_key) {
+            std::cout << "Processing: " << DOILink << '\n';
```
mtrojan-ub, line 74: Usually we should not write to std::cout directly. Please use `LOG_INFO(doi_link)` instead.

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,98 @@
...
+        const std::string title_key(JSON::LookupString("/key", item_iter, ""));
+        const std::string title_value(JSON::LookupString("/value", item_iter, ""));
+        if (PublicationTitleKey == title_key) {
+            PublicationTitle = title_value;
+        }
+    }
+
+    for (auto item_iter : *top_node_array) {
+        const std::string one_key(JSON::LookupString("/key", item_iter, ""));
+        const std::string DOILink(JSON::LookupString("/value", item_iter, ""));
```
mtrojan-ub, line 72: This should be renamed to `doi_link`

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,98 @@
...
+#include "JSON.h"
+#include "StlHelpers.h"
+#include "StringUtil.h"
+#include "VuFind.h"
+#include "util.h"
+
+using namespace std;
+
+const std::string DspaceServerURL("https://publikationen.uni-tuebingen.de/rest/items/");
+const std::string DOIKey("dc.relation.uri");
```
mtrojan-ub, line 21: If these are global constants, they should all be in uppercase with underscores:

```
DSPACE_SERVER_URL
DOI_KEY
PUBLICATION_TITLE_KEY
```

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,98 @@
...
+const std::string PublicationTitleKey("dc.title");
+const std::vector<std::string> notification_email_addresses = { "andrii.lysohor@uni-tuebingen.de" };
```
mtrojan-ub, line 23: The notification email addresses still need to be defined, they will be most likely taken from a configuration file later

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,98 @@
...
+
+void SendNotificationsForDOI(const std::string email_subject, const std::string email_message) {
+    if (EmailSender::SimplerSendEmail("no-reply@ub.uni-tuebingen.de", notification_email_addresses, email_subject, email_message,
+                                      EmailSender::VERY_HIGH)
```
mtrojan-ub, line 27: We only use VERY_HIGH priority if we send error mails. Since this is just a normal mail, we can also send it with normal priority.

cpp/data/system_updates/123.vufind.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE tuefind_publications ADD doi_link VARCHAR(255) NULL AFTER external_document_guid;
```
mtrojan-ub, line 1: Do we really need to store the whole URL in the database? I think the DOI itself would be enough

cpp/data/system_updates/123.vufind.sql
```diff
@@ -0,0 +1,2 @@
...
+ALTER TABLE tuefind_publications ADD doi_notification DATE NULL AFTER doi_link;
```
mtrojan-ub, line 2: Maybe use DATETIME instead?

mtrojan-ub: (with DEFAULT NOW()? then maybe you dont even have to set an explicit value in the query)

cpp/data/system_updates/123.vufind.sql
mtrojan-ub: As discussed, we should still have an additional column called `doi`

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,100 @@
...
+
+void SendNotificationsForDOI(const std::string email_subject, const std::string email_message) {
+    if (EmailSender::SimplerSendEmail("no-reply@ub.uni-tuebingen.de", NOTIFICATION_EMAIL_ADRESSES, email_subject, email_message,
+                                      EmailSender::MEDIUM)
+        > 299)
+        LOG_ERROR("Failed to send the DOI notification email!");
+}
+
+void UpdateItem(DbConnection * const db_writer, const std::string publication_id) {
+    db_writer->queryOrDie("UPDATE tuefind_publications SET doi_notification = NOW() WHERE id= " + publication_id + "");
```
mtrojan-ub, line 34: - `+ ""` at the end of the line has no effect at all
- Whenever we build a query string out of variables, we should always escape them. e.g. `WHERE id=" + db_writer->escapeAndQuoteString(publication_id)`

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,100 @@
...
+
+using namespace std;
+
+const std::string DSPACE_SERVER_URL("https://publikationen.uni-tuebingen.de/rest/items/");
+const std::string DOI_KEY("dc.relation.uri");
+const std::string PUBLIKATION_TITLE_KEY("dc.title");
+unsigned int INSTANCE_COUNT(15000);
+const std::vector<std::string> NOTIFICATION_EMAIL_ADRESSES = { "andrii.lysohor@uni-tuebingen.de" };
+
+void SendNotificationsForDOI(const std::string email_subject, const std::string email_message) {
```
mtrojan-ub, line 26: Whenever we pass a const std::string to a function, it should always be passed as a reference for performance reasons, e.g. `const std::string &email_subject`, `const std::string &email_message`, and so on (the **&** is very important).

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,100 @@
...
+    if (EmailSender::SimplerSendEmail("no-reply@ub.uni-tuebingen.de", NOTIFICATION_EMAIL_ADRESSES, email_subject, email_message,
+                                      EmailSender::MEDIUM)
+        > 299)
+        LOG_ERROR("Failed to send the DOI notification email!");
+}
+
+void UpdateItem(DbConnection * const db_writer, const std::string publication_id) {
```
mtrojan-ub, line 33: Here as well, please change to `const std::string &publication_id`

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,100 @@
...
+    db_writer->queryOrDie("UPDATE tuefind_publications SET doi_notification = NOW() WHERE id= " + publication_id + "");
+}
+
+std::string ParseDataFromXML(std::shared_ptr<const JSON::ArrayNode> top_node_array, const std::string fild) {
```
mtrojan-ub, line 37: `fild` seems to be a typo, do you mean `field`?
and it should be a reference => `const std::string &field`

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,100 @@
...
+#include "StringUtil.h"
+#include "VuFind.h"
+#include "util.h"
+
+using namespace std;
+
+const std::string DSPACE_SERVER_URL("https://publikationen.uni-tuebingen.de/rest/items/");
+const std::string DOI_KEY("dc.relation.uri");
+const std::string PUBLIKATION_TITLE_KEY("dc.title");
+unsigned int INSTANCE_COUNT(15000);
```
mtrojan-ub, line 23: INSTANCE_COUNT doesn't seem to be a proper name... It will be passed to the Downloader to override the time_limit parameter. So a better name could be e.g. DSPACE_TIME_LIMIT.

But the real question here is =>  the default time_limit is 20000, so why do we even bother to explicitly set one to 15000? Maybe we can just leave the parameter empty so the default is taken?

cpp/data/system_updates/123.vufind.sql
mtrojan-ub: Looks better now! Don't forget to open a PR in ubtue/tuefind as well, which adds these columns to `module/TueFind/sql/mysql.sql`

cpp/dspace/doi.cc
```diff
@@ -0,0 +1,115 @@
...
+#include "VuFind.h"
+#include "util.h"
+
+using namespace std;
+
+const std::string DSPACE_SERVER_URL("https://publikationen.uni-tuebingen.de/rest/items/");
+const std::string DOI_KEY("dc.relation.uri");
+const std::string PUBLIKATION_TITLE_KEY("dc.title");
+const std::string DOI_URL_FORMAT("http://dx.doi.org/");
+const std::vector<std::string> NOTIFICATION_EMAIL_ADRESSES = { "andrii.lysohor@uni-tuebingen.de" };
```
mtrojan-ub, line 24: This address should be changed (will be clarified by thefass, maybe even meistertask)

PR #4298, opened 2022-11-09
update filter in harvest/core, and add setDataProvider, getDataProvider in CORE.h, CORE.cc

steven-lolong:


cpp/harvest/core.cc
```diff
@@ -318,17 +318,46 @@ void Filter(int argc, char **argv) {
...
+        if (argc == 7) {
+            if (!filter) {
+                // this means keep the record but clean the member of data provider
+                // clean the member of data provider means delete the member that its id is not in the list
+                // this is 'keep' option
+                auto data_provider_ids = work.getDataProviderIds();
+                const bool is_in_the_list(MiscUtil::Intersect(data_provider_ids, filter_data_provider_ids).size() > 0);
+                if (!data_provider_ids.empty() && is_in_the_list) {
+                    std::vector<nlohmann::json> new_data_provider;
+                    const auto data_providers = work.getDataProvider();
```
mtrojan-ub, line 330: This logic should be moved to CORE.h and CORE.cc => there should be a function called `removeDataProvider(unsigned long data_provider_id)` that should do this inside the library. We do not want to use nlohmann::json outside of CORE.h and CORE.cc.

cpp/harvest/core.cc
```diff
@@ -318,17 +318,39 @@ void Filter(int argc, char **argv) {
...
+        if (argc == 7) {
+            if (!filter) {
+                // this means keep the record but clean the member of data provider
+                // clean the member of data provider means delete the member that its id is not in the list
+                // this is 'keep' option
+                auto data_provider_ids = work.getDataProviderIds();
+                const bool is_in_the_list(MiscUtil::Intersect(data_provider_ids, filter_data_provider_ids).size() > 0);
+
+                if (!data_provider_ids.empty() && is_in_the_list) {
+                    work.removeDataProviders(filter_data_provider_ids);
```
mtrojan-ub, line 330: You can use the the intersection result as input for `removeDataProviders()` instead for better performance. Rest looks good!

PR #4313, opened 2022-12-01
add system variable 'VUFIND_LOCAL_MODULES' in /etc/profile.d/vufind.sh and '123.sh' for system_updater

steven-lolong:


cpp/data/system_updates/123.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -o errexit
+
+if [ $TUEFIND_FLAVOUR=="ixtheo" ] 
+then
+  echo "export VUFIND_LOCAL_MODULES=TueFindSearch,TueFind,IxTheo" >> /etc/profile.d/vufind.sh
```
mtrojan-ub, line 7: please use a different command like e.g. `sed` to make sure that this statement is added in the same spot where the installer would add it to.

cpp/installer.cc
```diff
@@ -674,9 +674,11 @@ void PermanentlySetEnvironmentVariables(const std::vector<std::pair<std::string,
 
 
 void SetVuFindEnvironmentVariables(const std::string &vufind_system_type_string) {
+    const std::string system_type_name_case_sensitive((vufind_system_type_string == "ixtheo" ? "IxTheo" : "KrimDok"));
```
mtrojan-ub, line 677: Also, please check explicitly for `krimdok`.

PR #4319, opened 2022-12-16
Adding new option to generate a 'gnd_to_wiki.csv' file in 'add_authority_external_ref.cc'

steven-lolong:


cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -144,6 +334,15 @@ int Main(int argc, char* argv[]) {
         ParseDataDnbFile(marc_output_filename_or_dnb_input, mapping_txt_filename);
         return EXIT_SUCCESS;
     }
+    if (marc_input_filename_or_create_flag == "--generate_gnd_to_wiki") {
```
mtrojan-ub, line 337: Please just replace the existing parameter `--create_mapping_file` and overwrite the logic there instead of adding a new one

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -130,9 +183,146 @@ void ParseGndWikidataMappingFile(std::string filename,
...
+    const int base_string_lenght = url_based.length();
+    const std::string sub_string_url_comp = url_comp.substr(0, base_string_lenght);
+    if (url_based.compare(sub_string_url_comp) == 0)
+        return true;
+
+    return false;
+}
+
+int GenerateGNDAuthorityExternalRef(char *argv[]) {
+    auto load_file_start = std::chrono::high_resolution_clock::now();
```
mtrojan-ub, line 209: can this be `const`? and is it possible to use constructor notation?
`const auto load_file_start(std::chrono::high_resolution_clock::now());`

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -130,9 +183,146 @@ void ParseGndWikidataMappingFile(std::string filename,
...
+
+    if (!input_file.is_open()) {
+        LOG_ERROR("can't open input file");
+        return 1;
+    }
+
+    const auto csv_file(FileUtil::OpenOutputFileOrDie(argv[3]));
+
+
+    const std::string d_nb_address("https://d-nb.info/gnd/");
```
mtrojan-ub, line 221: You can write it as `dnb_address`, there is no necessity for an additional underscore.

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -130,9 +183,146 @@ void ParseGndWikidataMappingFile(std::string filename,
...
+        std::cout << "\r"
+                  << "Parsed: " << total_line_parsed << " line(s), "
+                  << " Total GND-ID: " << total_numbers_of_gnd_id_generated << ", Total GND with Wikidata: " << total_number_of_wikidata
+                  << ", Total GND with Wikipedia: " << total_number_of_wikipedia;
+        //   << "GND-ID url: " << tmp_id;
+        std::cout.flush();
+        ++total_line_parsed;
+    }
+
+    auto load_file_end = std::chrono::high_resolution_clock::now();
```
mtrojan-ub, line 310: same here. load_file_end and duration will never change, so they should be `const`. Also please use constructor notation if possible.

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -130,9 +183,146 @@ void ParseGndWikidataMappingFile(std::string filename,
...
+    if (url_based.compare(sub_string_url_comp) == 0)
+        return true;
+
+    return false;
+}
+
+int GenerateGNDAuthorityExternalRef(char *argv[]) {
+    auto load_file_start = std::chrono::high_resolution_clock::now();
+    std::ifstream input_file(argv[2]);
+    std::ofstream summary_file(argv[4]);
```
mtrojan-ub, line 211: There is no real need for a summary file => if it is just about a handful of statistics at the end, please just write them to stdout.

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -1,8 +1,10 @@
 /** \file add_authority_wikidata_ids.cc
  *  \brief functionality to acquire wikidata id corresponding to their gnds
  *  \author andreas-ub
+ *  \author Steven Lolong (steven.lolong@uni-tuebingen.de)
  *
  *  \copyright 2021 Universitätsbibliothek Tübingen.  All rights reserved.
+ *  \copyright 2022 Universitätsbibliothek Tübingen.  All rights reserved.
```
mtrojan-ub, line 7: There should only be 1 `\copyright` related line. For multiple years, please merge them together like this:
`\copyright 2021-2022 Universitätsbibliothek Tübingen.  All rights reserved.`

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -18,14 +20,56 @@
...
 
+/**
+ * ============= Generate gnd_to_wiki ===============
+ * New scenario to generate gnd_to_wiki.txt
+ * by Steven Lolong
+ * update-1: 16.12.2022
+ *
+ * Case of "generate_gnd_to_wiki"
+ * The input_txt_file is the file generated by this command:
+ *      jq -c --stream '.' < authorities-gnd-person_lds.jsonld |grep -E 'https\\:/\\/d-nb\\.info\\/gnd\\/|wikidata|wikipedia'
```
mtrojan-ub, line 44: As discussed, the comments regarding integration (jq and so on) should be moved into a bash script that actually downloads the latest version and starts the processing including all the jq commands + calling this program.

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -130,19 +181,164 @@ void ParseGndWikidataMappingFile(std::string filename,
...
+    nlohmann::json line_parsed;
+    std::string gnd_id_temp_string;
+    std::string wikidata_temp_string;
+    std::string tmp_id;
+
 
-int Main(int argc, char* argv[]) {
-    if (argc != 4)
+    while (std::getline(input_file, line)) {
+        line_parsed = nlohmann::json::parse(line);
```
mtrojan-ub, line 237: The whole `GenerateGNDAuthorityExternalRef` function is very long. Maybe you can split it up a bit, e.g. move the parsing logic of a single line in the while loop into a separate function?

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -130,19 +181,164 @@ void ParseGndWikidataMappingFile(std::string filename,
...
+    const std::string dnb_address("https://d-nb.info/gnd/");
+    const std::string wikidata_address("http://www.wikidata.org/entity/");
+    const std::string wikipedia_address("https://de.wikipedia.org/wiki/");
+    std::string gnd_id;
+    bool is_start_group = false;
+    GNDStructure gnd_data;
+    int top_level_number(-1), second_level_number(-1), total_numbers_of_gnd_id_generated(0), total_line_parsed(0),
+        total_number_of_wikidata(0), total_number_of_wikipedia(0);
+
+    const int dnb_add_str_lenght(dnb_address.length()), wikidata_address_str_lenght(wikidata_address.length());
```
mtrojan-ub, line 227: lenght => length

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -130,19 +181,164 @@ void ParseGndWikidataMappingFile(std::string filename,
...
+
+int Main(int argc, char *argv[]) {
+    if (argc < 4)
         Usage();
 
     const std::string marc_input_filename_or_create_flag(argv[1]);
     const std::string marc_output_filename_or_dnb_input(argv[2]);
     const std::string mapping_txt_filename(argv[3]);
 
+    // if (marc_input_filename_or_create_flag == "--create_mapping_file") {
```
mtrojan-ub, line 328: Please remove this block of old commented out code entirely. If somebody wants to have a look at the old code, he can still look into the git history.

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -130,19 +181,164 @@ void ParseGndWikidataMappingFile(std::string filename,
...
+bool IsThisCloseBracketForId(const std::string &url) {
+    int url_length = url.length();
+    std::string is_about(url.substr(url_length - 5, 5));
+    if (is_about.compare("about") == 0)
+        return true;
+
+    return false;
+}
+
+bool DoesItMatch(const std::string &url_based, const std::string &url_comp) {
```
mtrojan-ub, line 197: That sounds very generic, can you please change the function name to be more specific?

cpp/pipelines/add_authority_external_ref.cc
mtrojan-ub: And it seems to be very specific to the lines in the actual format => i think it would be good if you could show me the original / prepared file on your machine so i can give good feedback about it.

cronjobs/add_authority_external_ref._automation.py
```diff
@@ -0,0 +1,183 @@
...
+                util.ExecOrDie(util.Which("rm"), [
+                               "-f", f"/tmp/{input_file_name_for_add_auth}"])
+                util.ExecOrDie(util.Which("rm"), [
+                               "-f", f"/tmp/{gnd_wiki_file}"])
+
+
+try:
+    Main()
+except Exception as e:
+    print("Checking last update of file is error: " +
```
mtrojan-ub, line 182: Since Main() contains all the logic (not only checking but also downloading), the error message in the except branch is not correct.

cronjobs/add_authority_external_ref._automation.py
```diff
@@ -0,0 +1,183 @@
...
+import functools
+import shutil
+
+# this script needs:
+# 1. requests
+# 2. tqdm
+
+# Start global variable
+
+url = "https://data.dnb.de/opendata/"
```
mtrojan-ub, line 56: This should rather be called base_url or index_url

mtrojan-ub: and there is not `global` keyword referencing this variable

cronjobs/add_authority_external_ref._automation.py
```diff
@@ -0,0 +1,183 @@
...
+input_file_name_for_add_auth = "input_file_for_add_authority_external_ref.txt"
+
+
+# End global variable
+
+
+def DownloadTheFile(file_name):
+    url_path = f'{url}{file_name}'
+    target_file = f'/tmp/{file_name}'
+    dw = requests.get(url_path, stream=True)
```
mtrojan-ub, line 74: Would it be better to re-use one of the functions from util.py for downloading a single file? E.g. `WgetFetch` or `RetrieveFileByUrl`

cronjobs/add_authority_external_ref._automation.py
```diff
@@ -0,0 +1,183 @@
...
+    add_auth_ext_ref_config_file.writelines(
+        [str(successful_generate_date), "\n"])
+
+
+def IsThereANewlyReleasedFile():
+    file_name = ""
+    global config_file
+    global current_file_date_integer
+    global last_file_update_date_int
+    global newer_file_name
```
mtrojan-ub, line 102: newer_file_name is declared not used in this function at all. Is it really necessary to declare a global variable for this, or can it be declared locally where it is needed?

cronjobs/add_authority_external_ref._automation.py
```diff
@@ -0,0 +1,183 @@
...
+            add_auth_ext_ref_config_file.close
+
+            if (read_line_1.strip()):
+                last_file_update_date_int = int(read_line_1)
+
+            if (last_file_update_date_int < current_file_date_integer):
+                print(f"Last update date: {last_file_update_date_int}")
+                print(
+                    f"current version on the web: {current_file_date_integer}")
+                return True
```
mtrojan-ub, line 126: Can we return the current_file_date_integer instead of just returning True/False? This way we could also get rid of one more global variable.

cronjobs/add_authority_external_ref._automation.py
mtrojan-ub: Also: We should send some kind of email to the admins with either a success message or an error message. There is a function in util.py SendEmail that can be used for this purpose.

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -18,120 +19,173 @@
...
-// Scenario 2:
-// use converted file from scenario 1 from cpp/data to create a map during
-//  pipeline processing. The norm_data_input is extended by wikidata ids where possible
-//  and saved to 024 field indicator1:7 where wikidata id is not yet present
-//  file can be taken from /mnt/ZE020150/FID-Entwicklung/ub_tools (gnd_to_wiki.txt)
+/**
+ * Generating input file must use the jq program. Since the .jsonld file is large enough than it is a must to parse it as a stream and pipe
+ * it using grep. The complete command for this task is: jq -c --stream '.' < authorities-gnd-person_lds.jsonld |grep -E
+ * 'https\\:/\\/d-nb\\.info\\/gnd\\/|wikidata|wikipedia'
+ */
```
mtrojan-ub, line 39: Please restore Scenario 2 and change it from TXT to CSV

cronjobs/add_authority_external_ref._automation.py
steven-lolong: in these two functions, I don't know how to determine the output file (downloaded file)

cronjobs/add_authority_external_ref._automation.py
steven-lolong: I will change the logic and may be difficult to read

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -18,120 +19,201 @@
...
+            // gnd = line[0], wikidata = line[1], wikipedia = line[2]
+            wiki_elements = { StringUtil::TrimWhite(line.at(1)), StringUtil::TrimWhite(line.at(2)) };
+            gnd_to_wikidataid_and_wikipedia_link->emplace(StringUtil::TrimWhite(line.at(0)), wiki_elements);
         }
-        file.close();
-    } else
-        LOG_ERROR("input or output files could not be opened");
+
+    } catch (const std::exception &except) {
+        LOG_ERROR("CSV to MRC has error with exception: " + std::string(except.what()));
```
mtrojan-ub, line 211: This function only covers parsing the CSV file and not writing to MRC, so the error message could instead be changed to `"Error parsing CSV file: " + ...`

mtrojan-ub: (And i am not sure whether we need to catch any exceptions anyway, since ParseCSVFileOrDie will exit the whole process on error)

cronjobs/add_authority_external_ref._automation.py
mtrojan-ub: I think they will store the file in the current working directory, but i guess it would be better if we had a function that just takes two parameters e.g. the URL and the full local target file path, and downloads that. A simple way to do tat would be e.g. to add a second optional parameter to WgetFetch which also passed the target path as additional `-O ` parameter if the target path is not `None`. What do you think?

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -18,120 +19,201 @@
...
 
+    std::cout << std::endl
+              << "Total GND-ID: " << total_numbers_of_gnd_id_generated << std::endl
+              << "Total GND with Wikidata: " << total_number_of_wikidata << std::endl
+              << "Total GND with Wikipedia: " << total_number_of_wikipedia << std::endl;
+    std::cout << "Total time of computation: " << duration.count() << " second(s)" << std::endl;
 
+    input_file.close();
+    return true;
+}
```
mtrojan-ub, line 197: By the way, do you have the git hooks enabled? It seems like the amount of empty newlines before GarseGndWikidataMappingFile does not match our convention.

cronjobs/add_authority_external_ref._automation.py
```diff
@@ -0,0 +1,198 @@
...
+import shutil
+
+# this script needs:
+# 1. requests
+# 2. tqdm
+
+# Start global variable
+
+base_url = "https://data.dnb.de/opendata/"
+config_file = "/mnt/ZE020150/FID-Entwicklung/ub_tools/config_file_add_authority_ext_ref.cnf"
```
mtrojan-ub, line 58: Please note that .cnf (=> .conf) files usually have a special meaning, but this file represents a state rather than a configuration. Also, it seems to be redundant to start a filename with the prefix `config_file_...` and add the extension `.cnf` to it. Also the naming in the middle is not consistant => if the program is called `add_authority_external_ref`, we should name the file exactly the same (and not e.g. use an abbreviation like `ext` for `external`).

Maybe we should call it something like `add_authority_external_ref.state`, `add_authority_external_ref.timestamp` or something similar?

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -18,120 +19,201 @@
...
-//  and saved to 024 field indicator1:7 where wikidata id is not yet present
-//  file can be taken from /mnt/ZE020150/FID-Entwicklung/ub_tools (gnd_to_wiki.txt)
+/**
+ * There are two scenarios to fullfil the task. The second scenario depend on the file generated by the first scenario.
+ *
+ * Scenario 1:
+ * use --create_mapping_file <input_txt_file> <output_csv_file> to generate the mapping file.
+ * All processes from downloading data to generating output file is managed by the script 'add_authority_external_ref.automation.py'
+ *
+ * Generating the input file for scenario 1 must use the jq program. Since the .jsonld file is large enough than it is a must to parse it as
```
mtrojan-ub, line 42: If i understand correctly, the jq-related handling will be done in the python program, so maybe the documentation here could also mention this (it is not necessary to do this by hand).

cronjobs/add_authority_external_ref._automation.py
mtrojan-ub: I don't think so, but we can discuss it verbally if you like.

mtrojan-ub:
... and i almost forgot: the python program must also be listed in cronjobs/Makefile

cronjobs/util.py
```diff
@@ -25,8 +25,12 @@
 import urllib.request
 
 
-def WgetFetch(url: str) -> bool:
-    return True if process_util.Exec("/usr/bin/wget", [ "--quiet", url ]) == 0 else False
+def WgetFetch(url: str, output: str = "None") -> bool:
```
mtrojan-ub, line 28: `None` is a keyword in python so you should be able to use it without quotes

cronjobs/util.py
```diff
@@ -85,7 +89,8 @@ def RetrieveFileByURL(url: str, timeout: int, accepted_content_types: List[str]
     return RetrieveFileByURLReturnCode.TIMEOUT
 
 
-default_email_recipient = "johannes.riedl@uni-tuebingen.de"
+# default_email_recipient = "johannes.riedl@uni-tuebingen.de"
```
mtrojan-ub, line 92: please restore the default value

cpp/pipelines/add_authority_external_ref.cc
```diff
@@ -18,120 +19,194 @@
...
+ */
+
 [[noreturn]] void Usage() {
     ::Usage(
-        "    :\n"
-        "     invocation modes:\n"
-        "     1.)   norm_data_marc_input norm_data_marc_output mapping_txt_file\n"
-        "     2.)   --create_mapping_file dnb_input_unzipped_file mapping_txt_file\n");
+        "\n\n"
+        "1) norm_data_marc_input <norm_data_marc_output> <mapping_txt_file>\n"
```
mtrojan-ub, line 52: norm_data_marc_input should be in brackets as well

cronjobs/add_authority_external_ref.py
```diff
@@ -0,0 +1,178 @@
...
+                print("Process 6/7 -- Copy the gnd_wiki_file to its destination")
+                util.ExecOrDie(util.Which("cp"), [
+                               "-f", f"/tmp/{gnd_wiki_file}", f"{share_folder}{gnd_wiki_file}"])
+
+                #  e. Update the latest version date on the config file (the second line)
+                print(
+                    "Process 7/7 -- Updating the config file and remove temporary files")
+                UpdateConfigFile(
+                    config_file, current_file_date_int)
+                util.ExecOrDie(util.Which("rm"), [
```
mtrojan-ub, line 153: Maybe in these cases it would be easier to use os.remove() instead of calling an external program (rm) to remove the file

cronjobs/add_authority_external_ref.py
```diff
@@ -0,0 +1,178 @@
...
+                           "Successfully generated gnd_wiki_file.", priority=1)
+    else:
+        util.SendEmail("gnd_wiki_file generator",
+                       "The file version on the web is the same or older.", priority=5)
+
+
+try:
+    Main()
+except Exception as e:
+    util.SendEmail("gnd_wiki_file generator", "An unexpected error occurred: "
```
mtrojan-ub, line 177: gnd_wiki_file generator => maybe we should replace the subject of the email by: `os.path.basename(__file__)`

cronjobs/add_authority_external_ref.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def Main():
+    share_folder = "/mnt/ZE020150/FID-Entwicklung/ub_tools/"
+    input_file_name_for_add_auth = "input_file_for_add_authority_external_ref.txt"
+    gnd_wiki_file = "gnd_to_wiki.csv"
+
+    if len(sys.argv) != 2:
+        util.SendEmail(os.path.basename(sys.argv[0]),
+                       "This script needs to be called with an email address and the system type!\n", priority=1)
```
mtrojan-ub, line 115: since the system type is not used, the message should be changed

PR #4320, opened 2022-12-20
Riedl

jriedl:


cpp/data/system_updates/127.sh
```diff
@@ -0,0 +1,26 @@
+# Insert is_publisher provided also to the HTML cache
+#!/bin/bash
```
mtrojan-ub, line 2: I thought that `#!/bin/bash` always needs to be in the first line of a .sh file or else it won't have any effect and will just be treated as a regular comment

cronjobs/clean_file_sessions.sh
```diff
@@ -0,0 +1,24 @@
...
+# This script cleans the systemd-private-XXX-apache-YYY/tmp/vufind_sessions direction from all session older than three days and not a logged in session (i.e. session_file size < 1K)
+
+if [[ $# != 1 ]]; then
+    echo "usage: $0 notification_email_address"
+    exit 1
+fi
+
+#cf. https://serverfault.com/questions/786211/access-files-in-system-tmp-directory-when-using-privatetmp
+apache_pid=$(systemctl show --property=MainPID --value apache2.service)
+nsenter -t $apache_pid -m find /tmp/vufind_sessions/ -mindepth 1 -maxdepth 1 -mtime +3 -size -1000 -delete
```
mtrojan-ub, line 12: Should we also:
a) check that this directory really exists before cleaning it up?
b) Also clean up the real /tmp/vufind_sessions directory if it exists (outside the sandbox)?

mtrojan-ub:
clean_file_sessions.sh and collect_journal_stats.sh should also be registered in the crontabs (templates + system updates)

cpp/installer.cc
```diff
@@ -326,6 +326,22 @@ void CreateVuFindDatabases(const VuFindSystemType vufind_system_type, DbConnecti
...
+        const std::string krim_translations_username(translations_ini_section->getString("sql_username"));
+        const std::string krim_translations_password(translations_ini_section->getString("sql_password"));
+        if (not db_connection_root->mySQLDatabaseExists(krim_translations_database)) {
+            Echo("creating " + krim_translations_database + " database");
+            db_connection_root->mySQLCreateDatabase(krim_translations_database);
+            db_connection_root->mySQLGrantAllPrivileges(krim_translations_database, krim_translations_username);
+            db_connection_root->mySQLGrantAllPrivileges(krim_translations_database, sql_username);
+            db_connection_root->mySQLGrantAllPrivileges(krim_translations_database, ub_tools_username);
+            db_connection_root->mySQLGrantGrantOption(krim_translations_database, ub_tools_username);
+            DbConnection::MySQLImportFile(INSTALLER_DATA_DIRECTORY + "/krim_translations.sql", krim_translations_database,
```
mtrojan-ub, line 342: If this is the same principle as for ixtheo, maybe we should just call it `krimdok.sql` instead of `krim_translations.sql`?
And what about system updates? (or do we do it by hand? then we must make sure that we don't forget ub55)

cpp/pipelines/collect_journal_stats.sh
```diff
@@ -0,0 +1,59 @@
...
+function Usage {
+    echo "usage: $0 ixtheo|krimdok"
+    exit 1
+}
+
+if [ $# != 1 ]; then
+    Usage
+fi
+
+system_type="$1"
```
mtrojan-ub, line 25: Should we use the TUEFIND_FLAVOUR environment variable instead?

cpp/pipelines/collect_journal_stats.sh
```diff
@@ -0,0 +1,59 @@
...
+rm -f "${log}"
+
+CreateTemporaryNullDevice
+echo -e "\n\nUsing \"${title_file}\" as input for ${system_type}" | tee --append "${log}"
+
+OVERALL_START=$(date +%s.%N)
+
+# Note: It is necessary to run this phase after articles have had their journal's PPN's inserted!
+StartPhase "Populate the Zeder Journal Timeliness Database Table"
+(collect_journal_stats ${system_type} ${title_file} /dev/null >> "${log}" 2>&1 && \
```
mtrojan-ub, line 54: I can't really remember why collect_journal_stats.cc is creating an output file anyway. Maybe we should just remove the writing logic as well as the output parameter from it?

(also it seems like TMP_NULL is created but never used, since /dev/null is used here )

PR #4327, opened 2023-01-11
Riedl

jriedl:


cpp/data/system_updates/129.sh
```diff
@@ -0,0 +1,27 @@
...
+cronjobs_template=/usr/local/ub_tools/cpp/data/installer/${TUEFIND_FLAVOUR}.cronjobs
+
+collect_journal_stats_line=$(grep collect_journal_stats ${cronjobs_template})
+
+if [ $(echo ${collect_journal_stats_line} | wc --lines) -ne 1 ]; then
+    echo "Invalid number of candidate lines in ${cronjobs_template}"
+    exit 1
+fi
+
+if  grep collect_journal_stats ${root_crontab}; then
```
mtrojan-ub, line 14: Careful, the root_crontab file might not exist on all machines (especially developer machines) so we usually check if the file exists before we do any changes, see e.g. `124.sh`

PR #4333, opened 2023-01-18
Riedl

jriedl:


cpp/convert_krim_keyword_csv_to_marc.cc
```diff
@@ -0,0 +1,156 @@
...
+
+enum column_name {
+    BEGRIFF = 2,
+    HÄUFIGKEIT,
+    PPN,
+    GND_aut,
+    GND_man,
+    ÜBERBEGRIFF,
+    PICA3,
+    WÖRTERTRENUNG_1,
```
mtrojan-ub, line 60:
```
WÖRTERTRENUNG_1
WÖRTERTRENNUNG2
WÖRTERTRENNUNG_3
```

1: missing N
2: missing _

cpp/data/system_updates/131.ixtheo.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE keyword_translations ADD COLUMN priority_entry TINYINT(1) DEFAULT false;
```
mtrojan-ub, line 1: We could add `AFTER german_updated` to maintain the same order as the installer (so later diffs are easier if we change multiple columns)

PR #4346, opened 2023-02-09
add a new feature '--replace-field-if-regex' in marc_augmentor

steven-lolong:
Example:
- marc_augmentor input_file_name output_file_name --replace-field-if-regex '007:/^(.*)$/HELLO_\1/' '007:'
- marc_augmentor input_file_name output_file_name --replace-field-if-regex '007:/^(.*)$/HELLO_\1/' '007:tu'

### DIFF
--- a/cpp/marc_tools/marc_augmentor.cc
+++ b/cpp/marc_tools/marc_augmentor.cc
@@ -398,13 +398,11 @@ bool ReplaceFieldRegex(MARC::Record * const record, const MARC::Tag &tag, const 
         if (field.getTag() != tag)
             continue;
 
-        std::string text_to_replace;
-        std::string field_value(field.getContents());
         MARC::Subfields subfields(field.getContents());
-
         if (subfields.hasSubfield(subfield_code))
             LOG_ERROR("Regex Replacement with subfield not supported! Please use '--replace-subfield-if-regex' instead.");
 
+        std::string field_value(field.getContents());
         bool global(HasGlobalFlag(replacement_regex));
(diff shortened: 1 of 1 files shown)


cpp/marc_tools/marc_augmentor.cc
```diff
@@ -373,6 +386,39 @@ bool ReplaceField(MARC::Record * const record, const MARC::Tag &tag, const char
...
+        ThreadSafeRegexMatcher replace_matcher(pattern_and_replacement[0]);
+        const std::string replacement_text(replace_matcher.replaceWithBackreferences(field_value, pattern_and_replacement[1], global));
+
+        field.setContents(replacement_text);
+        replaced_at_least_one = true;
+    }
+
+    return replaced_at_least_one;
+}
+
```
jriedl, line 421: This function is very similar to ReplaceSubfieldRegex(). Maybe it is possible to factor out some aspects or to unify and reduce the addition of subfields to the addition of fields.

cpp/marc_tools/marc_augmentor.cc
```diff
@@ -373,6 +386,39 @@ bool ReplaceField(MARC::Record * const record, const MARC::Tag &tag, const char
...
+            return false;
+    }
+
+    bool replaced_at_least_one(false);
+    for (auto &field : *record) {
+        if (field.getTag() != tag)
+            continue;
+
+        MARC::Subfields subfields(field.getContents());
+        if (subfields.hasSubfield(subfield_code))
```
jriedl, line 402: Subfields::Subfields() has a constructor with string but this will garble up things, if the contents of the have fields have no subfields as it is the case with control fields.

PR #4358, opened 2023-03-06
add new small tool to copy field for remid

steven-lolong:


cpp/harvest/remid.sh
```diff
@@ -30,10 +31,13 @@ marc_filter $FILE_TEMP $FILE_TEMP_CLEAN --remove-fields '852:  \x1FaDE-Tue135' \
     --remove-fields '935:  \x1Faremi\x1F2LOK'  
 
 # add new tag(s) to remid
-marc_augmentor $FILE_TEMP_CLEAN $FILE_NEW  --insert-field '852:  \x1FaDE-Tue135' \
+marc_augmentor $FILE_TEMP_CLEAN $FILE_TEMP1  --insert-field '852:  \x1FaDE-Tue135' \
     --insert-field '084:  \x1Fa0\x1F2ssgn' \
     --insert-field '935:  \x1Faremi\x1F2LOK'
 
+# copy field
+marc_field_copy_for_remit_purposes $FILE_TEMP1 $FILE_NEW
```
mtrojan-ub, line 39: remit => remid

cpp/marc_tools/marc_field_copy_for_remit_purposes.cc
```diff
@@ -0,0 +1,88 @@
+/** \brief A tool for adding keywords extracted from titles to MARC records.
```
mtrojan-ub, line 1: This program is specific to remid, and if we have to do some additional changes in the data, we will also put it in there. This means:
1) The file should be moved from cpp/marc_tools tp cpp/harvest
2) It doesn't make much sense to call it something like "copy" in the name, please keep it more generic, e.g. `marc_field_copy_for_remit_purposes.cc` => `remid.cc` or `remid_augment.cc`
3) the \brief section seems to be a copy from another program, please change it to something more appropriate

cpp/marc_tools/marc_field_copy_for_remit_purposes.cc
```diff
@@ -0,0 +1,88 @@
...
+} // namespace
+
+int Main(int argc, char **argv) {
+    if (argc < 3)
+        Usage();
+
+
+    const std::string input_filename(argv[1]);
+    const std::string output_filename(argv[2]);
+    const std::string filter_field("084");
```
mtrojan-ub, line 52: It's OK to leave it like this, just some small hints for the future:
1) Usually it's not necessary to declare constants for tag / subfield names if they are only used once (especially if it is likely that more tag operations are added later in the same program).
2) speaking of `filtering` usually means that we throw either tags or entire records away completely if certain fields exist, which is not the case here, so the variable names might be misleading.

PR #4372, opened 2023-03-31
add issn_lookup_in_bnb.py

steven-lolong:
Look for ISSN in the data compound if found then generate a new file containing the record(s) of it.

### DIFF
--- a/cpp/harvest/issn_lookup_in_bnb.py
+++ b/cpp/harvest/issn_lookup_in_bnb.py
@@ -1,0 +1,82 @@
+#!/bin/python3
+#
+#    @brief  The automation script to organise process of checking whether the issn is in k10plus if so then get the data and put it in output file.
+#           The output file format is mrc.
+#
+#    @author Steven Lolong (steven.lolong@uni-tuebingen.de)
+#
+#    Copyright (C) 2023 Library of the University of Tübingen
+#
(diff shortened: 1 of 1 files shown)


cpp/harvest/issn_lookup.py
```diff
@@ -0,0 +1,82 @@
...
+    yaz_client.sendline("open " + base_url)
+    yaz_client.sendline("marccharset " + marc_charset)
+    yaz_client.sendline("format " + output_format)
+
+    return yaz_client
+
+
+def Main():
+    if len(sys.argv) != 3:
+        print("[usage]\n\tpython issn_lookup_in_bnb.py input_file output_file")
```
mtrojan-ub, line 44: Maybe you could add some more information to the usage
- input file should be txt with 1 ISSN per line
- output file will be mrc with all journals found for these ISSNs

cpp/harvest/issn_lookup.py
```diff
@@ -0,0 +1,82 @@
...
+
+    yaz_client.sendline("exit")
+
+    input_file.close()
+
+
+try:
+    Main()
+except Exception as e:
+    print(e + "\n \n" + traceback.format_exc(20))
```
mtrojan-ub, line 82: Do we also need to call `sys.exit(-1)` after printing the exception here?

PR #4374, opened 2023-03-31
add marc_issn_lookup

steven-lolong:


cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,171 @@
...
+#include "FileUtil.h"
+#include "MARC.h"
+#include "util.h"
+
+namespace {
+[[noreturn]] void Usage() {
+    ::Usage(
+        "marc_input_articles marc_input_journals marc_output_articles"
+        "\n"
+        "- marc_input_artiles is a file containing all article information taken from CORE. \n"
```
mtrojan-ub, line 33: typo3: `marc_input_artiles` => `marc_input_articles`

Also, please remove whitespace before `\n`

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,171 @@
...
+        "- marc_input_journals is a file containing journal information. Please use issn_lookup.py to generate this file. \n"
+        "- marc_output_articles is an output file generated by this tool. \n");
+
+    std::exit(EXIT_FAILURE);
+}
+struct SubFieldInfo {
+    std::string i;
```
mtrojan-ub, line 40: According to our internal coding convention, all class/struct member variable names must end with an underscore, e.g. `i` => `i_` and so on.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,171 @@
...
+    std::string w;
+    std::string x;
+    bool is_found;
+
+    SubFieldInfo(bool b) { this->is_found = b; }
+
+    void Builder(MARC::Record &record) {
+        for (auto &field : record) {
+            if (field.getTag() == "001")
+                this->w = "(DE-627)" + field.getContents();
```
mtrojan-ub, line 51: After following the earlier mentioned convention to rename the struct members e.g. from `w` to `w_`, it is no longer necessary to access them via `this->...`, you can access them directly instead (`w_ = ...`)

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }
+        ++onprogress_counter;
+        output_file->write(record);
+    }
+}
+} // end of namespace
+
+int Main(int argc, char **argv) {
+    if (argc < 4)
```
mtrojan-ub, line 164: Since we always expect a fix amount of arguments, the check should be `if (argc != 4)`

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,171 @@
...
+void ISSNLookup(char **argv) {
+    auto input_file(MARC::Reader::Factory(argv[1]));
+    auto output_file(MARC::Writer::Factory(argv[3]));
+    int onprogress_counter(1);
+    std::string ppn("");
+
+    std::cout << "Updating in progress...\n";
+    while (MARC::Record record = input_file->read()) {
+        std::cout << "Record - " << onprogress_counter << "\r";
+        ppn = "";
```
mtrojan-ub, line 140: The ppn variable doesn't seem to be used at all - can it be removed?

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,171 @@
...
+
+        for (auto &field : record) {
+            if (field.getTag() == "773") {
+                const std::string issn(field.getFirstSubfieldWithCode('x'));
+                if (issn != "") {
+                    SubFieldInfo sub_field_info(LookupISSNInDataSouce(argv[2], issn));
```
mtrojan-ub, line 146: In our sample data, there are about 3.500 ISSNs (1.500 are unique). If `LookupISSNInDataSouce` is called for every ISSN, this means that the whole `marc_input_journals` file will be read multiple thousand times. I think it would be better to read the file only once (outside the loop) and create some kind of cache, like e.g. a std::map<std::string, SubFieldInfo> where the key is the ISSN to lookup.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,171 @@
...
+    }
+    return false;
+}
+
+SubFieldInfo LookupISSNInDataSouce(const std::string input_journal_filename, const std::string issn) {
+    auto data_source(MARC::Reader::Factory(input_journal_filename));
+    SubFieldInfo sub_field_info = SubFieldInfo(false);
+
+    while (MARC::Record record = data_source->read()) {
+        for (auto &field : record) {
```
mtrojan-ub, line 93: Can this be `const auto`?

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,171 @@
...
+                    this->t = "";
+            }
+        }
+        this->i = "In:";
+        this->is_found = true;
+    }
+};
+
+bool IsOnlineISSN(const MARC::Record &record) {
+    for (auto &field : record) {
```
mtrojan-ub, line 76: Can this also be const auto?

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,171 @@
...
+        if (field.getTag() == "300") {
+            const std::string sub_field_a(field.getFirstSubfieldWithCode('a'));
+            if (sub_field_a == "Online-Ressource")
+                return true;
+            else
+                return false;
```
mtrojan-ub, line 82: 300 is a repeatable field: https://www.loc.gov/marc/bibliographic/bd300.html

So in theory there could be multiple 300. I think it would be better to just remove the `else return false` here. This way the function will return true if it finds any `300a == "Online-Ressource"`, else it will return false after the loop.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,171 @@
...
+
+SubFieldInfo LookupISSNInDataSouce(const std::string input_journal_filename, const std::string issn) {
+    auto data_source(MARC::Reader::Factory(input_journal_filename));
+    SubFieldInfo sub_field_info = SubFieldInfo(false);
+
+    while (MARC::Record record = data_source->read()) {
+        for (auto &field : record) {
+            if (field.getTag() == "022") {
+                const std::string sub_field_a(field.getFirstSubfieldWithCode('a'));
+                if ((sub_field_a != "") && (sub_field_a == issn)) {
```
mtrojan-ub, line 96: Here as well: `(sub_field_a != "")` => `(not sub_field_a.empty())`. Same a few lines below for `sub_field_l`.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,171 @@
...
+            if (sub_field_a == "Online-Ressource")
+                return true;
+            else
+                return false;
+        }
+    }
+    return false;
+}
+
+SubFieldInfo LookupISSNInDataSouce(const std::string input_journal_filename, const std::string issn) {
```
mtrojan-ub, line 88: What you currently do: Find the first journal for this ISSN, check whether it's print or online, and return the corresponding information.

BUT: You can have 10 journals for the same ISSN, and the 7th could be the online journal.
So that's what i meant: you have to put all journal candidates into a set and then check it in the corresponding priority:
1) Return the online journal if you have exact 1
2) Return a print journal if you have exact 1
3) return nothing (is_found = false)

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,171 @@
...
+                    }
+                }
+            }
+        }
+    }
+    return sub_field_info;
+}
+
+void UpdateSubfield(MARC::Subfields &subfields, const SubFieldInfo &sub_field_info) {
+    if (!subfields.replaceFirstSubfield('i', sub_field_info.i))
```
mtrojan-ub, line 118: I think there is an existing function `insertOrReplaceSubfield` on a MARC::Field which should be more efficient here.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,189 @@
...
+    void Builder(MARC::Record &record) {
+        for (auto &field : record) {
+            if (field.getTag() == "001")
+                w_ = "(DE-627)" + field.getContents();
+
+            if (field.getTag() == "245") {
+                MARC::Subfields subfields(field.getSubfields());
+                std::string subfield_a(subfields.getFirstSubfieldWithCode('a'));
+                std::string subfield_b(subfields.getFirstSubfieldWithCode('b'));
+                if (subfield_a != "" && subfield_b != "")
```
mtrojan-ub, line 59: Please also check against empty string with `string.empty()` and not `string.empty()` and so on

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,190 @@
...
+            journal_cache[sfi.first].is_valid_ = false;
+        else if (sfi.second.online_version_counter_ == 0)
+            if (sfi.second.printed_version_counter_ == 1)
+                journal_cache[sfi.first].is_valid_ = true;
+            else
+                journal_cache[sfi.first].is_valid_ = false;
+    }
+}
+
+std::map<std::string, SubFieldInfo> JournalCacheBuilder(const std::string &input_journal_filename) {
```
mtrojan-ub, line 111: `BuildJournalCache` might be a better name, else we would assume its the constructor of a factory class.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,190 @@
...
+    auto input_journal_file(MARC::Reader::Factory(input_journal_filename));
+    int record_counter(0);
+
+    std::cout << "Build a cache for journal \n";
+    while (MARC::Record record = input_journal_file->read()) {
+        SubFieldInfo sub_field_info_of_record;
+        sub_field_info_of_record.Builder(record);
+
+        // if sub_info is exist in the cache
+        if (auto sub_info_search = journal_cache.find(sub_field_info_of_record.x_); sub_info_search != journal_cache.end()) {
```
mtrojan-ub, line 122: Do we really need the variable `sub_info_search` or can we just remove it?

mtrojan-ub: `if (journal_cache.find(sub_field_info_of_record.x_) != journal_cache.end()) {`

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,190 @@
...
+            if (journal_cache[sub_field_info_of_record.x_].is_online_) {
```
mtrojan-ub, line 123: Maybe we should move the increasing logic of the statistics out of these blocks to avoid similar code in separate branches?

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,190 @@
...
+                else if (subfield_a != "" && subfield_b == "")
+                    t_ = subfield_a;
+                else if (subfield_a == "" && subfield_b != "")
+                    t_ = subfield_b;
+                else
+                    t_ = "";
+            }
+
+            if (field.getTag() == "300")
+                ((field.getFirstSubfieldWithCode('a') == "Online-Ressource") ? is_online_ = true : is_online_ = false);
```
mtrojan-ub, line 70: This can be simplified:
`is_online_ = (field.getFirstSubfieldWithCode('a') == "Online-Ressource");`

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -0,0 +1,190 @@
...
+
+
+            if (field.getTag() == "773")
+                x_ = field.getFirstSubfieldWithCode('x');
+        }
+
+        i_ = "In:";
```
mtrojan-ub, line 77: Do we really need to store `i_` in the cache if it's always the same hardcoded value `In:` for all thousands of entries? (remove memory usage and move hardcoded value to the `UpdateSubfield` function)

PR #4378, opened 2023-04-06
update logic in marc_issn_lookup

steven-lolong:


cpp/harvest/marc_issn_lookup.cc
```diff
@@ -43,6 +43,7 @@ struct SubFieldInfo {
     std::string t_;
     std::string w_;
     std::string x_;
+    std::string l_; // this is for reference issn from main
```
mtrojan-ub, line 46: l_ must be changed to issns_ which is e.g. a std::set<std::string>

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -106,70 +110,103 @@ void UpdateSubfield(MARC::Subfields &subfields, const SubFieldInfo &sub_field_in
...
-    LOG_INFO("Online (valid): " + std::to_string(online_valid) + ", " + "Online (invalid): " + std::to_string(online_invalid) + ", "
-             + "Print (valid): " + std::to_string(print_valid) + ", " + "Print (invalid): " + std::to_string(print_invalid));
+    // LOG_INFO("Online (valid): " + std::to_string(online_valid) + ", " + "Online (invalid): " + std::to_string(online_invalid) + ", "
+    //          + "Print (valid): " + std::to_string(print_valid) + ", " + "Print (invalid): " + std::to_string(print_invalid));
 }
 
 
 std::map<std::string, SubFieldInfo> BuildJournalCache(const std::string &input_journal_filename) {
-    std::map<std::string, SubFieldInfo> journal_cache;
+    std::map<std::string, SubFieldInfo> journal_cache, online_issn_cache, printed_issn_cache;
```
mtrojan-ub, line 180: As discussed, the journal_cache should be changed to a std::set<SubFieldInfo>

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -43,6 +44,8 @@ struct SubFieldInfo {
     std::string t_;
     std::string w_;
     std::string x_;
+    std::string l_;
```
mtrojan-ub, line 47: I think that there is no real use in keeping the separate `l_` property.
Moreover, i think we should rename `list_of_l_` to `issns_`.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -105,80 +114,184 @@ void UpdateSubfield(MARC::Subfields &subfields, const SubFieldInfo &sub_field_in
...
-    std::map<std::string, SubFieldInfo> journal_cache;
+std::vector<SubFieldInfo> BuildJournalCache(const std::string &input_journal_filename) {
+    std::vector<SubFieldInfo> journal_cache;
     auto input_journal_file(MARC::Reader::Factory(input_journal_filename));
-    int record_counter(0);
+
 
     while (MARC::Record record = input_journal_file->read()) {
         SubFieldInfo sub_field_info_of_record(record);
+        bool is_exist_in_journal_cache(false);
```
mtrojan-ub, line 136: `is_exist_in_journal_cache` should be renamed to `exists_in_journal_cache`.

To find an existing element, i think it would be easier to just take 022a as well as 022l directly out of the record (without generating sub_field_info_of_record), and check if one of them is already contained in `list_of_l_` (= `issns_`).

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -105,80 +114,184 @@ void UpdateSubfield(MARC::Subfields &subfields, const SubFieldInfo &sub_field_in
...
+            if (elemt.x_ == sub_field_info_of_record.x_) {
+                if (elemt.is_online_) {
+                    if (sub_field_info_of_record.is_online_) {
+                        // data is in cache already
+                        is_exist_in_journal_cache = true;
+                        ++elemt.online_version_counter_;
+                        elemt.is_valid_ = false;
+                        continue;
+                    }
+                    if (not sub_field_info_of_record.is_online_) {
```
mtrojan-ub, line 148: You can just use `else` in all of these cases without negating the `if` condition explicitly

cpp/harvest/marc_issn_lookup.cc
mtrojan-ub: I think the main problem is that SubFieldInfo has become a hybrid struct which is used for 2 different purposes:
1) To represent the data we store in the cache
2) As an intermediate data structure where information about a record is stored so we can share parts of the parsing logic with 1+

Should we split that up into some kind of hierarchy? Let's discuss!

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -43,6 +44,7 @@ struct SubFieldInfo {
     std::string t_;
     std::string w_;
     std::string x_;
+    std::vector<std::string> issns; // this is for reference issn from main
```
mtrojan-ub, line 47: Since it's a class property, the variable name should still end with an underscore => `issns_`

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -106,79 +115,91 @@ void UpdateSubfield(MARC::Subfields &subfields, const SubFieldInfo &sub_field_in
...
+        bool exist_in_journal_cache(false);
+
+        for (auto &elemt : journal_cache) {
+            if (IsInISSNs(sub_field_info_of_record.issns, elemt.issns)) {
+                if (elemt.is_online_) {
+                    if (sub_field_info_of_record.is_online_) {
+                        ++elemt.online_version_counter_;
+                        InsertIfNotExist(sub_field_info_of_record.issns, elemt.issns);
+                        elemt.is_valid_ = false;
+                        exist_in_journal_cache = true;
```
mtrojan-ub, line 163: `exist_in_journal_cache` is set to true in all cases, so we can move it 2 levels higher and only set it once.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -106,79 +115,91 @@ void UpdateSubfield(MARC::Subfields &subfields, const SubFieldInfo &sub_field_in
...
+    for (const auto &issn : issns_input) {
+        if (std::find(issns.begin(), issns.end(), StringUtil::ASCIIToUpper(issn)) != issns.end())
+            return true;
+    }
+    return false;
+}
+
+
+// avoiding duplication in issns's cache
+void InsertIfNotExist(const std::vector<std::string> &issns_input, std::vector<std::string> &issns) {
```
mtrojan-ub, line 139: Actually, our convention is: If a variable is modified inside a function, we always pass a pointer instead of a reference. So the second argument should be:
`std::vector<std::string> &issns` => `std::vector<std::string> * const issns`

This also increases the code readability at the points where the function is called.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -106,79 +115,91 @@ void UpdateSubfield(MARC::Subfields &subfields, const SubFieldInfo &sub_field_in
...
+
+        for (auto &elemt : journal_cache) {
+            if (IsInISSNs(sub_field_info_of_record.issns, elemt.issns)) {
+                if (elemt.is_online_) {
+                    if (sub_field_info_of_record.is_online_) {
+                        ++elemt.online_version_counter_;
+                        InsertIfNotExist(sub_field_info_of_record.issns, elemt.issns);
+                        elemt.is_valid_ = false;
+                        exist_in_journal_cache = true;
+                        continue;
```
mtrojan-ub, line 164: Also, it seems like the `continue` statements have basically no effect since in none of the cases there will be any other statements behind them inside the for loop, so you might as well remove all of them.

mtrojan-ub: maybe we should rather replace it by a `break` statement and also move it 2 levels higher?

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -91,6 +104,46 @@ struct SubFieldInfo {
...
+    std::cout << "******** End of Cache ***********" << std::endl;
+}
+
+
+bool IsInISSNs(const std::string &issn, const std::vector<std::string> &issns) {
+    return (std::find(issns.begin(), issns.end(), StringUtil::ASCIIToUpper(issn)) != issns.end() ? true : false);
+}
+
+
+bool IsInISSNs(const std::vector<std::string> &issns, const std::vector<std::string> &issns_input) {
```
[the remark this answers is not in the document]
mtrojan-ub: Or as an alternative you could use use MiscUtil::Intersect, depending on what you want to achieve.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -99,86 +152,123 @@ void UpdateSubfield(MARC::Subfields &subfields, const SubFieldInfo &sub_field_in
...
-        } else if (sfi.second.online_version_counter_ == 0)
-            if (sfi.second.printed_version_counter_ == 1) {
-                journal_cache[sfi.first].is_valid_ = true;
-                LOG_INFO(sfi.first + ": Print (valid)");
-                ++print_valid;
-            } else {
-                journal_cache[sfi.first].is_valid_ = false;
-                LOG_INFO(sfi.first + ": Print (invalid)");
-                ++print_invalid;
+std::vector<SubFieldInfo> MergeIssn(const std::vector<SubFieldInfo> &journal_cache) {
```
mtrojan-ub, line 196: 1) Instead of passing a reference and creating a copy, could we instead just pass a pointer and do the operations directly on the cache? (it can have multiple thousand entries and might be slow to work on a copy)

2) MergeIssn => maybe we could find a better function name like e.g. `MergeDuplicateSubFieldInfos`

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -99,86 +152,123 @@ void UpdateSubfield(MARC::Subfields &subfields, const SubFieldInfo &sub_field_in
...
+
+        for (auto &elemt : journal_cache) {
+            if (IsInISSNs(sub_field_info_of_record.issns_, elemt.issns_)) {
+                exist_in_journal_cache = true;
+                if (elemt.is_online_) {
+                    if (sub_field_info_of_record.is_online_) {
+                        InsertIfNotExist(sub_field_info_of_record.issns_, &elemt.issns_);
+                        elemt.is_valid_ = false;
+                    } else
+                        InsertIfNotExist(sub_field_info_of_record.issns_, &elemt.issns_);
```
mtrojan-ub, line 245: This can be simplified even more => if you call InsertIfNotExist in if AND else you might move it 1 level higher as well and remove the empty else branch entirely.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -223,8 +304,9 @@ int Main(int argc, char **argv) {
     if (argc != 4)
         Usage();
 
-    std::map<std::string, SubFieldInfo> journal_cache(BuildJournalCache(argv[2]));
+    std::vector<SubFieldInfo> journal_cache(BuildJournalCache(argv[2]));
     ISSNLookup(argv, journal_cache);
 
+    PrettyPrintCache(journal_cache);
```
mtrojan-ub, line 310: Maybe we could add an optional flag `[--verbose]` in the usage and only print the cache if it is set.

PR #4382, opened 2023-04-20
update remid.sh

steven-lolong:
Limit the number of records per download

### DIFF
--- a/cpp/harvest/remid.sh
+++ b/cpp/harvest/remid.sh
@@ -21,8 +21,12 @@ declare -r FILE_NEW="ixtheo_remid_`date +'%y%m%d'`_001.xml"
 declare -r TEMP_FILE_NAME_PREFIX="tmp_file"
 declare -r CLEAN_FILE_NAME_PREFIX="clean_file"
 
+declare -r URL_PART_1="http://sru.hebis.de/sru/DB=2.1?query=pica.abr+%3D+%22REMID%22+or+pica.prv+%3D+%22REMID%22&version=1.1&operation=searchRetrieve&recordSchema=marc21&maximumRecords="
+declare -r URL_PART_2="&startRecord="
+declare -r URL_PART_3="&recordPacking=xml&sortKeys=LST_Y%2Cpica%2C0%2C%2C"
+
 # maximum number of records per download 
-declare -r RECORDS_PER_CALL=100
+declare -r RECORDS_PER_CALL=500
 
 
 function ColorEcho {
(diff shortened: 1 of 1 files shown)


cpp/harvest/remid.sh
```diff
@@ -1,47 +1,123 @@
...
-# also: insert <collection> root element at start/end.
-sed -i '1 a <collection xmlns="http://www.loc.gov/MARC21/slim" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.loc.gov/standards/marcxml/schema/MARC21slim.xsd">' $FILE_TEMP
+    TempFile="${TEMP_FILE_NAME_PREFIX}_${Counter}.xml"
+    CleanFile="${CLEAN_FILE_NAME_PREFIX}_${Counter}.xml"
 
-echo '</collection>' >> $FILE_TEMP
+    StartRecord=$(( ( Counter - 1) * RECORDS_PER_CALL + 1 ))
+    ColorEcho "Downloading start from record ${StartRecord}"
+    # download some records into the file
+    wget "http://sru.hebis.de/sru/DB=2.1?query=pica.abr+%3D+%22REMID%22+or+pica.prv+%3D+%22REMID%22&version=1.1&operation=searchRetrieve&recordSchema=marc21&maximumRecords=${RECORDS_PER_CALL}&startRecord=${StartRecord}&recordPacking=xml&sortKeys=LST_Y%2Cpica%2C0%2C%2C" -O $TempFile
```
mtrojan-ub, line 62: Please put the first part of the URL into a constant and then just concatenate startRecord & maximumRecords parameters to that here

cpp/harvest/remid.sh
```diff
@@ -1,47 +1,123 @@
...
+# the output file
 declare -r FILE_NEW="ixtheo_remid_`date +'%y%m%d'`_001.xml"
 
-# Download MARC XML from Hebis via SRU
+# related to file name pattern
+declare -r TEMP_FILE_NAME_PREFIX="tmp_file"
+declare -r CLEAN_FILE_NAME_PREFIX="clean_file"
+
+# maximum number of records per download 
+declare -r RECORDS_PER_CALL=100
```
mtrojan-ub, line 25: Please increase this to 500 and add the information to the comment above that 1000 is max

cpp/harvest/remid.sh
```diff
@@ -1,47 +1,123 @@
...
+
+    (( Counter++ ))
+done
+
+# merge all clean files
+ColorEcho "++++ Merge all clean files ++++"
+
+Counter=1
+echo '<?xml version="1.0" encoding="UTF-8" ?>' >> $FILE_ORI
+while [ $Counter -le $NumOfIteration ]
```
mtrojan-ub, line 79: Maybe it would be more efficient to merge this second loop into the first and concat directly after downloading

PR #4383, opened 2023-04-21
update remid_augment, add serial_splitter

steven-lolong:


cpp/harvest/remid_augment.cc
```diff
@@ -20,36 +20,63 @@
...
+    auto output_file_marc(MARC::Writer::Factory(argv[3]));
+    std::set<std::string> issns;
+    bool serial;
+    std::ofstream issn_file(argv[4]);
 
-int Main(int argc, char **argv) {
-    if (argc < 3)
-        Usage();
+    while (MARC::Record record = input_file->read()) {
+        serial = false;
```
mtrojan-ub, line 48: In these cases you can just move the declaration here as well to save 1 line of code above:
`bool serial(false);`

cpp/harvest/remid_augment.cc
```diff
@@ -20,63 +20,69 @@
...
 
 namespace {
 
 [[noreturn]] void Usage() {
-    std::cerr << "Usage: " << ::progname << " marc_input marc_output\n";
+    std::cerr << "Usage: " << ::progname << "  marc_input marc_output issn_output\n";
     std::cerr << "       marc_input is the marc input file\n";
-    std::cerr << "       marc_output is the marc output file\n";
+    std::cerr << "       marc_output is the marc output file without serial\n";
+    std::cerr << "       issn_output is the text output file for issn that ha\n";
```
mtrojan-ub, line 32: Regarding `issn_output`: This is technically wrong, because the output file will not contain the ISSNs => it will contain the ZDB IDs. I think it would be better to rename the parameter to `journal_output` or `serial_output`. The usage should then state that the journal_output contains the ZDB IDs as a text file and that all journals will be removed from the regular marc_output file.

cpp/harvest/remid_augment.cc
mtrojan-ub: This is still open, even after the changes

cpp/harvest/remid_augment.cc
```diff
@@ -20,63 +20,69 @@
...
-        ++total_record;
-        if (check_field == marc_record.end())
-            continue;
-
-        for (const auto &field : marc_record.getTagRange(filter_field)) {
-            if (field.getTag() != filter_field)
-                continue;
+        for (auto &field : record.getTagRange("035")) {
+            const std::string issn(field.getFirstSubfieldWithCode('a'));
+            if (StringUtil::ASCIIToUpper(issn.substr(0, 11)) == "(DE-599)ZDB") {
```
mtrojan-ub, line 57: Instead of hardcoding the 11 here you could just count the strlen of the prefix that you compare against, to avoid hardcoding

PR #4391, opened 2023-05-04
update some logic and log info of cache - marc_issn_lookup.cc 

steven-lolong:


cpp/harvest/marc_issn_lookup.cc
```diff
@@ -73,15 +77,21 @@ struct SubFieldInfo {
...
-                InsertIssnIfNotExist(x_, &issns_);
+                if (not field.getFirstSubfieldWithCode('a').empty()) {
+                    x_ = StringUtil::ASCIIToUpper(field.getFirstSubfieldWithCode('a'));
+                    InsertIssnIfNotExist(x_, &issns_);
+                }
 
-                if (not field.getFirstSubfieldWithCode('l').empty())
+                if (not field.getFirstSubfieldWithCode('l').empty()) {
+                    if (x_.empty())
+                        x_ = StringUtil::ASCIIToUpper(field.getFirstSubfieldWithCode('l'));
```
mtrojan-ub, line 92: I think this is dangerous since it will change lots of results

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -57,11 +58,14 @@ struct SubFieldInfo {
     std::string t_;
```
mtrojan-ub, line 58: As discussed, i think it would be better to rename SubFieldInfo to CacheEntry and use actual property names instead of the short values in this struct.

mtrojan-ub: That looks a lot better, thanks!

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -53,32 +54,35 @@ void InsertIfNotExist(const std::vector<std::string> &issns_input, std::vector<s
...
 
 
-struct SubFieldInfo {
-    std::string t_;
-    std::string w_;
-    std::string x_;
+struct CacheEntry {
+    std::string title_;
+    std::string preferred_issn_;
+    std::string ppn_;
```
mtrojan-ub, line 60: should title_ and ppn_ also be changed to preferred_title_ and preferred_ppn_?

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -105,25 +109,37 @@ struct SubFieldInfo {
...
-void PrettyPrintSubFieldInfo(const SubFieldInfo &sfi) {
-    std::cout << "t: " << sfi.t_ << std::endl;
-    std::cout << "w: " << sfi.w_ << std::endl;
-    std::cout << "x: " << sfi.x_ << std::endl;
-    std::cout << "online: " << (sfi.is_online_ ? "yes" : "no") << std::endl;
-    std::cout << "valid: " << (sfi.is_valid_ ? "yes" : "no") << std::endl;
+void PrettyPrintCacheEntry(const CacheEntry &ce) {
+    std::cout << "ppn: " << ce.ppn_ << std::endl;
+    std::cout << "title (t): " << ce.title_ << std::endl;
+    std::cout << "w: "
```
mtrojan-ub, line 115: i think we can remove w from debug output since it's basically the ppn which we output 2 lines above.

PR #4393, opened 2023-06-05
split the sub field of tag 264

steven-lolong:


cpp/harvest/remid_augment.cc
```diff
@@ -55,6 +55,33 @@ int Main(int argc, char **argv) {
...
+        std::vector<MARC::Record::Field> new_fields_264;
+
+        for (auto &field : record.getTagRange("264")) {
+            char last_subfield_code = 'a';
+            const char ind1 = field.getIndicator1();
+            const char ind2 = field.getIndicator2();
+            MARC::Record::Field new_field(MARC::Tag("264"), ind1, ind2);
+
+            for (auto &sub_field : field.getSubfields()) {
+                if (sub_field.code_ == 'a')
```
mtrojan-ub: Please add brackets to the outer if-block here.

PR #4397, opened 2023-06-12
update solr plugin for tuefind90

steven-lolong:


cpp/Makefile.inc
```diff
@@ -20,9 +20,9 @@ else
 endif
 
 ifeq ($(CCC),clang++)
-  CCCFLAGS += -std=gnu++2a -Wno-vla-extension #-Wno-c++1y-extensions
+  CCCFLAGS += -std=gnu++20 -Wno-vla-extension #-Wno-c++1y-extensions
```
mtrojan-ub, line 23: It looks like we have 22.04 related changes in this 20.04 related branch. Is it really a good option to merge it now? We won't be able to compile it unless we migrate all of our machines to 22.04.

PR #4401, opened 2023-06-14
add core_issn_lookup

steven-lolong:


cpp/harvest/.gitignore
```diff
@@ -2,3 +2,4 @@ core
 marc_issn_lookup
 nacjd
 remid_augment
+core_issn_lookup
```
mtrojan-ub, line 5: Please don't forget moving the cc file and the .gitignore entry from `cpp/harvest/core_issn_lookup.cc` to `cpp/issn_lookup.cc`

cpp/lib/include/IssnLookup.h
```diff
@@ -0,0 +1,82 @@
...
+ * \param issn single ISSN
+ */
+void ExtractingData(ISSNInfo * const issn_info, const nlohmann::json &issn_info_json, const std::string issn);
+
+
+/**
+ * \brief Show the content of ISSNInfo structure as debuging purpose
+ * \param issn_info the variable of ISSNInfo
+ */
+void PrettyPrintISSNInfo(const ISSNInfo &issn_info);
```
mtrojan-ub, line 80: I think it might be clearer to move the PrettyPrint function directly into the struct as a member function.

cpp/lib/include/IssnLookup.h
```diff
@@ -0,0 +1,82 @@
...
+    }
+};
+
+
+/**
+ * \brief Get the ISSN info from https://portal.issn.org/
+ * \param issn  single ISSN
+ * \param issn_info is used to save the raw issn info
+ */
+bool GetISSNInfo(const std::string issn, nlohmann::json &issn_info);
```
mtrojan-ub, line 64: I think the only public function in IssnLookup.h should have the following signature:
`bool GetISSNInfo(const std::string &issn, ISSNInfo * const issn_info);`

cpp/lib/include/IssnLookup.h
```diff
@@ -0,0 +1,61 @@
...
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <iostream>
+#include <nlohmann/json.hpp>
```
mtrojan-ub, line 24: Since the user of this interface will not interact with the json library at all, you might as well remove the include here so it's only included in IssnLookup.cc

cpp/lib/src/IssnLookup.cc
```diff
@@ -0,0 +1,120 @@
...
+                }
+            }
+        }
+    }
+}
+
+bool GetISSNInfo(const std::string &issn, ISSNInfo * const issn_info) {
+    const std::string issn_url("https://portal.issn.org/resource/ISSN/" + issn + "?format=json");
+
+    Downloader downloader(issn_url, Downloader::Params(), TIMEOUT_IN_SECONDS * 1000);
```
mtrojan-ub, line 80: As far as i know, you can remove the timeout & params arguments from the function call so the default values will be taken (also no need to override the default timeout, so you can also remove the constant definition entirely.)

cpp/lib/src/IssnLookup.cc
```diff
@@ -0,0 +1,120 @@
...
+
+    if (downloader.anErrorOccurred()) {
+        const HttpHeader http_header(downloader.getMessageHeader());
+        LOG_ERROR("Error while downloading data for issn " + issn + ": " + downloader.getLastErrorMessage() + ", HTTP status code "
+                  + std::to_string(http_header.getStatusCode()) + "!");
```
mtrojan-ub, line 85: Instead of calling getMessageHeader() and after that getStatusCode(), you can just call downloader.getResponseCode() for the same result.

On the other hand: imagine the error has accured during a TCP connection problem (e.g. general networking problem), then you will not be able to get any HTTP related details.

cpp/harvest/core_issn_lookup.cc
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+#include <iostream>
+#include "IssnLookup.h"
+#include "util.h"
+
+
+[[noreturn]] void Usage() {
+    ::Usage(
+        "issn [--verbose]\n"
```
mtrojan-ub, line 28: It's good that you renamed the parameter to `--verbose`, since this similar to many of our other programs.
For even bigger similarity, you could move the `--verbose` parameter to the beginning.

PR #4404, opened 2023-06-16
Add issn.org as a searching source of marc_issn_lookup

steven-lolong:


cpp/lib/include/IssnLookup.h
```diff
@@ -38,9 +38,9 @@ struct ISSNInfo {
         type_,               // type
         issn_,               // issn
         is_part_of_,         // isPartOf
-        publication_,        // publication
-        url_;                // url
-    std::vector<std::string> names_;
+        publication_;        // publication
+
+    std::vector<std::string> names_, url_;
```
mtrojan-ub, line 43: If we name a variable and it may contain multiple values, it's good to name it with the plural form
url_ => urls_

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -320,14 +372,40 @@ void ISSNLookup(char **argv, std::vector<CacheEntry> &journal_cache) {
...
+                    if (not is_issn_in_k10plus) {
+                        IssnLookup::ISSNInfo issn_info;
+                        if (IsInISSNInfoCache(issn, *issn_org_cache, &issn_info)) {
+                            // issn is in the issn info cache already
                             MARC::Subfields subfields(field.getSubfields());
-                            UpdateSubfield(subfields, elemt);
+                            UpdateSubfieldUsingISSNOrg(subfields, issn_info);
                             field.setSubfields(subfields);
+                        } else {
+                            if (IssnLookup::GetISSNInfo(issn, &issn_info)) {
```
mtrojan-ub, line 393: If the ISSN is not in the cache, we should check whether there has been a failed lookup before and avoid doing multiple failed lookups for the same ISSN in that case (e.g. by checking ppns_with_issn_not_recognized before?)

PR #4411, opened 2023-06-22
Update marc_issn_lookup

steven-lolong:
Use info from issn.org if issn is invalid in k10plus, use memory for debugging if needed, and reduce the number of parameters in ISSNLookup and ShowInfoForDebugging by creating a DebugInfo structure.
Update variable types in IssnLookup.h


cpp/lib/src/IssnLookup.cc
```diff
@@ -35,16 +35,29 @@ void ExtractingData(const std::string &issn, const nlohmann::json &issn_info_jso
                 if (ar.at("@id") == issn_uri) {
                     issn_info->issn_ = issn;
                     for (auto &[key, val] : ar.items()) {
-                        if (key == "mainTitle")
-                            issn_info->main_title_ = val;
+                        if (key == "mainTitle") {
+                            if (val.is_structured())
+                                for (auto val_item : val)
```
mtrojan-ub, line 40: Usually in all for-loops we should use references as well as const whenever possible
`for (const auto &val_item : val)`

PR #4413, opened 2023-06-23
Add the capability of deleting duplication of field 773 based on its title.

steven-lolong:


cpp/harvest/marc_issn_lookup.cc
```diff
@@ -395,6 +396,22 @@ void CleanDuplicationOfField773(MARC::Record * const record) {
     }
 }
 
+void CleanDuplicationOfField773ByTitle(MARC::Record * const record) {
+    std::vector<std::string> titles;
+    const ControlNumberGuesser control_number_guesser;
```
mtrojan-ub, line 401: ControlNumberGuesser::NormaliseTitle is a static function, so it should be possible to call it directly without initializing an instance of the class first.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -395,6 +396,22 @@ void CleanDuplicationOfField773(MARC::Record * const record) {
...
+    for (auto field(record->begin()); field != record->end(); ++field) {
+        if (field->getTag() == "773") {
+            const std::string title(control_number_guesser.NormaliseTitle(field->getFirstSubfieldWithCode('t')));
+            if ((std::find(titles.begin(), titles.end(), title)) != titles.end())
```
mtrojan-ub, line 405: Right now this logic is keeping the first 773 entry with the same title and throwing away the others. But in fact, we cannot rely on the first being the best, so we must check for the first occurrence with that title which actually has a w subfield, keep that one and remove all the others with the same title.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -395,6 +396,22 @@ void CleanDuplicationOfField773(MARC::Record * const record) {
     }
 }
 
+void CleanDuplicationOfField773ByTitle(MARC::Record * const record) {
+    std::vector<std::string> titles;
+    const ControlNumberGuesser control_number_guesser;
+    for (auto field(record->begin()); field != record->end(); ++field) {
+        if (field->getTag() == "773") {
```
mtrojan-ub, line 403: Please also make sure that you only check 773 with the correct indicators.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -50,6 +50,47 @@ struct DebugInfo {
     DebugInfo() = default;
 };
 
+struct TitleInfo773 {
+    std::string title_, subfield_w_;
+    char ind1_, ind2_;
```
mtrojan-ub, line 55: indicators should not be stored in the cache but checked during iteration. only 0 8 is relevant.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -50,6 +50,47 @@ struct DebugInfo {
...
+    TitleInfo773() = default;
+
+    bool IsEqual(const TitleInfo773 &ti) const {
+        if ((title_ == ti.title_) && (ind1_ == ti.ind1_) && (ind2_ == ti.ind2_) && (subfield_w_ == ti.subfield_w_))
+            return true;
+
+        return false;
+    }
+
+    bool CompareButIgnoreW(const TitleInfo773 &ti) const {
```
mtrojan-ub, line 66: If you have functions doing similar things, it is always a good thing to name them in a similar way.
e.g.
IsEqual
CompareButIgnoreW => IsEqualWithoutW

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -50,6 +50,47 @@ struct DebugInfo {
...
+        if ((title_ == ti.title_) && (ind1_ == ti.ind1_) && (ind2_ == ti.ind2_))
+            return true;
+
+        return false;
+    }
+
+    void Update(const MARC::Record::Field &field) {
```
mtrojan-ub, line 73: Instead of writing a separate function with a unique name it might be better to convert this into a constructor

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -396,22 +436,49 @@ void CleanDuplicationOfField773ByISSN(MARC::Record * const record) {
...
-            if ((std::find(titles.begin(), titles.end(), title)) != titles.end())
-                record->erase(field);
-            else {
-                titles.emplace_back(title);
-                field->getSubfields().replaceFirstSubfield('t', title);
+            TitleInfo773 ti;
+            ti.Update(*field);
+            if (not ti.subfield_w_.empty()) {
+                if (IsInTitleInfo773Cache(*found_title_info_cache, ti, false))
+                    record->erase(field);
```
mtrojan-ub, line 446: Can you really erase fields while iterating over them without invalidating the iterator?

PR #4420, opened 2023-07-04
Ubuntu2204 slolong

mtrojan-ub:


cpp/data/installer/scripts/install_ubuntu_packages.sh
```diff
@@ -1,4 +1,15 @@
 #!/bin/bash
+
+##################
+# script for installation of based-machine
+# os: ubuntu 22.04
+# by Steven Lolong (steven.lolong@uni-tuebingen.de)
```
mtrojan-ub, line 6: Most changes in this file must be reverted

cpp/installer.cc
```diff
@@ -223,14 +225,20 @@ void MountDeptDriveAndInstallSSHKeysOrDie(const VuFindSystemType vufind_system_t
     const std::string GITHUB_ROBOT_PRIVATE_KEY_LOCAL(SSH_KEYS_DIR_LOCAL + "github-robot");
     const std::string GITHUB_ROBOT_PUBLIC_KEY_REMOTE(SSH_KEYS_DIR_REMOTE + "github-robot.pub");
     const std::string GITHUB_ROBOT_PUBLIC_KEY_LOCAL(SSH_KEYS_DIR_LOCAL + "github-robot.pub");
+    const std::string copy_file_syntax(ExecUtil::Which("cp"));
     if (not FileUtil::Exists(SSH_KEYS_DIR_LOCAL))
         FileUtil::MakeDirectoryOrDie(SSH_KEYS_DIR_LOCAL, false, 0700);
     if (not FileUtil::Exists(GITHUB_ROBOT_PRIVATE_KEY_LOCAL)) {
-        FileUtil::CopyOrDie(GITHUB_ROBOT_PRIVATE_KEY_REMOTE, GITHUB_ROBOT_PRIVATE_KEY_LOCAL);
+        const std::vector<std::string> exec_param{ GITHUB_ROBOT_PRIVATE_KEY_REMOTE, GITHUB_ROBOT_PRIVATE_KEY_LOCAL };
+        // disable using this function for a while until the patch the error in 'copy_file_range()' in new linux kernel
```
mtrojan-ub, line 233: These changes must be reverted & CopyOrDie must be fixed before merging

mtrojan-ub: Instead of using a cp command every time FileUtil::CopyOrDie is called, we should fix the CopyOrDie function internally instead

cpp/installer.cc
```diff
@@ -381,15 +408,20 @@ void RegisterSystemUpdateVersion() {
 
 
 static void GenerateAndInstallVuFindServiceTemplate(const VuFindSystemType system_type, const std::string &service_name) {
+    Echo("Installer -> generating and install vufind service template");
     FileUtil::AutoTempDirectory temp_dir;
 
     Template::Map names_to_values_map;
     names_to_values_map.insertScalar("solr_heap", system_type == KRIMDOK ? "4G" : "8G");
+    // names_to_values_map.insertScalar("solr_heap", system_type == KRIMDOK ? "4G" : "4G");
```
mtrojan-ub, line 416: please remove this comment

cpp/pipelines/ixtheo_marc_pipeline_fifo.sh
```diff
@@ -176,17 +176,17 @@ wait
...
                                          GesamtTiteldaten-post-phase"$PHASE"-"${date}".mrc >> "${log}" 2>&1 && \
 EndPhase || Abort) &
 
 
 # Note: It is necessary to run this phase after articles have had their journal's PPN's inserted!
-StartPhase "Populate the Zeder Journal Timeliness Database Table"
-(collect_journal_stats ixtheo GesamtTiteldaten-post-phase"$((PHASE-1))"-"${date}".mrc \
-                              GesamtTiteldaten-post-phase"$PHASE"-"${date}".mrc >> "${log}" 2>&1 && \
-EndPhase || Abort) &
+# StartPhase "Populate the Zeder Journal Timeliness Database Table"
```
mtrojan-ub, line 186: This phase should be re-enabled again in git (you can disable it locally if necessary for testing, but please dont change it in git)

docker/ixtheo/my_vu.cnf
```diff
@@ -0,0 +1,3 @@
+[client]
```
mtrojan-ub, line 1: i guess that these docker changes should also be removed from git

virtualMachine/README.md
```diff
@@ -0,0 +1,31 @@
+# Pre
```
mtrojan-ub, line 1: Your local virtualMachine settings should also be removed from git

cpp/Makefile.inc
```diff
@@ -20,9 +20,9 @@ else
 endif
 
 ifeq ($(CCC),clang++)
-  CCCFLAGS += -std=gnu++2a -Wno-vla-extension #-Wno-c++1y-extensions
+  CCCFLAGS += -std=gnu++20 -Wno-vla-extension #-Wno-c++1y-extensions
```
mtrojan-ub, line 23: This has already changed in `ubuntu2204` branch. Please execute `git pull origin ubuntu2204` and solve all merge conflicts before i can continue the review.

cpp/data/installer/scripts/install_ubuntu_packages.sh
```diff
@@ -19,7 +23,7 @@ ColorEcho "installing/updating ub_tools dependencies..."
 
 apt-get --yes update
 
-# install additional libraries for docker environment
+# install additional
```
mtrojan-ub, line 26: part of this comment got lost

cpp/data/installer/scripts/install_ubuntu_packages.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
+
+cd /
```
mtrojan-ub, line 3: Is there a specific reason to add `cd /`?

cpp/data/installer/scripts/install_ubtools.sh
```diff
@@ -5,6 +5,21 @@
 # compile cpp installer
 # run it
 
+# ./install_ubtools.sh parameter
```
mtrojan-ub, line 8: These additional CLI options should not be documented here, because they will just be passed through to the installer. This is documented in the wiki. If you want to add a comment here, please just add a comment that refers to either the wiki page or installer.cc.

cpp/data/installer/scripts/install_ubtools.sh
```diff
@@ -5,6 +5,21 @@
...
+#   - vufind < ixtheo | krimdok >  <--test | --production> [--omit-cronjobs] [--omit-systemctl] [--omit-systemctl] 
+#  
+# example:
+# ./install_ubtools.sh vufind ixtheo --test --omit-cronjobs | tee ~/output.log
+
+function ColorEcho {
+    echo -e "\033[1;34m" $1 "\033[0m"
+}
+
+ColorEcho "*************** The installation script is running *******************"
```
mtrojan-ub, line 21: If you really want to add ColorEcho and add comments like this, we should use it on multiple occurences. E.g. either we divide this script into sections and call it at every section head, or we change all  echo calls to use ColorEcho instead.

cpp/data/installer/scripts/install_ubuntu_packages.sh
```diff
@@ -1,4 +1,8 @@
...
+export JAVA_TOOL_OPTIONS="-Dfile.encoding=UTF8"
```
mtrojan-ub, line 4: Should we also add this to `/etc/profile.d/vufind.sh`? see installer.cc => SetVuFindEnvironmentVariables

cpp/installer.cc
```diff
@@ -143,12 +143,13 @@ TemporaryChDir::~TemporaryChDir() {
 
 
 void GitActivateCustomHooks(const std::string &repository) {
+    Echo("Installer -> git activate custom hooks");
```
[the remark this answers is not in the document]
mtrojan-ub: There are still 12 more references where this should be changed

docker/ixtheo/apache2/vufind-vhosts.conf
```diff
@@ -1,6 +1,6 @@
 ServerName localhost
 DocumentRoot /usr/local/vufind/public
-ScriptAlias /cgi-bin/ /var/www/cgi-bin/
+#ScriptAlias /cgi-bin/ /var/www/cgi-bin/
```
mtrojan-ub, line 3: The original value should be kept

cpp/lib/src/FileUtil.cc
```diff
@@ -1140,6 +1143,40 @@ void CopyOrDie(const std::string &from_path, const std::string &to_path) {
         LOG_ERROR("failed to copy \"" + from_path + "\" to \"" + to_path + "\"!");
 }
 
+void CopyOrDieXFs(const std::string fromPath, const std::string toPath) {
+    int src;          /* file descriptor for source file */
```
mtrojan-ub, line 1147: This function should just use the CP shell command

cpp/lib/include/FileUtil.h
```diff
@@ -503,6 +504,12 @@ bool Copy(const std::string &from_path, const std::string &to_path, const bool t
           const loff_t offset = 0, const int whence = SEEK_CUR);
 void CopyOrDie(const std::string &from_path, const std::string &to_path);
 
+/**
+ * \brief Copy file that support cross filesystem's type. The maximum size of copy file is 2GB only.
```
mtrojan-ub, line 508: Now the implementation uses cp, so the size limitation can be removed from the comment.

PR #4422, opened 2023-07-05
MARC21 Änderung 2024.1 #2462

LysogorAndGmail:


cpp/lib/src/BSZUtil.cc
```diff
@@ -226,14 +227,37 @@ void ExtractYearVolumeIssue(const MARC::Record &record, std::string * const year
...
+            }
+            break;
+        }
+    }
+
+    *volume = filtered_dates[0];
+    *issue = filtered_dates[2];
 }
 
+std::string explode(const std::string &data, const std::string &delimiters) {
```
mtrojan-ub, line 248: There is no need to write a custom explode function. Please use StringUtil::Split() instead.

cpp/lib/src/BSZUtil.cc
```diff
@@ -226,14 +227,37 @@ void ExtractYearVolumeIssue(const MARC::Record &record, std::string * const year
...
+    if (field_773 == record.end())
         return;
 
-    *volume = field_936->getFirstSubfieldWithCode('d');
-    *issue = field_936->getFirstSubfieldWithCode('e');
+    std::vector<std::string> filtered_dates;
+    for (const auto &field : record.getTagRange("773")) {
+        if (field.getIndicator1() == '1') {
+            for (const auto &subfield : field.getSubfields()) {
+                filtered_dates.emplace_back(explode(subfield.value_, ":"));
```
mtrojan-ub, line 238: As discussed, all functions like this should be backwards-compatible. You replaced the existing 936 logic by a 773 logic, but the function should still try to find 936 in case 773 does not exist as a fallback.

cpp/find_latest_journal_year.cc
```diff
@@ -183,6 +190,19 @@ void ProcessRecords(MARC::Reader * const marc_reader, std::unordered_map<std::st
     LOG_INFO("Processed " + std::to_string(record_count) + " MARC record(s).");
 }
 
+std::string explode(const std::string &data, const std::string &delimiters) {
```
mtrojan-ub, line 193: Same here, please just use StringUtil::Split() instead of writing a new explode function.
(Another note: If you ever have any need for such a general function, please implement it within a re-usable library instead of adding it to a single script.)

cpp/zotaut/dump_delivered_marc_records.cc
```diff
@@ -79,19 +79,36 @@ void GetJournalEntriesFromDb(DbConnection * const db_connection, const std::stri
...
-                csv_row += ";" + _936_field.getFirstSubfieldWithCode('j');
-                csv_row += ";" + _936_field.getFirstSubfieldWithCode('d');
-                csv_row += ";" + _936_field.getFirstSubfieldWithCode('e');
-                break;
+            const auto _773_field(record.findTag("773"));
+            const auto g_773_contents(_773_field->getFirstSubfieldWithCode('g'));
+            if (not g_773_contents.empty()) {
+                std::vector<std::string> subfields;
+                std::vector<std::string> filtered_dates;
+                for (const auto &field : record.getTagRange("773")) {
```
mtrojan-ub, line 87: This iteration should be improved because you call findTag() and getTagRange() twice on the same 773 field. And you should check the indicator before trying to get any subfields.

cpp/zotaut/dump_delivered_marc_records.cc
```diff
@@ -79,19 +79,36 @@ void GetJournalEntriesFromDb(DbConnection * const db_connection, const std::stri
...
+                            StringUtil::Split(subfield.value_, ':', &subfields, true);
+                            filtered_dates.emplace_back(subfields[1]);
+                        }
+                    }
+                }
+                csv_row += ";" + filtered_dates[1];
+                csv_row += ";" + filtered_dates[0];
+                csv_row += ";" + filtered_dates[2];
+            } else {
+                for (const auto &_936_field : record.getTagRange("936")) {
```
mtrojan-ub, line 99: Also, please add correct indicator checks to all existing cases regarding 936 where they are missing

cpp/find_latest_journal_year.cc
```diff
@@ -157,12 +157,28 @@ void ProcessRecords(MARC::Reader * const marc_reader, std::unordered_map<std::st
...
-        for (const auto &field : record.getTagRange("936")) {
-            const MARC::Subfields subfields(field.getSubfields());
-            year_as_string = subfields.getFirstSubfieldWithCode('j');
-            if (not year_as_string.empty())
-                break;
+
+        std::vector<std::string> subfields;
+        std::vector<std::string> filtered_dates;
+        for (const auto &field : record.getTagRange("773")) {
+            if (field.getIndicator1() == '1') {
```
mtrojan-ub, line 164: You must not only check whether indicator1 is 1, but also whether indicator2 is 8, see issue

cpp/find_latest_journal_year.cc
```diff
@@ -157,12 +157,28 @@ void ProcessRecords(MARC::Reader * const marc_reader, std::unordered_map<std::st
...
+                for (const auto &subfield : field.getSubfields()) {
```
mtrojan-ub, line 165: - Instead of iterating all subfields, you should only iterate subfields with the code `g`.
- You should only call `emplace_back` if the key before the `:` equals `year`.

cpp/find_latest_journal_year.cc
```diff
@@ -157,12 +157,28 @@ void ProcessRecords(MARC::Reader * const marc_reader, std::unordered_map<std::st
             continue;
 
         std::string year_as_string;
-        for (const auto &field : record.getTagRange("936")) {
-            const MARC::Subfields subfields(field.getSubfields());
-            year_as_string = subfields.getFirstSubfieldWithCode('j');
-            if (not year_as_string.empty())
-                break;
+
+        std::vector<std::string> subfields;
```
mtrojan-ub, line 161: - `subfields` is a bad variable name here because it will not contain entire subfields. instead it contains splitted parts of a subfield value, so you should find a better name
- variables should always be declared at the latest possible point. if you compare this with filtered_dates, that one is valid during the whole loop so it is ok to declare filtered_dates before the loop, but since subfields will be reset when calling StringUtil::Split() you should move the declaration next to that line.

cpp/find_latest_journal_year.cc
```diff
@@ -157,12 +157,28 @@ void ProcessRecords(MARC::Reader * const marc_reader, std::unordered_map<std::st
...
+        for (const auto &field : record.getTagRange("773")) {
+            if (field.getIndicator1() == '1') {
+                for (const auto &subfield : field.getSubfields()) {
+                    StringUtil::Split(subfield.value_, ':', &subfields, true);
+                    filtered_dates.emplace_back(subfields[1]);
+                }
+            }
         }
+        if (not filtered_dates.empty()) {
+            year_as_string = filtered_dates[1];
```
mtrojan-ub, line 172: I don't really understand the logic at this point.
Usually filtered_dates contains only 1 value, but then you must use filtered_dates[0].
On the other hand, the name of this program is "find_latest_journal_year". That means that if filtered_dates contain multiple values, you must detect the biggest number out of all values in the container.
Another option is: I doubt that there will be multiple 773 fields with the same indicators, so you could as well remove the whole filtered_dates variable and write the resullt to year_as_string instead and then just use the existing logic for that variable.

cpp/lib/src/BSZUtil.cc
```diff
@@ -221,19 +222,58 @@ void ExtractArchiveMembers(const std::string &archive_name, std::vector<std::str
 }
 
 
-void ExtractYearVolumeIssue(const MARC::Record &record, std::string * const year, std::string * const volume, std::string * const issue) {
-    const auto field_008(record.findTag("008"));
-    if (field_008 != record.end())
-        *year = field_008->getContents().substr(7, 4);
+IssueInfo ExtractYearVolumeIssue(const MARC::Record &record) {
+    IssueInfo issue_info;
+
```
mtrojan-ub, line 227: 008 still missing here

cpp/pipelines/collect_journal_stats.cc
```diff
@@ -186,21 +187,11 @@ void CollectZederArticles(MARC::Reader * const reader,
...
-        std::string issue(_936_field->getFirstSubfieldWithCode('e'));
-        if (issue.empty())
-            issue = _936_field->getFirstSubfieldWithCode('d');
-        else
-            volume = _936_field->getFirstSubfieldWithCode('d');
-        const std::string year(_936_field->getFirstSubfieldWithCode('j'));
+        const auto issue_info(BSZUtil::ExtractYearVolumeIssue(record));
+        const std::string pages(issue_info.pages_);
+        std::string volume(issue_info.volume_);
+        std::string issue(issue_info.issue_);
```
mtrojan-ub, line 193: Special volume/issue logic still missing here

mtrojan-ub (approved):
Looks good, but we should find a good point when to merge this and run it for a couple of days on our test servers before installing it on our live servers.

cpp/zotaut/dump_delivered_marc_records.cc
```diff
@@ -79,19 +80,22 @@ void GetJournalEntriesFromDb(DbConnection * const db_connection, const std::stri
...
-            for (const auto &_936_field : record.getTagRange("936")) {
-                csv_row += ";" + _936_field.getFirstSubfieldWithCode('j');
-                csv_row += ";" + _936_field.getFirstSubfieldWithCode('d');
-                csv_row += ";" + _936_field.getFirstSubfieldWithCode('e');
-                break;
-            }
+            const auto issue_info(BSZUtil::ExtractYearVolumeIssue(record));
+
+            if (not issue_info.year_.empty())
+                csv_row += ";" + issue_info.year_;
```
mtrojan-ub, line 86: This could be a problem. Since we're exporting a CSV file here, the field delimiter `;` should always be appended even if the value is empty, else the amount of columns per csv_row will vary

cpp/zotaut/dump_delivered_marc_records.cc
```diff
@@ -79,19 +80,29 @@ void GetJournalEntriesFromDb(DbConnection * const db_connection, const std::stri
...
+
+            std::string data_volume("");
+            if (not issue_info.volume_.empty())
+                data_volume = issue_info.volume_;
+
+            std::string data_issue("");
+            if (not issue_info.issue_.empty())
+                data_issue = issue_info.issue_;
+
+            csv_row += ";" + data_year;
```
mtrojan-ub, line 97: Now that's definitely better than before, however i suggest to also use TextUtil::CSVEscape on the variables just to be sure (in theory they might contain a semicolon in some cases).

mtrojan-ub (approved):
OK now - we should merge this next monday & keep an eye on it during next week before we pull the live servers!

PR #4431, opened 2023-07-18
add 776x to cache

steven-lolong:


cpp/harvest/marc_issn_lookup.cc
```diff
@@ -148,6 +148,13 @@ struct CacheEntry {
                 else
                     preferred_title_ = "";
             }
+
+            if (field.getTag() == "776") {
+                if (not field.getFirstSubfieldWithCode('x').empty()) {
+                    preferred_issn_ = StringUtil::ASCIIToUpper(field.getFirstSubfieldWithCode('x'));
```
mtrojan-ub, line 154: I would assume that most entries have a 022 field, and in addition a 776 field.
With your suggested change, preferred_issn_ will always be overridden by 776, but this should only be done if it is still empty at this point.
And also => why didn't you add the ppn + issn to related_ppns_and_its_issns? (and by the way, that struct member variable name should end with underscore_ like the others)

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -148,6 +148,13 @@ struct CacheEntry {
                 else
                     preferred_title_ = "";
             }
+
+            if (field.getTag() == "776") {
```
mtrojan-ub, line 152: We should only consider 776 with indicators 0 and 8

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -148,6 +148,13 @@ struct CacheEntry {
...
+                if (not field.getFirstSubfieldWithCode('x').empty()) {
+                    preferred_issn_ = StringUtil::ASCIIToUpper(field.getFirstSubfieldWithCode('x'));
+                    InsertIssnIfNotExist(preferred_issn_, &issns_);
```
mtrojan-ub, line 155: Insert is OK, but please use the value from x directly (uppercase) instead of the class member.

cpp/harvest/marc_issn_lookup.cc
```diff
@@ -148,6 +148,17 @@ struct CacheEntry {
...
                     preferred_title_ = "";
             }
+
+            if (field.getTag() == "776") {
+                if (field.getIndicator1() == '0' and field.getIndicator2() == '8') {
+                    if (not field.getFirstSubfieldWithCode('x').empty()) {
+                        if (preferred_issn_.empty())
+                            preferred_issn_ = StringUtil::ASCIIToUpper(field.getFirstSubfieldWithCode('x'));
+
+                        InsertIssnIfNotExist(preferred_issn_, &issns_);
```
mtrojan-ub, line 158: You can move this statement into the if branch as well

PR #4433, opened 2023-07-21
add xml to ouput format

steven-lolong:


cpp/marc_tools/marc_grep.cc
```diff
@@ -84,7 +85,7 @@ char help_text[] =
     "\n"
     "  Output label format:\n"
     "    label_format = matched_field_or_subfield | control_number | control_number_and_matched_field_or_subfield\n"
-    "                   | no_label | marc_binary | marc_xml | control_number_and_traditional\n"
+    "                   | no_label | marc_binary | marc_xml | control_number_and_traditional | xml\n"
```
mtrojan-ub, line 88: I think we should rename this to xml_debug to make sure it is not used in production

PR #4435, opened 2023-07-25
Updating installer 

steven-lolong:
fixing errors when importing the store procedure from the SQL file


cpp/installer.cc
```diff
@@ -323,6 +323,9 @@ void CreateVuFindDatabases(const VuFindSystemType vufind_system_type, DbConnecti
         }
     }
 
+    std::string error__;
```
mtrojan-ub, line 326: - Variable names ending with underscores are reserved for class members only
- This is written to but never read

PR #4442, opened 2023-08-21
Riedl

jriedl:


cpp/new_translation_items_alert.cc
```diff
@@ -0,0 +1,233 @@
...
+    return (bool)std::atoi(result.getNextRow()["notify_time_exceeded"].c_str());
+}
+
+
+void UpdateLastNotifiedTo(DbConnection * const db_connection, const std::string &user, const std::string new_last_notified,
+                          const bool debug = false) {
+    if (debug)
+        return;
+
+    const std::string update_statement("UPDATE translators SET last_notified ='" + new_last_notified + "' WHERE translator='" + user + "'");
```
mtrojan-ub, line 183: If there are just a few couple of seconds between notification & DB update, we could also pass NOW() here without quotes to save another query if that is exact enough. But since you already implemented it maybe we just leave it as-is.

mtrojan-ub:
Do we also need to add a cronjob in the installer crontab(s) and system updates?

PR #4450, opened 2023-09-11
adding TitleNormalization function

steven-lolong:
remove unwanted hidden code in title

### DIFF
--- a/cpp/lib/src/IssnLookup.cc
+++ b/cpp/lib/src/IssnLookup.cc
@@ -22,19 +22,28 @@
 #include <nlohmann/json.hpp>
 #include "Downloader.h"
 #include "IssnLookup.h"
+#include "StringUtil.h"
 
 
 namespace IssnLookup {
 
 std::string TitleNormalization(const std::string &title) {
     std::string new_title;
+    /*
+     * Remove non-readable characters used by the librarian to annotate the unused string/word in the title for some purposes.
+     * The non-readable characters are start string code (FFFFFFC2 and FFFFFF98) and string delimiter code (FFFFFFC2 and FFFFFF9C).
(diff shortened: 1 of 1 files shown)


cpp/lib/src/IssnLookup.cc
```diff
@@ -26,7 +26,15 @@
 
 namespace IssnLookup {
 
+std::string TitleNormalization(const std::string &title) {
+    std::string new_title;
 
+    for (const auto &c : title)
+        if (!((int(c) == -62) || (int(c) == -104) || (int(c) == -100)))
```
mtrojan-ub, line 33: Please add documentation about why we normalize & what these characters are

cpp/lib/src/IssnLookup.cc
```diff
@@ -22,10 +22,27 @@
...
+    std::string new_title;
+    /*
+     * Remove non-readable characters used by the librarian to annotate the unused string/word in the title for some purposes.
+     * The non-readable characters are start string code (FFFFFFC2 and FFFFFF98) and string delimiter code (FFFFFFC2 and FFFFFF9C).
+     * The looping below will remove the special character from the title.
+     */
+    for (const char &c : title) {
+        std::string str_hex_code = StringUtil::ToHexString(unsigned(c));
+
+        if (!((str_hex_code == "FFFFFFC2") || (str_hex_code == "FFFFFF98") || (str_hex_code == "FFFFFF9C")))
```
mtrojan-ub, line 40: Basically OK, but => how costly are these operations if we execute them 32.000 * length(str) times? Do we need to optimize performance? Some ideas:
- Instead of converting every character in all titles to hex string, convert the hex codes to character only once (and maybe store them in a static array?)
- use strcmp instead of == operator

PR #4468, opened 2023-11-06
update the server address of mysql connection for  overrides in docker environment only

steven-lolong:


mtrojan-ub:
Looks good! But i just realized that there is no database_solrmarc.conf for krimdok yet (though we have one on the servers), can you please add it?

PR #4469, opened 2023-11-07
add sftp connection modul for bsz_util

steven-lolong:


cronjobs/sftp_connection.py
```diff
@@ -0,0 +1,101 @@
...
+    except Exception as excp:
+      util.Error(f"failed to list directory ({excp})")
+
+
+  def downloadFile(self, remote_file_path, local_file_path=None):
+    if local_file_path is None:
+      local_file_path = remote_file_path
+
+    util.Info(f"downloading file from {remote_file_path} to {local_file_path}")
+    sftp_client = self._sftp.open_sftp()
```
mtrojan-ub, line 63: The sftp_client should be stored in a class variable, and open_sftp should be called only once in the __init__ function.

mtrojan-ub:
Looks good, thanks! However, i'm gonna convert it to draft since the SFTP server has not fully been configured yet. Also, as discussed, please add a proper system update and installer change that will add the new library (via pip?)

mtrojan-ub:
Also, please add an abstract class to make sure that ftp_connection and sftp_connection actually share the same function definitions.

cpp/data/system_updates/144.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -o errexit
+
+# install paramiko library as a dependency of sftp
+
+pip install paramiko
```
mtrojan-ub, line 6: we should also add this to the installer

PR #4493, opened 2024-01-23
New translator features

jriedl:


cpp/data/system_updates/146.ixtheo.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE keyword_translations MODIFY COLUMN status enum ('reliable_synonym','reliable','unreliable', 'unreliable_cat2', 'unreliable_synonym','replaced','replaced_synonym','new','new_synonym');
```
mtrojan-ub, line 1: Is 'macs' missing here as possible enum value? (it is still present if you compare it with installer/ixtheo.sql)

cpp/data/system_updates/147.ixtheo.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE keyword_translations ADD COLUMN translation_disabled TINYINT(1) AFTER priority_entry;
```
mtrojan-ub, line 1: Should we add e.g. DEFAULT 0 here? The field does not seem to be covered in the initial INSERT statement.

PR #4501, opened 2024-02-21
Angabe Land im Benutzerkonto (für KrimDok)

LysogorAndGmail:
Related to ubtue/tuefind#2731

### DIFF
--- a/cpp/data/system_updates/150.vufind.sql
+++ b/cpp/data/system_updates/150.vufind.sql
@@ -1,0 +1,1 @@
+ALTER TABLE vufind.user RENAME COLUMN ixtheo_country TO tuefind_country;
--- a/cpp/data/system_updates/150.vufind.sql
+++ b/cpp/data/system_updates/150.vufind.sql
@@ -1,1 +1,0 @@
-ALTER TABLE vufind.user RENAME COLUMN ixtheo_country TO tuefind_country;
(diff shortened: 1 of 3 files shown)


cpp/data/system_updates/150.vufind.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE vufind.user RENAME COLUMN ixtheo_country TO tuefind_country;
```
mtrojan-ub, line 1: I think this SQL update is pretty important and there are multiple things that should be considered.
- If you create a match named 150.vufind.sql, it will be executed on all systems. So it would also be executed on KrimDok systems where the ixtheo_country field is not yet present, which would lead to an error. To solve this, you must create 2 separate updates (e.g. 150.ixtheo.sql and 151.krim_translations.sql) that will only be executed on systems where the corresponding database is available. In the ixtheo update you must rename the column but in the krim_translations update you must create it.
- For IxTheo renaming the column is like the primary goal, but we also want to keep all tuefind related columns together, so you should add a second statement `MODIFY COLUMN ... AFTER last_language`, so that the column is moved between last_language and tuefind_institution column.
- All changes in this update should also be reflected in the corresponding installer schemes in the corresponding TueFind pull request (move the field from module/IxTheo/sql/mysql.sql to module/TueFind/sql/mysql.sql)

mtrojan-ub (approved):

PR #4508, opened 2024-04-17
Zotero legacy handling

jriedl:


cpp/zotaut/sync_legacy_records.cc
```diff
@@ -0,0 +1,279 @@
...
+
+
+auto LegacyEntryInformationHash([](const LegacyEntryInformation &A) { return std::hash<std::string>()(StringUtil::Join(A.dois_, "")); });
+auto LegacyEntryInformationEqual([](const LegacyEntryInformation &A, const LegacyEntryInformation &B) { return A.dois_ == B.dois_; });
+using LegacyEntryInformationSet =
+    std::unordered_set<LegacyEntryInformation, decltype(LegacyEntryInformationHash), decltype(LegacyEntryInformationEqual)>;
+
+
+const std::string CONF_FILE_PATH(UBTools::GetTuelibPath() + "zotero.conf");
+const std::string LEGACY_RECORD_SECTION_NAME("LegacyRecordStock");
```
mtrojan-ub, line 78: Shouldn't the logic for parsing the legacy section be moved to `ZoteroHarvesterConfig.h/cc`? Not sure whether processing the config with this section might fail in other programs if the section is unknown to the regular config parser

PR #4509, opened 2024-04-18
Adding timezone abbreviation mapping

steven-lolong:


cpp/lib/src/TimeUtil.cc
```diff
@@ -763,39 +763,271 @@ time_t UTCStructTmToTimeT(const struct tm &tm) {
...
-        *adjustment = -1 * 3600;
-    else if (rfc822_zone == "M")
-        *adjustment = -12 * 3600;
-    else if (rfc822_zone == "N")
-        *adjustment = +1 * 3600;
-    else if (rfc822_zone == "Y")
-        *adjustment = +12 * 3600;
-    else // Unrecognized tiem zone.
-        return false;
+    std::map<std::string, float> tmz_map;
```
mtrojan-ub, line 1017: This should be static and the Construct... should be called with constructor notation

PR #4510, opened 2024-04-29
Slolong tuefind91

steven-lolong:


mtrojan-ub:
Please don't forget: The ConditionalInsertField has a build error right now because the jcip-annotations-1.0.jar no longer exists in the solr/vendor/modules/extraction/lib directory.

mtrojan-ub:
Also, starting the vufind service via systemctl doesnt seem to work

PR #4514, opened 2024-05-13
Update nacjd

steven-lolong:


cpp/harvest/nacjd.cc
```diff
@@ -178,6 +178,29 @@ bool DownloadID(std::ofstream &json_new_titles, const std::string &id, const boo
     return true;
 }
 
+void GetIDFromLOK(const MARC::Record &record, std::set<std::string> * const parsed_marc_ids) {
+    for (auto field(record.begin()); field != record.end(); ++field) {
```
mtrojan-ub, line 182: If you want to iterate over LOK and local data blocks, you should use the helper functions from MARC.h: `findStartOfAllLocalDataBlocks()` and `getLocalTagRange()`.

If you need some examples, you can have a look at e.g. `normalise_urls.cc` or `krimdok_check_local_holdings.cc`.

cpp/harvest/nacjd_pipeline.sh
```diff
@@ -11,15 +11,16 @@ declare -r BSZ_FILENAME="krimdok_nacjd_$(date +%y%m%d)_001"
...
 declare -r DOWNLOAD_FILE="$DOWNLOAD_DIR/$BSZ_FILENAME.json"
 declare -r CONVERT_FILE="$DOWNLOAD_DIR/$BSZ_FILENAME.xml"
+declare -r BSZ_DIR=/usr/local/ub_tools/bsz_daten
 mkdir -p "$DOWNLOAD_DIR"
 
 # Use empty MRC file as dummy input.
 # On consecutive downloads we want to use the real KrimDok GesamtTiteldaten... file here.
-declare -r EMPTY_MRC_FILE="$DOWNLOAD_DIR/empty.mrc"
-touch "$EMPTY_MRC_FILE"
+declare -r GESAMTTITLE_DATEN = "$BSZ_DIR/GesamtTitelDaten-$(date +%y%m%d).mrc"
```
mtrojan-ub, line 19: - A better variable name would be `GESAMTTITELDATEN_FILE` or `REFERENCE_FILE`.
- You should move the declaration up next to the other declarations.

PR #4579, opened 2024-11-12
Update system call for composer

steven-lolong:
The current system call using 'ExecOrDie' will kill all sub-processes when an error occurs running the composer install. Phing will fail during installation, but Tuefind doesn't need It. Using 'system' for the system call will solve this problem.


cpp/installer.cc
```diff
@@ -809,7 +810,8 @@ void ConfigureVuFind(const bool production, const VuFindSystemType vufind_system
     Echo("Configuring vufind");
     // We need to increase default_socket_timeout for big downloads on slow mirrors, especially Solr (default 60 seconds) .
     TemporaryChDir tmp2(VUFIND_DIRECTORY);
-    ExecUtil::ExecOrDie(ExecUtil::LocateOrDie("php"), { "-d", "default_socket_timeout=600", ExecUtil::LocateOrDie("composer"), "install" });
+    system("composer install");
```
mtrojan-ub, line 813: Multiple aspects:
1) The override of default_socket_timeout is removed (that's why we initially called PHP instead of composer directly)
2) Due to our convention (see ExecUtil) system calls should be called with double colons at the start => `::system()`
3) If it is necessary to use a different system call, then we should create a wrapper function in ExecUtil to do so
4) You're just calling system() without checking the result => please check the manpage, you must check the result code against -1 and maybe also output a log message containing `errno`.

PR #4591, opened 2024-12-05
Add cross-link and remove dangling to the workflow to solve issue #3081 on Tuefind.

steven-lolong:


cpp/pipelines/krimdok_marc_pipeline.sh
```diff
@@ -172,6 +172,20 @@ EndPhase || Abort) &
...
+    GesamtTiteldaten-post-phase"$PHASE"-"${date}".mrc >> "${log}" 2>&1 && \
+EndPhase || Abort) &
+wait
+
+
+StartPhase "Remove Dangling References"
+(remove_dangling_references GesamtTiteldaten-post-phase"$((PHASE-1))"-"${date}".mrc \
+                            GesamtTiteldaten-post-phase"$PHASE"-"${date}".mrc >> "${log}" \
+                            dangling_references.log 2>&1 && \
+EndPhase || Abort) &
```
jriedl, line 186: An `wait` is needed here to avoid unintended parallelism.

PR #4592, opened 2024-12-11
krimdok_marc_pipeline.sh: Fix broken GesamtTiteldaten references

mtrojan-ub:


cpp/pipelines/krimdok_marc_pipeline.sh
```diff
@@ -197,7 +197,7 @@ wait
 
 StartPhase "Appending Literary Remains Records"
 (create_literary_remains_records --no-subsystems \
-                                 GesamtTiteldaten-post-phase"$((PHASE-3))"-"${date}".mrc \
+                                 GesamtTiteldaten-post-phase"$((PHASE-2))"-"${date}".mrc \
```
mtrojan-ub, line 200: Note: Fixing these references should solve the problem, but i wonder if it would be better to move the phases in a way that multiple authority-related phases are executed in a sequence instead of jumping between title and authority related phases.

PR #4595, opened 2025-01-23
Preparing for Ubuntu 24.04 LTS

steven-lolong:


mtrojan-ub:
Aren't there any changes to cpp/data/installer/scripts/install_ubuntu_packages.sh?

cpp/installer.cc
```diff
@@ -809,9 +809,16 @@ void ConfigureVuFind(const bool production, const VuFindSystemType vufind_system
...
     TemporaryChDir tmp2(VUFIND_DIRECTORY);
-    ExecUtil::ExecOrDie(ExecUtil::LocateOrDie("php"), { "-d", "default_socket_timeout=600", ExecUtil::LocateOrDie("composer"), "install" });
-    // We explicitly need to use sudo here, even if we're already root, or it will fail,
-    // see https://stackoverflow.com/questions/16151018/how-to-fix-npm-throwing-error-without-sudo
+
+    // execv will exit with -1 when creating a sub-process
+    int exit_code;
+    const std::string exec_composer =
+        ExecUtil::LocateOrDie("php") + " -d default_socket_timeout=600 " + ExecUtil::LocateOrDie("composer") + " install";
+    if ((exit_code = ::system(exec_composer.c_str())) != 0) {
```
mtrojan-ub, line 817: I still think that ::system should not be called here directly. Instead we should use a corresponding function from ExecUtil. If ExecUtil::ExecOrDie is not aware of the error in this case, this should be seen as a bug and be fixed in the function itself (since it is used in countless cases where the same problem could occur). In a very special case we could also add a new function to ExecUtil to call `::system`, but only if there is absolutely no alternative and we have a very good reason to do so.

PR #4604, opened 2025-03-20
Adjust rss_aggregator and update cronjob

hlindeboom:
Downloads RSS Feeds when --use-web-proxy is set and generates a subsystem specific xml file based on the provided subsystem. Updated the cronjob accordingly and created a shell script for the system_update.

### DIFF
--- a/cpp/data/installer/ixtheo.cronjobs
+++ b/cpp/data/installer/ixtheo.cronjobs
@@ -46,7 +46,7 @@ LOG_DIR=/usr/local/var/log/tuefind
 45 22 * * * "$BIN/new_journal_alert" churchlaw {churchlaw_host} "IxTheo Team<no-reply@ub.uni-tuebingen.de>" "ChurchLaw" > "$LOG_DIR/new_journal_alert_churchlaw.log" 2>&1
 0 23 * * * "$BIN/update_tad_email_acl.sh" "$EMAIL"
 15 23 * * * "$BIN/rss_aggregator" "--download-feeds" "--use-web-proxy" "relbib" "$EMAIL" "/usr/local/vufind/public/relbib_docs/news.rss" > "$LOG_DIR/rss_aggregator_relbib.log" 2>&1
-25 23 * * * "$BIN/rss_aggregator" "ixtheo" "$EMAIL" "/usr/local/vufind/public/docs/news.rss" > "$LOG_DIR/rss_aggregator_ixtheo.log" 2>&1
+30 23 * * * "$BIN/rss_aggregator" "ixtheo" "$EMAIL" "/usr/local/vufind/public/docs/news.rss" > "$LOG_DIR/rss_aggregator_ixtheo.log" 2>&1
 45 23 * * * "$BIN/rss_subset_aggregator" "--mode=email" "RelBib Team<no-reply@ub.uni-tuebingen.de>" "relbib" > "$LOG_DIR/rss_subset_aggregator_relbib.log" 2>&1
 50 23 * * * "$BIN/rss_subset_aggregator" "--mode=email" "IxTheo Team<no-reply@ub.uni-tuebingen.de>" "ixtheo" > "$LOG_DIR/rss_subset_aggregator_ixtheo.log" 2>&1
 30 22 * * 0 "$BIN/generate_kalliope_originators.py" > "$LOG_DIR/generate_kalliope_originators.py" 2>&1'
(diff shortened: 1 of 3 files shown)


cpp/rss_aggregator.cc
```diff
@@ -46,7 +46,7 @@ namespace {
 
 [[noreturn]] void Usage() {
     ::Usage(
-        "[--use-web-proxy] subsystem_type email_address xml_output_path\n"
+        "optional:[--use-web-proxy] subsystem_type email_address xml_output_path\n"
```
mtrojan-ub, line 49: Whether or not to dowload and whether or not to use the web proxy when downloading are 2 different things and should be expressed within 2 different parameters

cpp/rss_aggregator.cc
```diff
@@ -46,7 +46,7 @@ namespace {
 
 [[noreturn]] void Usage() {
     ::Usage(
-        "[--use-web-proxy] subsystem_type email_address xml_output_path\n"
+        "optional:[--download-feeds] optional:[--use-web-proxy] subsystem_type email_address xml_output_path\n"
```
mtrojan-ub, line 49: brackets `[]` already mean that something is optional, so you don't need to write `optional:` as prefix.

And since `--use-web-proxy` only makes sense if `--download-feeds` is given, the documentation should be like this:

`[--download-feeds [--use-web-proxy]]`

cpp/data/installer/ixtheo.cronjobs
```diff
@@ -45,8 +45,8 @@ LOG_DIR=/usr/local/var/log/tuefind
 30 22 * * * "$BIN/new_journal_alert" bibstudies {bibstudies_host} "IxTheo Team<no-reply@ub.uni-tuebingen.de>" "BibStudies" > "$LOG_DIR/new_journal_alert_bibstudies.log" 2>&1
 45 22 * * * "$BIN/new_journal_alert" churchlaw {churchlaw_host} "IxTheo Team<no-reply@ub.uni-tuebingen.de>" "ChurchLaw" > "$LOG_DIR/new_journal_alert_churchlaw.log" 2>&1
 0 23 * * * "$BIN/update_tad_email_acl.sh" "$EMAIL"
-15 23 * * * "$BIN/rss_aggregator" "--use-web-proxy" "relbib" "$EMAIL" "/usr/local/vufind/public/relbib_docs/news.rss" > "$LOG_DIR/rss_aggregator_relbib.log" 2>&1
-20 23 * * * "$BIN/rss_aggregator" "--use-web-proxy" "ixtheo" "$EMAIL" "/usr/local/vufind/public/docs/news.rss" > "$LOG_DIR/rss_aggregator_ixtheo.log" 2>&1
+15 23 * * * "$BIN/rss_aggregator" "--download-feeds" "--use-web-proxy" "relbib" "$EMAIL" "/usr/local/vufind/public/relbib_docs/news.rss" > "$LOG_DIR/rss_aggregator_relbib.log" 2>&1
+25 23 * * * "$BIN/rss_aggregator" "ixtheo" "$EMAIL" "/usr/local/vufind/public/docs/news.rss" > "$LOG_DIR/rss_aggregator_ixtheo.log" 2>&1
```
mtrojan-ub, line 49: maybe we should use 30 instead of 25

cpp/data/system_updates/157.sh
```diff
@@ -0,0 +1,19 @@
...
+        exit 0
+    fi
+
+    if [ ! -r ${root_crontab} ]; then
+        echo "${root_crontab} not readable - Aborting"
+        exit 1
+    fi
+
+    sed --in-place 's|15 23 \* \* \* "\$BIN/rss_aggregator" "--use-web-proxy" "relbib" "\$EMAIL" "/usr/local/vufind/public/relbib_docs/news.rss" > "\$LOG_DIR/rss_aggregator_relbib.log" 2>\&1|15 23 * * * "$BIN/rss_aggregator" "--download-feeds" "--use-web-proxy" "relbib" "$EMAIL" "/usr/local/vufind/public/relbib_docs/news.rss" > "$LOG_DIR/rss_aggregator_relbib.log" 2>\&1|' $root_crontab
+    sed --in-place 's|20 23 \* \* \* "\$BIN/rss_aggregator" "--use-web-proxy" "ixtheo" "\$EMAIL" "/usr/local/vufind/public/docs/news.rss" > "\$LOG_DIR/rss_aggregator_ixtheo.log" 2>\&1|25 23 * * * "$BIN/rss_aggregator" "ixtheo" "$EMAIL" "/usr/local/vufind/public/docs/news.rss" > "$LOG_DIR/rss_aggregator_ixtheo.log" 2>\&1|' $root_crontab
```
mtrojan-ub, line 18: in the ixtheo.cronjobs file it says 25 but in the system update it says 20, i guess we should use 30 for ixtheo in both cases

PR #4608, opened 2025-04-08
Riedl

jriedl:


jriedl:
The non-generate_merge_order parts can be ignored and just merged

cpp/dpt/convert_dpt_bookchapters.cc
```diff
@@ -355,7 +355,7 @@ void ExtractBookInformation(File * const dpt_books_file) {
 
 
 int Main(int argc, char *argv[]) {
-    if (argc < 2)
+    if (argc < 3)
```
mtrojan-ub, line 358: argc is also tested against < 6 on line 381 / maybe these tests should be switched because of --argc?

PR #4619, opened 2025-05-21
add a new column 'type' to table 'vufind.tuefind_rss_feeds'

steven-lolong:


cpp/data/system_updates/159.vufind.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `vufind`.`tuefind_rss_feeds` 
+ADD COLUMN `type` SET('feed', 'news', 'calendar') NOT NULL DEFAULT 'news' AFTER `subsystem_types`;
```
mtrojan-ub, line 2: - Instead of SET we should use ENUM
- Allowed values should only be 'news' or 'calendar', not 'feed'

PR #4626, opened 2025-07-14
Update the parser for id location in the object #3353

steven-lolong:


cronjobs/add_authority_external_ref.py
```diff
@@ -141,25 +141,34 @@ def Main():
...
+                
                 util.ExecOrDie(util.Which("rm"), [
                                "-f", f"/tmp/{newer_file_name}"])
                 util.ExecOrDie(util.Which("rm"), [
                                "-f", f"/tmp/{input_file_name_for_add_auth}"])
+            
+                file_size = os.path.getsize(f"/tmp/{gnd_wiki_file}")
+                
+                if(file_size == 0):
+                    print("Process 6/7 -- Sending error message via email")
```
mtrojan-ub, line 153: This is not step 6/7 => If the GND file is empty, it should not be copied to the destination. The script should abort after sending the mail and print a proper message.

PR #4704, opened 2026-05-27
To optimize data import, select only full-text records modified within the past 24 hours

steven-lolong:


cpp/data/system_updates/168.sh
```diff
@@ -0,0 +1,58 @@
...
+        "index.default_pipeline": "add_timestamp"
+      }'
+
+curl -X PUT "$URL_SET_DEFAULT_PIPELINE_FULLTEXT_CACHE_HTML" \
+    "${username_login_args[@]}" \
+    "${authorization_args[@]}" \
+    -H "Content-Type: application/json" \
+    -d '{
+        "index.default_pipeline": "add_timestamp"
+      }'
```
jriedl: This script only adjusts existing installations. Logic for new installations is still to be integrated into the installer

cpp/pipelines/create_diffs_for_solr_import.cc
```diff
@@ -168,8 +178,10 @@ int Main(int argc, char *argv[]) {
                                                  &ids_need_to_be_deleted);
     std::thread thread_collect_ids_to_be_imported(CollectIDsToBeImported, previous_ppn_to_record_hash_map, current_ppn_to_record_hash_map,
                                                   &ids_need_to_be_imported);
+    std::thread thread_collect_ids_to_be_imported_from_full_text_cache(CollectIDsToBeImportedFromFullTextCache, &ids_need_to_be_imported);
```
jriedl: The existing threads work on different sets, so concurrency does not pose problem. The new thread also writes to  `ids_need_to_be_imported`, so thread safety must be taken into account. Alternatively, the addition is just done threadless afterwards.

cpp/data/system_updates/168.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
```
jriedl, line 1: Just call `create_pipeline.sh` ?

PR #4708, opened 2026-06-12
fix the missing explainOther content needed by Explain feature #3504

steven-lolong:


mtrojan-ub:
At the moment i think it would be better to remove the whole component, but this is still to be discussed (see ubtue/tuefind#3504). Also i would not recommend merging this to master because master is still related to 10.2 and we are unaware of the side effects.

As discussed, i converted it to draft and leave it open for further discussion.

PR #4709, opened 2026-06-12
Downloader: Use cookies across multiple downloader instances

mtrojan-ub:
After enabling debugging_ = true, before this change:

> 2026-06-12T09:09:05 INFO extract_keywords_for_translation: in void Downloader::debugFunction(CURL *, curl_infotype, char *, size_t) --> informational text: WARNING: failed to save cookies in : Failed writing received data to disk/application

It would work within downloader_test before though if you call it once via shell and pass multiple urls, but if you call it seperately it won't. Also, before this change it depended on whether there is only 1 downloader object being created and re-used, or a new object being created for each download.

**Example 1)**
Works before the change: (As one-liner, because [implementation re-uses same downloader instance](https://github.com/ubtue/ub_tools/blob/4d892df0857eeca323140f6264f2880da8d857d4/cpp/test/downloader_test.cc#L73))
```
downloader_test 'https://httpbin.dev/cookies/set?foo=bar' 'https://httpbin.dev/cookies/set'
```
Did not work before the change:
```
downloader_test 'https://httpbin.dev/cookies/set?foo=bar'
downloader_test 'https://httpbin.dev/cookies/set'
```

**Example 2)**
extract_keywords_for_translation: Does not re-use cookies before the change, since a [new downloader is initialized for every single query](https://github.com/ubtue/ub_tools/blob/4d892df0857eeca323140f6264f2880da8d857d4/cpp/pipelines/extract_keywords_for_translation.cc#L229)

### DIFF
--- a/cpp/lib/include/Downloader.h
+++ b/cpp/lib/include/Downloader.h
@@ -101,7 +101,8 @@ public:
         std::string authentication_username_;
         std::string authentication_password_;
         bool use_cookies_txt_;
-        std::string cookies_txt_path_;
+        std::string
+            cookies_txt_path_; // default empty = RAM only, set this to a specific file path to share cookies across instances+processes
         bool fail_on_error_;
 
     public:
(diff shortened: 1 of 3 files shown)


cpp/lib/include/Downloader.h
```diff
@@ -113,7 +114,8 @@ class Downloader {
                         const bool ignore_ssl_certificates = false, const std::string &proxy_host_and_port = "",
                         const std::vector<std::string> &additional_headers = {}, const std::string &post_data = "",
                         const std::string &authentication_username = "", const std::string &authentication_password = "",
-                        const bool use_cookies_txt = true, const bool fail_on_error = true);
+                        const bool use_cookies_txt = true, const std::string cookies_txt_path = "/tmp/cookies.txt",
```
jriedl: Using the same cookie file between processes might lead to race conditions on file level. As discussed, potentially stick with the standard beahviour and enable shared cookie file only in specific explicit use cases.

PR #4715, opened 2026-08-03
Adjustment for non-Elasticsearch machine 

steven-lolong:


cpp/data/system_updates/168.sh
```diff
@@ -1,5 +1,22 @@
 #!/bin/bash
 
+FILE="/usr/local/var/lib/tuelib/Elasticsearch.conf"
+
+if [ ! -f "$FILE" ]; then
+    echo "Error: Configuration file $FILE not found."
```
mtrojan-ub, line 6: Both messages are not really errors here. On machines that dont have elasticsearch installed, the updates will not be skipped. Please use something like:

"Skipping Elasticsearch-related system update 168 due to missing Elasticsearch configuration on this machine."

cpp/data/system_updates/168.sh
```diff
@@ -1,5 +1,22 @@
 #!/bin/bash
 
+FILE="/usr/local/var/lib/tuelib/Elasticsearch.conf"
+
+if [ ! -f "$FILE" ]; then
```
mtrojan-ub, line 5: This check does not really make sense, because the file will exist on all machines even if Elasticsearch is not installed. The file will be copied from the network drive to all machines during `make install`, no matter whether Elasticsearch is installed or not.

E.g. the file exists on ub16, but elasticsearch is not installed there.

Maybe we should just check whether the `/usr/share/elasticsearch` directory exists?

mtrojan-ub: Or maybe this is not even enough => when pulling on ub26 it might migrate the ub26 elasticsearch, but when pulling it on ub16 which also points to ub26 via config, will this also try to migrate the same (remote) server that has already been migrated?

Please be very precise about on which servers / installation types this update should run or not run.

mtrojan-ub: ... if things get to complicated, executing the script once manually on all machines that are affected would also be viable i think (=entirely removing the system update), but i just think it does not make sense how the system update is implemented now

mtrojan-ub: As discussed with @steven-lolong we should discuss this question with @jriedl when he's back in the office.

